Issue #54, opened 2013-05-16
Do we really need email/author fields on Post model


rjmackay:
We have a similar split between incident_person and user tables in 2.x
Would it make more sense to just keep all this info in the users table in 3.x?
- Any unauthenticated report should get a user autocreated. reset/activate link would be emailed to them
- If they want to submit another report they have to sign in
- Completely anonymous reports have no user and no email details.

Issue #53, opened 2013-05-16
Add media model and API


rjmackay:
Working on the migration script and figured that just dropping image urls into a single Post_Varchar isn't going to work well.
We may need places to store extra info: other image sizes, image meta data, etc.
And we need to share this between: image fields, any future file upload types, category images, site banner, any other file uploads.

I'm thinking
- A media model
- /media/:id api endpoint
  - Needs to allow uploading a file (image, etc) . probably to live in `application/media/uploads`
  - Allow adding caption
  - Add a controller application/classes/Controller/Api/Media.php
- Media field - references media table
  - Image field? with extra handling for image media..

Issue #17, opened 2013-03-13
API /users


rjmackay:
Write spec (via wiki docs) for users api, then build

PR #60, opened 2013-06-04
Ushahidi 2.x to 3.x Migration Script

rjmackay:
This is the first cut at a 2.x to 3.x migration script. You can import from DB or API.

To view task docs:
`./minion --task=ushahidi:import2x --help`

To import from API
`./minion --task=ushahidi:import2x --source=api --url=http://zombiereports.com`

To import from DB
`./minion --task=ushahidi:import2x --source=db --database=ushahidi_zombie --username=ushahidi --password=ushahidi`

**Things not covered:**
- Can't get unapproved reports via API - this is a 2.x issue
- Can't get incident person data from the API - 2.x issue
- Can't get hidden categories from API - 2.x issue
- We don't have anywhere to store category color, icon or visibility yet
- Not importing users - this code would be best built with the users API #17
- Not importing custom fields or data yet - I'd rather just spin this out to another task to build later.
- Can't import photos - no way to handle media yet #53
- Not importing incident person data from DB - need to figure out where to store it. Related #54

**Testing**
I'm struggling to figure out how to test this. I can write some behat magic to run the migration script but it would rely on external factors: 1. external 2.x db or 2. external 2.x api
Which would mean it failing anytime those weren't set up. Any ideas?

**Tasks**
- [x] Import from api
- [x] Import from database
- [x] Import functional tests
- [ ] Unit test import functions
- [x] Better handling of errors when creating form

### DIFF
--- a/application/classes/Task/Ushahidi/Import2x.php
+++ b/application/classes/Task/Ushahidi/Import2x.php
@@ -753,8 +753,7 @@ class Task_Ushahidi_Import2x extends Minion_Task {
 					);
 				}
 				// Source report has person info
-				// @todo handle incident_person w/o email, requires changes to Posts API
-				elseif (! empty($report['person_email']) /*OR ! empty($report['person_first']) OR ! empty($report['person_last'])*/)
+				elseif (! empty($report['person_email']) OR ! empty($report['person_first']) OR ! empty($report['person_last']))
 				{
 					// Already got this user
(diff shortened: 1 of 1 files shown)


alexhalbi:
I only get the "Please enable the Minion module for CLI support.".error when starting this on Windows with: 
C:\xampp\htdocs\ushahidi-platform>php httpdocs\index.php --uri=minion
Please enable the Minion module for CLI support.

PR #276, opened 2015-07-07
T1641 post filtering

rjmackay:
[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform/276)

PR #277, opened 2015-07-07
Prevent created timestamp overwrite

rjmackay:
Added step to remove created date sent by client to avoid overwriting of the existing timestamp

Add api integration tests to ensure timestamp is as expected after entity updates

T1493

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform/277)

### DIFF
--- a/application/classes/Ushahidi/Repository.php
+++ b/application/classes/Ushahidi/Repository.php
@@ -259,6 +259,10 @@ abstract class Ushahidi_Repository implements
 			return 0; // nothing would be updated, just ignore
 		}
 
+		if(array_key_exists('created', $input)){
+			unset($input['created']);
+		}
+
 		$query = DB::update($this->getTable())->set($input);
 		foreach ($where as $column => $value)
 		{


review-ninja:
<a href="http://app.review.ninja/ushahidi/platform/pull/277" target="_blank">ReviewNinja

src/Core/Usecase/UpdateUsecase.php
```diff
@@ -69,9 +69,13 @@ public function isSearch()
 	// Usecase
 	public function interact()
 	{
+    //remove created time stamp if present to prevent overwriting
+    if ($this->payload['created']) {
```
rjmackay: I don't think this should happen in the Usecase.. there are 2 places we could handle it:
1. If you add `created` to the immutable fields on the Entity that would work.. however it might not work for `updated` since that needs to change
2. Drop/Overwrite the created and updated values in the repository class.. I think thats probably the best place for now.

rjmackay:
Looks like you've merged changes from T1641 / #276 into this PR again.. they need to be separate. Can you update again?

billdoran:
Ok, I think that's rectified now.

rjmackay:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/277)

PR #661, opened 2015-08-27
Changed icon tag validator to support format of icon class with spaces

willdoran:
@rjmackay @jasonmule Updated validator to support broader format for icon classes

### DIFF
--- a/application/classes/Ushahidi/Validator/Tag/Update.php
+++ b/application/classes/Ushahidi/Validator/Tag/Update.php
@@ -54,8 +54,8 @@ class Ushahidi_Validator_Tag_Update extends Validator
 				['color'],
 			],
 			'icon' => [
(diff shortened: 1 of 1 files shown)


application/classes/Ushahidi/Validator/Tag/Update.php
```diff
@@ -54,7 +54,8 @@ protected function getRules()
 				['color'],
 			],
 			'icon' => [
-				['alpha_dash'],
+				// alphas, numbers, punctuation, and spaces
```
rjmackay: punctuation? I can see `icon-names` and `icon something` .. but other punctuation?

willdoran: I'll pull out the punctuation, none of the font awesome icons have numbers so I guess those can be removed until we hit an icon pack that has them.

PR #662, opened 2015-08-27
changed mechanism for filtering of multiple stages to include another…

### LINKED ISSUE #401: Filter by task
Confirm issue seen previously where filtering by stages from multiple forms gives unexpected results.

willdoran:
… level

@rjmackay @jasonmule I believe this fixes the ordering and priority issue with the previous query. It does however add a whole new level.

Fixes #401

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform/662)

### DIFF
--- a/application/classes/Ushahidi/Repository/Post.php
+++ b/application/classes/Ushahidi/Repository/Post.php
@@ -350,17 +350,20 @@ class Ushahidi_Repository_Post extends Ushahidi_Repository implements
 				// and a check that the current post has not already completed this stage
 				->on('form_stages.form_id', 'IN', $forms_sub)
 				->on('form_stages.id', 'NOT IN', $stages_posts)
-        // We group the results by post id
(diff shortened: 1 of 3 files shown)


rjmackay:
I'm still getting super inconsistent results.
Here's what I've done to test:
- Create a post type
- Add 3 steps
- Reorder the steps
- Create another post type, with a single step
- Add a few posts of each type, with various stages complete
  - I've intentionally completed some stages out of order: ie. 1st and third stage, or just the 2nd stage.

Results are really inconsistent:
- I get the right things in "Current Stage" displayed in the list. So that JS seems to be solid
- Filtering by stages only returns posts of the correct post type - so thats right
- However I'm not getting posts in the right current stage at all. Seems pretty similarly broken to before.

Db dump I tested with here: https://dl.dropboxusercontent.com/u/123897/ushahidi-2015-09-04.sql

PR #676, opened 2015-09-01
519 notifications

jasonmule:
- API to handle notification subscriptions when a user wants to receive notifications when a new post is added to a collection.

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform/676)


application/classes/Ushahidi/Repository/Notification.php
```diff
@@ -0,0 +1,100 @@
...
+
+use Ushahidi\Core\Entity;
+use Ushahidi\Core\SearchData;
+use Ushahidi\Core\Entity\Notification;
+
+class Ushahidi_Repository_Notification extends Ushahidi_Repository
+{
+	protected function notificationExists(Entity $entity)
+	{
+		$result = DB::select('id')
```
rjmackay: Ideally use `$this->selectQuery()` here or at least reference `this->getTable()` instead of hard coding the table name

application/classes/Ushahidi/Repository/Notification.php
```diff
@@ -0,0 +1,100 @@
...
+	{
+		// Return a 204 since we won't be creating this subscription
+		if ($this->notificationExists($entity)) {
+			return;
+		}
+
+		$state = [
+			'contact_id' => $entity->contact_id,
+			'created' => time(),
+			'is_subscribed' => 1,
```
rjmackay: When would is_subscribed be 0?

application/classes/Ushahidi/Validator/Notification/Update.php
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+use Ushahidi\Core\Tool\Validator;
+
+class Ushahidi_Validator_Notification_Update extends Validator
+{
+	protected $repo;
+	protected $default_error_source = 'notification';
+
+	public function __construct(Ushahidi_Repository_Notification $repo)
```
rjmackay: Shouldn't be relying on `Ushahidi_Repository_Notification` rather use `Ushahidi\Core\Entity\NotificationRepository`
If that doesn't exist yet, create it :)

application/routes/default.php
```diff
@@ -215,6 +215,32 @@
 	));
 
 /**
+ * Notification subscriptions API Route
+ */
+Route::set('notifications-api', $apiBase . 'notifications(/<id>)',
```
rjmackay: Would these not have been caught by the `api` route above anyway?

src/Core/Entity/Notification.php
```diff
@@ -0,0 +1,49 @@
...
+	protected $set_id;
+	protected $is_subscribed;
+	protected $created;
+
+	// DataTransformer
+	protected function getDefinition()
+	{
+		return [
+			'id'            => 'int',
+			'contact_id'    => 'int',
```
rjmackay: I think we should also have aliases for `contact` and `set` here.. see what we do with foreign keys on other Entities

application/classes/Ushahidi/Formatter/Notification.php
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * @author     Ushahidi Team <team@ushahidi.com>
+ * @package    Ushahidi\Application
+ * @copyright  2014 Ushahidi
+ * @license    https://www.gnu.org/licenses/agpl-3.0.html GNU Affero General Public License Version 3 (AGPL3)
+ */
+
+use Ushahidi\Core\Traits\FormatterAuthorizerMetadata;
+
+class Ushahidi_Formatter_Notification extends Ushahidi_Formatter_API
```
rjmackay: Need to add formatter handling for relations so that we're returning id + URL

application/classes/Ushahidi/Validator/Notification/Create.php
```diff
@@ -0,0 +1,41 @@
...
+ * @package    Ushahidi\Application
+ * @copyright  2014 Ushahidi
+ * @license    https://www.gnu.org/licenses/agpl-3.0.html GNU Affero General Public License Version 3 (AGPL3)
+ */
+
+use Ushahidi\Core\Tool\Validator;
+use Ushahidi\Core\Entity\ContactRepository;
+use Ushahidi\Core\Entity\SetRepository;
+
+class Ushahidi_Validator_Notification_Create extends Validator
```
rjmackay: If we're allowing contact creation through the API, I suspect we should insist those contacts have an associated user id.

application/classes/Ushahidi/Validator/Contact/Update.php
```diff
@@ -0,0 +1,90 @@
...
+ * @copyright  2014 Ushahidi
+ * @license    https://www.gnu.org/licenses/agpl-3.0.html GNU Affero General Public License Version 3 (AGPL3)
+ */
+
+use Ushahidi\Core\Entity;
+use Ushahidi\Core\Entity\Contact;
+use Ushahidi\Core\Tool\Validator;
+use Ushahidi\Core\Entity\ContactRepository;
+
+class Ushahidi_Validator_Contact_Update extends Validator
```
rjmackay: If we're allowing contact creation through the API, I suspect we should insist those contacts have an associated user id.

src/Core/Usecase/Notification/CreateNotification.php
```diff
@@ -0,0 +1,47 @@
...
+ * @copyright  2014 Ushahidi
+ * @license    https://www.gnu.org/licenses/agpl-3.0.html GNU Affero General Public License Version 3 (AGPL3)
+ */
+
+namespace Ushahidi\Core\Usecase\Notification;
+
+use Ushahidi\Core\Entity;
+use Ushahidi\Core\Usecase\CreateUsecase;
+
+class CreateNotification extends CreateUsecase
```
rjmackay: What differs here from the parent CreateUsecase? - can you call that out in a comment?

src/Core/Usecase/Notification/UpdateNotification.php
```diff
@@ -0,0 +1,47 @@
...
+ * @license    https://www.gnu.org/licenses/agpl-3.0.html GNU Affero General Public License Version 3 (AGPL3)
+ */
+
+namespace Ushahidi\Core\Usecase\Notification;
+
+use Ushahidi\Core\Data;
+use Ushahidi\Core\Entity;
+use Ushahidi\Core\Usecase\UpdateUsecase;
+
+class UpdateNotification extends UpdateUsecase
```
rjmackay: What differs here from the parent UpdateUsecase? - can you call that out in a comment?

src/Core/Usecase/Notification/UpdateNotification.php
```diff
@@ -0,0 +1,47 @@
...
+		// ... verify that the entity is in a valid state
+		$this->verifyValid($entity);
+
+		// ... persist the changes
+		$this->repo->update($entity);
+
+		// ... check that the entity can be read by the current user
+		if ($this->auth->isAllowed($entity, 'read') and
+			// ... check that the subscription status changed
+			$entity->hasChanged('is_subscribed')) {
```
rjmackay: Does this plan on just setting is_subscribed = 0 rather than deleting a subscription?

.. I think we should still return the full Entity after the Update regardless. Its more consistent for the API user.. they always get an object back.

rjmackay:
There's very little to change about this.. super impressive work!
Could you summarize how you expect to use this in the client? (ie. what API calls will get made) just so I'm not missing something

application/classes/Ushahidi/Repository/Notification.php
jasonmule: When a user is not subscribed. We are tracking subscriptions using this flag instead of deleting them?

application/classes/Ushahidi/Validator/Notification/Update.php
jasonmule: I know, I know :-)

src/Core/Usecase/Notification/UpdateNotification.php
jasonmule: I am probably not going to be needing this since we are always returning an object back to the user.

jasonmule:
# Using the Contacts API
So, on the notifications page, the user expects to be able to subscribe to a notification by selecting what contact she wants to use -- if there is no contact, add it from a pre-filled email address retrieved from the user object. The user should also be able to add a phone number as a contact on this page.

# Using the Notifications API
On the notifications page, a user can use a newly saved or existing contact id from the Contacts API above as well as a collection id to subscribe to a new notification. On the client, the collection id is retrieved when a user follows the "Add notification" link from a collection.

Does this make sense?

rjmackay:
Reviewed 6 of 23 files at r1, 20 of 21 files at r2.
Review status: all files reviewed at latest revision, 15 unresolved discussions, all commit checks successful.

---

<sup>**[application/classes/Ushahidi/Formatter/Notification.php, line 31 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JzTNb7TjAOQC3L7QBsl)** ([raw file](https://github.com/ushahidi/platform/blob/465848fd3e6e6b69080b998b5d246fc7fa138ee2/application/classes/Ushahidi/Formatter/Notification.php#L31)):</sup>
Need to do the same for user_id

---

<sup>**[application/classes/Ushahidi/Repository/Notification.php, line 13 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JzTOa1vEgHx-0BcU7l1)** ([raw file](https://github.com/ushahidi/platform/blob/465848fd3e6e6b69080b998b5d246fc7fa138ee2/application/classes/Ushahidi/Repository/Notification.php#L13)):</sup>
We don't need this. Its a left over in other repos..

---

<sup>**[application/classes/Ushahidi/Repository/Notification.php, line 46 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JzTQBHK5OhK-zCn2q4k)** ([raw file](https://github.com/ushahidi/platform/blob/465848fd3e6e6b69080b998b5d246fc7fa138ee2/application/classes/Ushahidi/Repository/Notification.php#L46)):</sup>
I really like this. Really nice re-use of existing traits that makes this super simple :)

---

<sup>**[application/classes/Ushahidi/Repository/Notification.php, line 61 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JzTQR-aHe-20pt996L4)** ([raw file](https://github.com/ushahidi/platform/blob/465848fd3e6e6b69080b998b5d246fc7fa138ee2/application/classes/Ushahidi/Repository/Notification.php#L61)):</sup>
not something to change now. But do you think this pattern makes sense? (using user instead of user_id) or should we just go to using XYZ_id everywhere and make the code simpler?

---

<sup>**[application/classes/Ushahidi/Repository/Notification.php, line 93 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JyGDUuv28uzWxoiA38A-r1-76)** ([raw file](https://github.com/ushahidi/platform/blob/358e7c888a23c9fd52792f2a80cae75e9321c022/application/classes/Ushahidi/Repository/Notification.php#L93)):</sup>
Got it. I'm not sure thats necessary. Is there some reason you think we need to keep the notifications around?

---

<sup>**[application/classes/Ushahidi/Validator/Contact/Update.php, line 18 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JyOze8GAQzOqa_rf7Bk-r1-17)** ([raw file](https://github.com/ushahidi/platform/blob/358e7c888a23c9fd52792f2a80cae75e9321c022/application/classes/Ushahidi/Validator/Contact/Update.php#L18)):</sup>
^ need to update the Contact Create validator to require user_id (at least when coming from the API. might need an alternate validator for contacts created when messages are received) *and* inherit from the Contact Update validator.

---

<sup>**[application/classes/Ushahidi/Validator/Notification/Create.php, line 16 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JzTRFFkSxWEaJ3Ut5Uz)** ([raw file](https://github.com/ushahidi/platform/blob/465848fd3e6e6b69080b998b5d246fc7fa138ee2/application/classes/Ushahidi/Validator/Notification/Create.php#L16)):</sup>
This class should extend the Update validator to avoid repeating rules. We do this elsewhere

---

<sup>**[application/classes/Ushahidi/Validator/Notification/Create.php, line 31 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JzTRM-U5ZhAQsgxLuQ1)** ([raw file](https://github.com/ushahidi/platform/blob/465848fd3e6e6b69080b998b5d246fc7fa138ee2/application/classes/Ushahidi/Validator/Notification/Create.php#L31)):</sup>
Need a not_empty rule here to ensure contacts have a user_id 

---

<sup>**[src/Core/Entity/Contact.php, line 30 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JzTTWI_rEitFQZa6sZt)** ([raw file](https://github.com/ushahidi/platform/blob/465848fd3e6e6b69080b998b5d246fc7fa138ee2/src/Core/Entity/Contact.php#L30)):</sup>
What are you using can_notify for? .. why not allow notifications to any contact?

---

<sup>**[src/Core/Entity/Contact.php, line 53 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JzTTcTq_33vXrreWwFX)** ([raw file](https://github.com/ushahidi/platform/blob/465848fd3e6e6b69080b998b5d246fc7fa138ee2/src/Core/Entity/Contact.php#L53)):</sup>
Should be boolean I think or maybe `bool`

---

<sup>**[src/Core/Entity/Notification.php, line 44 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JzTTliok9PK70M2N3Ui)** ([raw file](https://github.com/ushahidi/platform/blob/465848fd3e6e6b69080b998b5d246fc7fa138ee2/src/Core/Entity/Notification.php#L44)):</sup>
should be `bool`

---

<sup>**[src/Core/Tool/Authorizer/ContactAuthorizer.php, line 47 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JzTTxfojv4a_Bcj_hac)** ([raw file](https://github.com/ushahidi/platform/blob/465848fd3e6e6b69080b998b5d246fc7fa138ee2/src/Core/Tool/Authorizer/ContactAuthorizer.php#L47)):</sup>
curious if this means a user can delete all their contacts? That would cause problems with any messages we have tied to those contacts

---

<sup>**[src/Core/Tool/Authorizer/NotificationAuthorizer.php, line 47 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JzTUG8vtppl38VbGFox)** ([raw file](https://github.com/ushahidi/platform/blob/465848fd3e6e6b69080b998b5d246fc7fa138ee2/src/Core/Tool/Authorizer/NotificationAuthorizer.php#L47)):</sup>
I assume this is to disallow delete? why? Can you comment it explicitly..

---

<sup>**[src/Core/Usecase/Notification/CreateNotification.php, line 23 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JzTUzvIrzkNcW-gP-7n)** ([raw file](https://github.com/ushahidi/platform/blob/465848fd3e6e6b69080b998b5d246fc7fa138ee2/src/Core/Usecase/Notification/CreateNotification.php#L23)):</sup>
We should see if we can turn this into a sharable trait at some point.. non blocking right now though

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/676)

jasonmule:
Review status: all files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

<sup>**[application/classes/Ushahidi/Repository/Notification.php, line 0 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JyGDUuv28uzWxoiA38A-r1-76)** ([raw file](https://github.com/ushahidi/platform/blob/358e7c888a23c9fd52792f2a80cae75e9321c022/application/classes/Ushahidi/Repository/Notification.php#L0)):</sup>
I thought that we may need to track unsbscriptions at some point.

---

<sup>**[application/classes/Ushahidi/Repository/Notification.php, line 0 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JzTQR-aHe-20pt996L4)** ([raw file](https://github.com/ushahidi/platform/blob/465848fd3e6e6b69080b998b5d246fc7fa138ee2/application/classes/Ushahidi/Repository/Notification.php#L0)):</sup>
I think using user instead of user_id makes sense.

---

<sup>**[application/classes/Ushahidi/Repository/Notification.php, line 13 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JzTOa1vEgHx-0BcU7l1)** ([raw file](https://github.com/ushahidi/platform/blob/465848fd3e6e6b69080b998b5d246fc7fa138ee2/application/classes/Ushahidi/Repository/Notification.php#L13)):</sup>
Could you clarify? Don't we need this to implement `setSearchConditions`?

---

<sup>**[src/Core/Entity/Contact.php, line 0 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JzTTWI_rEitFQZa6sZt)** ([raw file](https://github.com/ushahidi/platform/blob/465848fd3e6e6b69080b998b5d246fc7fa138ee2/src/Core/Entity/Contact.php#L0)):</sup>
I had put this there so that users can turn off notifications for a particular contact but looking at this again, it makes sense to just send notifications out for the contacts a user has added.

---

<sup>**[src/Core/Tool/Authorizer/ContactAuthorizer.php, line 0 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JzTTxfojv4a_Bcj_hac)** ([raw file](https://github.com/ushahidi/platform/blob/465848fd3e6e6b69080b998b5d246fc7fa138ee2/src/Core/Tool/Authorizer/ContactAuthorizer.php#L0)):</sup>
They shouldn't. Will fix this.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/676)

rjmackay:
Looking good.. a few comments inline. The main change I'd like is just to ditch the is_subscribed flag, and delete old subscriptions completely..

---

Review status: all files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

<sup>**[application/classes/Ushahidi/Repository/Notification.php, line 13 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JzTOa1vEgHx-0BcU7l1)** ([raw file](https://github.com/ushahidi/platform/blob/465848fd3e6e6b69080b998b5d246fc7fa138ee2/application/classes/Ushahidi/Repository/Notification.php#L13)):</sup>
You're correct. My mistake sorry!

---

<sup>**[application/classes/Ushahidi/Repository/Notification.php, line 82 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-Jzh_NyDMRzsy5i0SFkJ)** ([raw file](https://github.com/ushahidi/platform/blob/465848fd3e6e6b69080b998b5d246fc7fa138ee2/application/classes/Ushahidi/Repository/Notification.php#L82)):</sup>
Shouldn't we at least be setting is_subscribed = 1 ? Though I guess it depends how we handle this in the UI

---

<sup>**[application/classes/Ushahidi/Repository/Notification.php, line 93 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JyGDUuv28uzWxoiA38A-r1-76)** ([raw file](https://github.com/ushahidi/platform/blob/358e7c888a23c9fd52792f2a80cae75e9321c022/application/classes/Ushahidi/Repository/Notification.php#L93)):</sup>
I'd rather we just deleted them completely. We're adding complexity (subscription are handled differently than other resource) for no real gain right now.

---

<sup>**[src/Core/Entity/Contact.php, line 0 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JzTTWI_rEitFQZa6sZt)** ([raw file](https://github.com/ushahidi/platform/blob/465848fd3e6e6b69080b998b5d246fc7fa138ee2/src/Core/Entity/Contact.php#L0)):</sup>
Ah right. It probably does make sense.. but maybe on a per-subscription basis? (ie. some subscriptions I might want via email, others via SMS). For now lets just send to all their contacts, and we can refine based on user feedback/research.

---

<sup>**[src/Core/Tool/Authorizer/ContactAuthorizer.php, line 47 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JzTTxfojv4a_Bcj_hac)** ([raw file](https://github.com/ushahidi/platform/blob/465848fd3e6e6b69080b998b5d246fc7fa138ee2/src/Core/Tool/Authorizer/ContactAuthorizer.php#L47)):</sup>
Fixed now?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/676)

rjmackay:
Review status: all files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

<sup>**[src/Core/Entity/Contact.php, line 30 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JzTTWI_rEitFQZa6sZt)** ([raw file](https://github.com/ushahidi/platform/blob/465848fd3e6e6b69080b998b5d246fc7fa138ee2/src/Core/Entity/Contact.php#L30)):</sup>
Actually just looked at the design framework for this again. You're correct, we should be toggling notifications per-contact. If you've already got this working: leave it in :)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/676)

jasonmule:
<sup>**[src/Core/Entity/Contact.php, line 30 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JzTTWI_rEitFQZa6sZt)** ([raw file](https://github.com/ushahidi/platform/blob/465848fd3e6e6b69080b998b5d246fc7fa138ee2/src/Core/Entity/Contact.php#L30)):</sup>
Could you clarify the the bit about toggling notifications per contact? The way I have this working is that can_notify simply toggles a contact to receive or stop receiving _all_ notifications.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/676)

rjmackay:
Review status: all files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

<sup>**[src/Core/Entity/Contact.php, line 30 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-JzTTWI_rEitFQZa6sZt)** ([raw file](https://github.com/ushahidi/platform/blob/465848fd3e6e6b69080b998b5d246fc7fa138ee2/src/Core/Entity/Contact.php#L30)):</sup>
Sorry. The way you have it is correct.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/676)

jasonmule:
For now, I am using a data provider to create messages when posts are added to a collection:

`bin/ushahidi dataprovider -p notifications incoming`

Also added an `Event` trait for triggering events -- currently used to get notified when new posts are added to a collection in `SetRepository`.

rjmackay:
- Dataproviders fetch() is really meant to be for incoming messages
- Notifications should really just create outgoing messages via the messages repo (and probably a usecase) then the dataproviders will pick those up and send them.

rjmackay:
Reviewed 9 of 31 files at r3, 3 of 3 files at r4.
Review status: 21 of 40 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

<sup>**[application/classes/Ushahidi/Repository/Set.php, line 41 \[r4\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-K-FsF7tZIW1x_IR9e9u)** ([raw file](https://github.com/ushahidi/platform/blob/4e461d7fc385e1cb0212e344030569c22d225481/application/classes/Ushahidi/Repository/Set.php#L41)):</sup>
Is there a reason this doesn't live in the Event trait too? or a separate Listener trait?

---

<sup>**[application/classes/Ushahidi/Repository/Set.php, line 247 \[r4\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-K-FrSbpiGICKkTBv9V8)** ([raw file](https://github.com/ushahidi/platform/blob/4e461d7fc385e1cb0212e344030569c22d225481/application/classes/Ushahidi/Repository/Set.php#L247)):</sup>
Could we move this to the `setListener` method? so theres just a single line here:
`$this->emit('PostSetEvent', $set_id, $post_id);`

---

<sup>**[composer.json, line 79 \[r4\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-K-FrjaeDKSvWvO2m5Xd)** ([raw file](https://github.com/ushahidi/platform/blob/4e461d7fc385e1cb0212e344030569c22d225481/composer.json#L79)):</sup>
do we need the `@dev` version of league/event ? or could we pull stable?

---

<sup>**[plugins/notifications/classes/DataProvider/Notifications.php, line 12 \[r4\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-K-Fp5MYyuL195lcBgFB)** ([raw file](https://github.com/ushahidi/platform/blob/4e461d7fc385e1cb0212e344030569c22d225481/plugins/notifications/classes/DataProvider/Notifications.php#L12)):</sup>
I don't think this makes much sense as a dataprovider.. except maybe as a temporary solution.

Ideally we could we just spin this out to a separate `notifications` CLI task.. which would either queue messages, or send directly and then record them into the messages repo. 
Do you think you have time to do that before launch? I'm guessing no. In which case we can launch with this, then tidy up later.

---

<sup>**[src/Core/Traits/Event.php, line 14 \[r4\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-K-FqyYSphwwG4CoA7uM)** ([raw file](https://github.com/ushahidi/platform/blob/4e461d7fc385e1cb0212e344030569c22d225481/src/Core/Traits/Event.php#L14)):</sup>
ideally this shouldn't live in the Core namespace.. or not all of it should since it has a dependencies on League\Event.. however I think we can tidy that up later. Probably by moving to a Ushahidi\App namespace, and keeping just the 'emit()' method in Core

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/676)

rjmackay:
Reviewed 2 of 31 files at r3.
Review status: 23 of 40 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/676)

jasonmule:
Noted. For some reason, I thought that this was going to be  a configurable data provider. I will go ahead and move this to a CLI task.

---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/676)

jasonmule:
<sup>**[plugins/notifications/classes/DataProvider/Notifications.php, line 12 \[r4\]](https://reviewable.io:443/reviews/ushahidi/platform/676#-K-Fp5MYyuL195lcBgFB)** ([raw file](https://github.com/ushahidi/platform/blob/4e461d7fc385e1cb0212e344030569c22d225481/plugins/notifications/classes/DataProvider/Notifications.php#L12)):</sup>
I will set this up to queue messages as a CLI task for starters

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/676)

rjmackay:
did you see my inline comment: 
>  Do you think you have time to do that before launch? I'm guessing no. In which case we can launch with this, then tidy up later.

We want to land this today if possible.. so don't worry about the CLI task if it'll take to long

jasonmule:
I moved this to a CLI task and removed the notifications data provider.

rjmackay:
Looking good


---

Review status: 18 of 38 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/676)

PR #683, opened 2015-09-02
T638 - Add features into configuration again

hollycorbett:
Features largely work and the beginnings of an UseCase put in place for updating the datasources.  Restricting the datasources doesn't yet work - haven't figured out where all of those options are coming from to meddle with the API endpoint.


application/classes/Ushahidi/Repository/Config.php
```diff
@@ -61,6 +61,7 @@ public function update(Entity $entity)
 	public function groups()
 	{
 		return [
+			'features',
```
rjmackay: So the problem is this enables writing as well as reading by an admin.. I'll have a think about the best solution

src/Core/Usecase/DataProvider/UtilizeDataProvider.php
```diff
@@ -0,0 +1,32 @@
...
+ * @copyright  2015 Ushahidi
+ * @license    https://www.gnu.org/licenses/agpl-3.0.html GNU Affero General Public License Version 3 (AGPL3)
+ */
+
+namespace Ushahidi\Core\Usecase\DataProvider;
+
+use Ushahidi\Core\Entity;
+use Ushahidi\Core\Usecase\UpdateUsecase;
+
+class UpdateDataProvider extends UpdateUsecase
```
rjmackay: Class name has to match file name.. otherwise this will fail to load

rjmackay: You'll also need to drop this into the DI set up so it gets used.

src/Core/Usecase/DataProvider/UpdateDataProvider.php
```diff
@@ -0,0 +1,32 @@
...
+	protected function verifyUpdateAuth(Entity $entity)
+	{
+
+		// feature
+		$entity_values = $entity->getDefinition();
+
+		// if not enabled
+		if (!feature($entity_values['name']))
+		{
+			throw HTTP_Exception::factory(403, 'This feature is not enabled for your deployment');
```
rjmackay: We should never be throwing Kohana exceptions from within the Usecases.. there are other exceptions you can use in src/Core/Exceptions.. or make a new one if needed.

src/Core/Usecase/DataProvider/UpdateDataProvider.php
```diff
@@ -0,0 +1,32 @@
...
+ * @copyright  2015 Ushahidi
+ * @license    https://www.gnu.org/licenses/agpl-3.0.html GNU Affero General Public License Version 3 (AGPL3)
+ */
+
+namespace Ushahidi\Core\Usecase\DataProvider;
+
+use Ushahidi\Core\Entity;
+use Ushahidi\Core\Usecase\UpdateUsecase;
+
+class UpdateDataProvider extends UpdateUsecase
```
rjmackay: We don't actually ever update data providers. We enable them in dataprovider config..

rjmackay:
Moved to https://github.com/ushahidi/platform/pull/684 .. thats based off a branch in this repo, not your fork. That way I can contribute too. Also cleaned up the history, and removed some stray old commits

Issue #638, opened 2015-08-23
Limit deployment features based on teir


rjmackay:
- API needs to disable some features if they’re disabled in the config
- Client should hide these features if they’re disabled
- Should be able to query config API and find out *which* features are disabled
- Feature limits will be set in a `feature` config

Current features that should be able to be limited are:
- Views: which views are available /enabled
- Data sources + notification channels ie. Email, SMSSync, etc
- User roles: starter plan only has a single admin, can create new members.
- CSV Upload fields: Can map to custom fields, can only map to title, description (date?)

rjmackay:
Reassigning to Will - lets see who gets to it first though :)

hollycorbett:
I'm back! On this now.

hollycorbett:
Okay @rjmackay, @zhalloran - thoughts/comments welcomed.  The config options thing is clear enough (figure I'll just put it in its own group_name in the config table) - I can get started on that while we debate :)

I can see three obvious potential ways of managing the backend:
1 - using the use-cases pattern
2 - using the acl pattern
3 - a separate piece of infrastructure 
	- a "filter" through which the API endpoints are passed and validated on a method-by-method basis... presumably run in before() of Api.php - after _parse_request but perhaps before "_check_access" (if not in it) - _check_deployment (or whatever is a good name?)

I think the third is best - mostly just cause it feels 'clearer' and like it should be a separate layer to the access.  But whichever :)

On fail, would return a 405 with 'invalid method' for the views, and for data/users/csv I guess a 405 with a more customized error message like 'your deployment does not allow creation/configuration of...'

SEEEWWWWW.... that would mean restricting:

Features

Views (is this all we mean):
 - map - /posts/:id/geojson
 - list - /posts/:id
 - chart - /posts/:id/stats
 - timeline - /posts/:id/stats
 (hrm... are chart and timeline separate features to control access to? obv we can change the display for the frontend but for now its the same data, just queried slightly differently - perhaps we should separate these into two different request paths as a precedent for future work if nothing else?)

Data Sources: here I think this will end up meaning changing the access to settings (i.e. unable to configure the sources)
 - /dataproviders/:id (OPTIONS?)
 - /config/:id (PUT) - where ID is 'data-provider'

Users
 - one admin v many and create new members - /users/:id (PUT) - would need parsed request to control access
 aside - curiously UserEndpoint.save() does not exist, but gets called in setting-users-create-controller.js

CSV Upload
[...]

Issue #640, opened 2015-08-23
Limit posts (and others?) on quota


rjmackay:
- Limit number of **published** posts
- Limit number of post types
- Limit number of admin users

rjmackay:
I think these could just be limits in the features config. Where `true` means unlimited while an integer is the limit.

- Need to ensure we return a sane response when attempting to create/publish resources beyond the quota. Not sure how/if we can do that well with our current validation tech.

PR #684, opened 2015-09-02
Add features into configuration again

rjmackay:
Part of ushahidi/platform#638

Todo:
- [x] Return 'available' or similar from the data provider API so we can handle that in the UI
- [x] Add tests, particularly for access control
- [x] Check my available data provider code actually works
- [x] Don't run disabled data providers (ie. from CLI tasks)

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform/684)

### DIFF
--- a/application/classes/Ushahidi/Repository/Config.php
+++ b/application/classes/Ushahidi/Repository/Config.php
@@ -61,6 +61,7 @@ class Ushahidi_Repository_Config implements
 	public function groups()
 	{
 		return [
+			'features',
 			'site',
 			'test',
 			'data-provider',
--- a/application/classes/Ushahidi/Repository/Config.php
+++ b/application/classes/Ushahidi/Repository/Config.php
@@ -26,7 +26,7 @@ class Ushahidi_Repository_Config implements
 	// ReadRepository
 	public function getEntity(Array $data = null)
 	{
-		return new ConfigEntity($data);
+	  return new ConfigEntity($data);
 	}
 
 	// ReadRepository
--- a/application/classes/Ushahidi/Rest.php
+++ b/application/classes/Ushahidi/Rest.php
@@ -353,7 +353,7 @@ abstract class Ushahidi_Rest extends Controller {
 			->allowed(array_keys($this->_action_map));
 		}
 
-		// Are we be expecting body content as part of the request?
+		// Are we the expecting body content as part of the request?
 		if (in_array($this->request->method(), $this->_methods_with_body_content))
 		{
 			$this->_parse_request_body();
@@ -404,7 +404,6 @@ abstract class Ushahidi_Rest extends Controller {
 	protected function _parse_request_body()
 	{
 			$payload = json_decode($this->request->body(), true);
-
 			// Ensure there were no JSON errors
 			$error = json_last_error();
 			if ($error AND $error !== JSON_ERROR_NONE)
--- a/application/classes/Ushahidi/Validator/Config/Update.php
+++ b/application/classes/Ushahidi/Validator/Config/Update.php
@@ -66,6 +66,15 @@ class Ushahidi_Validator_Config_Update extends Validator
 					]
 				];
 				break;
+
+			case 'data-provider':
+				$rules = [
+					'providers' => [
+						[[$this, 'isProviderAvailable'], [':value', ':validation']]
+					]
+				];
+				break;
+
 			default:
 				$rules = [];
 				break;
@@ -89,5 +98,16 @@ class Ushahidi_Validator_Config_Update extends Validator
 		}
 
 		return false;
+	}
+
+	public function isProviderAvailable($enabled_providers, $validation)
+	{
+		$enabled_providers = array_filter($enabled_providers);
+		$available_providers = array_filter(\Kohana::$config->load('features.data-providers')->as_array());
+
+		$diff = array_diff_key($enabled_providers, $available_providers);
+		if ($diff) {
+			$validation->error('providers', 'providerNotAvailable', [$diff]);
+		}
 	}
 }
--- a/application/classes/Ushahidi/Validator/Config/Update.php
+++ b/application/classes/Ushahidi/Validator/Config/Update.php
@@ -102,14 +102,14 @@ class Ushahidi_Validator_Config_Update extends Validator
 
 	public function isProviderAvailable($enabled_providers, $validation)
 	{
-    if ($enabled_providers != null) {
-  		$enabled_providers = array_filter($enabled_providers);
-	  	$available_providers = array_filter(\Kohana::$config->load('features.data-providers')->as_array());
+		if ($enabled_providers != null) {
+			$enabled_providers = array_filter($enabled_providers);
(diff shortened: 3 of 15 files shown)


rjmackay:
Something seems to have gone horribly broken and its trying to write config to a file, not the db.. very confused

rjmackay:
@hollycorbett added some todos, feel free to pick this up again while I sleep. Or just work on update the client to match these changes.

rjmackay:
Once todos on this are done, we can land this. Work for quota based limits (#640) should happen on a follow up diff.

willdoran:
@rjmackay I've added one method for preventing unavailable data providers from being used. This will work for the twitter and email provider as well. However, I wanted to see if it was better instead to prevent the data providers at an earlier point by preventing their registration here:

https://github.com/ushahidi/platform/blob/08d2bcd8a19e469a7d11618f51c20a3da4bb2799/modules/data-provider/classes/DataProvider/Core.php#L100

What do you think? 

If we did do that then we would have to call registration again when the tier is up/down graded, that seems like a simpler option and a more simple control point as the data providers should be agnostic of whether they are available or not.

rjmackay:
@willdoran I don't think we can limit the providers earlier or we wouldn't be able to return disabled/unavailable providers in the API

willdoran:
Ok, cool. That works for me.

rjmackay:
Reviewed 2 of 6 files at r2, 2 of 3 files at r3, 8 of 8 files at r5, 1 of 1 files at r6, 2 of 2 files at r7.
Review status: all files reviewed at latest revision, 7 unresolved discussions, all commit checks successful.

---

<sup>**[application/classes/Ushahidi/Repository/Config.php, line 40 \[r7\]](https://reviewable.io:443/reviews/ushahidi/platform/684#-JzNHdid3XAkcBmCySl0)** ([raw file](https://github.com/ushahidi/platform/blob/078e9a5390e6c7a90e5d7e62a600deb87a7c7ef7/application/classes/Ushahidi/Repository/Config.php#L40)):</sup>
Why are we loading available features for every config group?
This should only happen with $group = features ...

---

<sup>**[application/classes/Ushahidi/Validator/Config/Update.php, line 105 \[r7\]](https://reviewable.io:443/reviews/ushahidi/platform/684#-JzNHscoW60ZfL5fJIhz)** ([raw file](https://github.com/ushahidi/platform/blob/078e9a5390e6c7a90e5d7e62a600deb87a7c7ef7/application/classes/Ushahidi/Validator/Config/Update.php#L105)):</sup>
indenting is borked

---

<sup>**[plugins/data-provider-email/classes/DataProvider/Email.php, line 28 \[r7\]](https://reviewable.io:443/reviews/ushahidi/platform/684#-JzNI9kTBPh48Gtl1TY-)** ([raw file](https://github.com/ushahidi/platform/blob/078e9a5390e6c7a90e5d7e62a600deb87a7c7ef7/plugins/data-provider-email/classes/DataProvider/Email.php#L28)):</sup>
indenting?

---

<sup>**[plugins/data-provider-email/classes/DataProvider/Email.php, line 88 \[r7\]](https://reviewable.io:443/reviews/ushahidi/platform/684#-JzNI_FGd2-kvfgOKzR-)** ([raw file](https://github.com/ushahidi/platform/blob/078e9a5390e6c7a90e5d7e62a600deb87a7c7ef7/plugins/data-provider-email/classes/DataProvider/Email.php#L88)):</sup>
Is there a way to do a general check before fetch() or send() is even called? Or just move these checks to the parent class, then call parent::fetch() first.

---

<sup>**[plugins/data-provider-email/classes/DataProvider/Email.php, line 90 \[r7\]](https://reviewable.io:443/reviews/ushahidi/platform/684#-JzNI8VlSb8HgcM9qrrw)** ([raw file](https://github.com/ushahidi/platform/blob/078e9a5390e6c7a90e5d7e62a600deb87a7c7ef7/plugins/data-provider-email/classes/DataProvider/Email.php#L90)):</sup>
indenting?

---

<sup>**[plugins/data-provider-email/classes/DataProvider/Email.php, line 161 \[r7\]](https://reviewable.io:443/reviews/ushahidi/platform/684#-JzNILb-dejKTJtuyRji)** ([raw file](https://github.com/ushahidi/platform/blob/078e9a5390e6c7a90e5d7e62a600deb87a7c7ef7/plugins/data-provider-email/classes/DataProvider/Email.php#L161)):</sup>
indent 

---

<sup>**[src/Core/Tool/Authorizer/ConfigAuthorizer.php, line 54 \[r7\]](https://reviewable.io:443/reviews/ushahidi/platform/684#-JzNItFvdZJZI3MxrOnb)** ([raw file](https://github.com/ushahidi/platform/blob/078e9a5390e6c7a90e5d7e62a600deb87a7c7ef7/src/Core/Tool/Authorizer/ConfigAuthorizer.php#L54)):</sup>
did you or I add test for this?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/684)

willdoran:
Review status: all files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[application/classes/Ushahidi/Repository/Config.php, line 40 \[r7\]](https://reviewable.io:443/reviews/ushahidi/platform/684#-JzNHdid3XAkcBmCySl0)** ([raw file](https://github.com/ushahidi/platform/blob/078e9a5390e6c7a90e5d7e62a600deb87a7c7ef7/application/classes/Ushahidi/Repository/Config.php#L40)):</sup>
Yep, I've changed this to simply use the features api as you suggested.

---

<sup>**[plugins/data-provider-email/classes/DataProvider/Email.php, line 88 \[r7\]](https://reviewable.io:443/reviews/ushahidi/platform/684#-JzNI_FGd2-kvfgOKzR-)** ([raw file](https://github.com/ushahidi/platform/blob/078e9a5390e6c7a90e5d7e62a600deb87a7c7ef7/plugins/data-provider-email/classes/DataProvider/Email.php#L88)):</sup>
I think moving these to the parent is probably the most effective. I'll look again to see if I can intercept before fetch and send

---

<sup>**[src/Core/Tool/Authorizer/ConfigAuthorizer.php, line 54 \[r7\]](https://reviewable.io:443/reviews/ushahidi/platform/684#-JzNItFvdZJZI3MxrOnb)** ([raw file](https://github.com/ushahidi/platform/blob/078e9a5390e6c7a90e5d7e62a600deb87a7c7ef7/src/Core/Tool/Authorizer/ConfigAuthorizer.php#L54)):</sup>
I don't think I worked on this file.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/684)

rjmackay:
Looking pretty good. I'd like some tests for the access control on config, ie. to ensure admins can't change their own feature config via the API. But we should land and create a follow up task for that

---

Review status: 13 of 15 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/684)

PR #688, opened 2015-09-03
Add deployment owner role

### LINKED ISSUE #421: Add deployment owner role who can manage subscription etc
- Add an `owner` role
- Role cannot be assigned to users through normal UI
  - Need to limit this in the API too.
  - ~~Probably something like `ClientContext` in the Authorizer to check for a special client with access to create owners~~
- Role gets extra access .. but that doesn't matter for this ticket

*Created by @rjmackay on 2015-07-17 08:25:21.*

*Imported from https://phabricator.ushahidi.com/T1704*

jasonmule:
- Do not create user accounts with Owner role through the API
- To run use:
`bin/ushahidi role create -r owner`

closes #421

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform/688)

### DIFF
--- a/src/Console/Role.php
+++ b/src/Console/Role.php
@@ -31,7 +31,7 @@ class Role extends Command
 			->setName('role')
 			->setDescription('Create roles')
 			->addArgument('action', InputArgument::OPTIONAL, 'list, create', 'list')
-			->addOption('role', ['r'], InputOption::VALUE_OPTIONAL, 'role')
+			->addOption('role', ['r'], InputOption::VALUE_REQUIRED, 'role')
 			;
 	}
 
--- a/src/Console/Role.php
+++ b/src/Console/Role.php
@@ -1,0 +1,69 @@
+<?php
+
+/**
+ * Ushahidi Role Console Command
+ *
+ * @author     Ushahidi Team <team@ushahidi.com>
+ * @package    Ushahidi\Console
+ * @copyright  2014 Ushahidi
+ * @license    https://www.gnu.org/licenses/agpl-3.0.html GNU Affero General Public License Version 3 (AGPL3)
+ */
+
+namespace Ushahidi\Console;
+
+use Ushahidi\Core\Entity\RoleRepository;
+
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class Role extends Command
+{
+	public function setRepo(RoleRepository $repo)
+	{
+		$this->repo = $repo;
+	}
+
+	protected function configure()
+	{
+		$this
+			->setName('role')
+			->setDescription('Create roles')
+			->addArgument('action', InputArgument::OPTIONAL, 'list, create', 'list')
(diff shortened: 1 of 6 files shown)


rjmackay:
Hey - so I wasn't really expecting a CLI task for adding owner.. though this *might* work.
For Cloud 
- we need to make sure the 'Owner' role can't be assigned to any users in the API (ie. should only be able to be assigned when we create the first user from the CLI).
- the owner role also needs to gain the same permissions as an admin user, so it'll need to be added into the AdminAccess Trait

jasonmule:
Noted. So I expected to use the CLI to create the owner role once. Going forward no one should be able to assign this role through the API.

rjmackay:
> Going forward no one should be able to assign this role through the API.
Not sure if you were saying thats what currently happen? or you'll update?

Currently:
- Once its in the role table, the user validator will allow creating users with the role.
- And this role will have no permissions beyond a standard user.. it needs to be equivalent to admin

jasonmule:
Currently, this role cannot be update/ created by a normal user through the API. I'll add to the AdminAccess Trait as well.

src/Console/Role.php
```diff
@@ -0,0 +1,69 @@
...
+		$this->repo = $repo;
+	}
+
+	protected function configure()
+	{
+		$this
+			->setName('role')
+			->setDescription('Create roles')
+			->addArgument('action', InputArgument::OPTIONAL, 'list, create', 'list')
+			->addOption('role', ['r'], InputOption::VALUE_OPTIONAL, 'role')
```
rjmackay: I think this should be required

src/Core/Tool/Authorizer/UserAuthorizer.php
```diff
@@ -43,6 +43,12 @@ protected function getAllPrivs()
 	/* Authorizer */
 	public function isAllowed(Entity $entity, $privilege)
 	{
+		// Users cannot create 'owner' role
+		if (in_array($privilege, ['register', 'create', 'update'])
+			and $entity->role === 'owner') {
```
rjmackay: Problem here is that an admin could edit the owner user and change their role to 'admin' .. I'm not sure how we avoid it. I gotta think about this a little

src/Core/Tool/Authorizer/UserAuthorizer.php
```diff
@@ -79,4 +85,15 @@ public function isAllowed(Entity $entity, $privilege)
 		// If no other access checks succeed, we default to denying access
 		return false;
 	}
+
+	/**
+	 * Checks if the entity has an 'owner' role
+	 * @param User $entity
+	 * @return boolean
+	 */
+	protected function userHasOwnerRole(User $entity)
```
rjmackay: unused?

rjmackay:
Hacked this a bit myself. There are some issues.. and it bugs me that we create a role through the CLI, but have to hard code the permissions it has in the authorizer.

src/Core/Tool/Authorizer/UserAuthorizer.php
jasonmule: Yeah. This is going to be a problem to implement if admin is not allowed to do that...

jasonmule:
Agreed. At first I thought that this was going to be convenient for doing this as a one-time task.

rjmackay:
As discussed on the call lets shelve this for now

PR #689, opened 2015-09-04
Add default post type via migrations

rjmackay:
Adds a default post type with a single location field

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform/689)

### DIFF
--- a/migrations/20150904021011_add_default_post_type.php
+++ b/migrations/20150904021011_add_default_post_type.php
@@ -1,0 +1,49 @@
+<?php
+
+use Phinx\Migration\AbstractMigration;
+
+class AddDefaultPostType extends AbstractMigration
+{
+
+    /**
+     * Migrate Up.
+     */
+    public function up()
+    {
+        $connection = $this->getAdapter()->getConnection();
+        $adapter = $this->getAdapter();
+
+        // Create default post type
+        // Note: Timestamp here *may* be in a different timezone, since
+        // we're using mysql time, not PHP time.
+        $this->execute(
+            "INSERT INTO forms (name, description, type, created)
+            VALUES
+                ( 'Basic Post', 'Post with a location', 'report', UNIX_TIMESTAMP(NOW()) )
+            "
+        );
+
+        $connection->prepare(
+            "INSERT INTO form_stages (label, priority, required, form_id)
+            VALUES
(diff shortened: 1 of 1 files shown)


rjmackay:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

<sup>**[migrations/20150904021011_add_default_post_type.php, line 15 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/689#-JyLCxr4Vjubnf3xMgcd)** ([raw file](https://github.com/ushahidi/platform/blob/de3364e673e89ba8695ea939cba3e997890dd9e7/migrations/20150904021011_add_default_post_type.php#L15)):</sup>
I'm thinking I should check that either:
- we have no post types 
- or at least don't already have a "Basic Post" post type
before creating anything.
@willdoran thoughts?

---

<sup>**[migrations/20150904021011_add_default_post_type.php, line 47 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/689#-JyLCljtY-4F5ZTZGSEy)** ([raw file](https://github.com/ushahidi/platform/blob/de3364e673e89ba8695ea939cba3e997890dd9e7/migrations/20150904021011_add_default_post_type.php#L47)):</sup>
@willdoran curious what you think here?
Migrating up and down will lead to many many post types..
However I don't want to accidentally delete a bunch of posts by deleting the post type.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/689)

rjmackay:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[migrations/20150904021011_add_default_post_type.php, line 48 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/689#-JyLCljtY-4F5ZTZGSEy)** ([raw file](https://github.com/ushahidi/platform/blob/de3364e673e89ba8695ea939cba3e997890dd9e7/migrations/20150904021011_add_default_post_type.php#L48)):</sup>
Checking for post types first.. so at least we don't end up with lots of default post types created. I think this works

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/689)

PR #707, opened 2015-09-17
Add superpowers flag to give some oauth clients access to set config

rjmackay:
Allows an oauth client to set config without a user session. The usecase is allowing our cloud management jobs to push config to deployments.

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform/707)

### DIFF
--- a/application/classes/Ushahidi/Console/Oauth/Client.php
+++ b/application/classes/Ushahidi/Console/Oauth/Client.php
@@ -62,6 +62,7 @@ class Ushahidi_Console_OAuth_Client extends Command {
 			->addOption('client', ['c'], InputOption::VALUE_OPTIONAL, 'client id')
 			->addOption('name', [], InputOption::VALUE_OPTIONAL, 'client name')
 			->addOption('secret', ['s'], InputOption::VALUE_OPTIONAL, 'secret key')
+			->addOption('superpowers', [], InputOption::VALUE_NONE, 'has super powers?')
 			;
 	}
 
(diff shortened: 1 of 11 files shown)


rjmackay:
@willdoran @jasonmule Can you give this a review?
I've rushed a bit with this.. hopefully your first thought isn't "whoa thats a bad idea".. but if it is I want to know :)

jasonmule:
Not really comfortable about having an admin and a super powered user with a back door on the same platform..

rjmackay:
So to clarify this doesn't add a 'superpower' user. Its a super powered oauth client. ie. no user login, just oauth client + secret.
I debated the naming a little.. the permissions are slightly different: 
- Superpowered client can change any config setting, including which features are enabled (used to power tiers in cloud)
- Admins can only edit site and data provider config..

.. you do make a good point though. Its feels like this is asking for confusion later.  Any thoughts on a better permissions model? or just better naming?

Some of it might get cleared up when we build custom roles, since we could split things out into 'permissions'.. but need a good solution now too.

jasonmule:
Understood. Maybe call it config manager since we are using it to only change configs? :-)

rjmackay:
That definitely a more useful description. The scope might expand later. ie. I can imagine giving access to modify users in future. Maybe `deployment manager` or just `manager`.. but that sounds too much like `admin` again.

The other option, is to avoid this entirely and set config through a CLI task ie. `bin/ushahidi config:set` .. I started down that road but got stuck.

rjmackay:
While we figure this out.. heres a PR for setting config via the console instead https://github.com/ushahidi/platform/pull/709

rjmackay:
Closing this PR (but leaving the branch). We can re-add later if needed

PR #756, opened 2015-10-06
Discover and queue posts from Saved Searches

jasonmule:
Queue notifications:
```bin/ushahidi notification queue```

Find posts from saved searches
```bin/ushahidi savedsearch```

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform/756)

### CODE REVIEW

**Comment by jasonmule** [comment]
@willdoran If you get a chance could you review this? It's a CLI task that checks for new posts for saved searches.

**Comment by rjmackay** [comment]
Related to this.. why do we have `notification_queue` and `posts_sets` in the DB.. are they essentially the same thing? or do they really need to be recorded separately?

File: application/classes/Ushahidi/Console/Notification/Collection.php [hunk 1]
```diff
@@ -45,8 +52,8 @@ public function setNotificationQueueRepo(NotificationQueueRepository $repo)
 	protected function configure()
 	{
 		$this
-			->setName('notification')
-			->setDescription('Manage notifications')
+			->setName('notification:collection')
+			->setDescription('Manage notifications for Collections')
 			->addArgument('action', InputArgument::OPTIONAL, 'list, queue', 'list')
```
rjmackay: I'm leaning towards removing the `action` argument from all our tasks. You can just override `execute` entirely instead..
I've done that for `config:set` .. what do you think?

src/Core/Entity/ContactRepository.php
```diff
@@ -25,4 +25,6 @@
 	 * @return boolean
 	 */
 	public function getByContact($contact, $type);
+
+	public function getNotificationContacts($set_id = 0, $limit = false, $offset = 0);
```
rjmackay: Add a docblock please :)

application/classes/Ushahidi/Repository/Contact.php
```diff
@@ -124,4 +124,32 @@ public function getByContact($contact, $type)
...
+	public function getNotificationContacts($set_id = 0, $limit = false, $offset = 0)
+	{
+		$query = DB::select('contacts.id', 'contacts.type', 'contacts.contact')
+			->distinct(TRUE)
+			->from('contacts')
+			->join('notifications')
+			->on('contacts.user_id', '=', 'notifications.user_id')
+			->where('contacts.can_notify', '=', '1');
+
+		if ($set_id) {
```
rjmackay: Does it ever make sense to call this without a set_id?
Also wonder if it might make more sense to be passing in the notification id instead..

application/classes/Ushahidi/Console/Notification/Collection.php
```diff
@@ -91,7 +98,7 @@ protected function executeQueue(InputInterface $input, OutputInterface $output)
 		];
 	}
 
-	private function _generate_messages($notification)
+	private function generateMessages($notification)
```
rjmackay: This should move into a Usecase eventually. **not now** but add a tech debt task to track it

application/classes/Ushahidi/Console/Notification/Collection.php
```diff
@@ -78,7 +85,7 @@ protected function executeQueue(InputInterface $input, OutputInterface $output)
 
 		foreach ($notifications as $notification) {
```
rjmackay: This should probably be in the usecase too (ref next comment)

application/classes/Ushahidi/Console/Notification/SavedSearch.php
```diff
@@ -0,0 +1,180 @@
...
+		$lastSearchTimestamp = $config->get('last_search_timestamp') ?: 0;
+		$lastSearchTimestamp = date('Y-m-d H:i:s', $lastSearchTimestamp);
+
+		// Start transaction
+		$this->db->begin();
+
+		foreach ($savedSearches as $savedSearch) {
+			// Get posts with the search filter
+			$search = service('factory.data')->get('search');
+			$search->q = $savedSearch->filter['q'];
```
rjmackay: This ignores all the other possible params. ALL the parts of `filter` should be passed straight through.

application/classes/Ushahidi/Console/Notification/SavedSearch.php
```diff
@@ -0,0 +1,180 @@
...
+			$search->created_after = $lastSearchTimestamp;
```
rjmackay: We talked about this a little while ago. This doesn't work since a post is likely to be created, then given a category (or some other change) later which makes it match a search.

- You might be able to use `updated_after` instead.. but I think that might end up sending dupe alerts
- I think we probably need to save the the result of a saved search, then on future runs we diff against the previous set of posts and newly added posts. Does that make sense?
We could use the posts_sets table to record this for saved searches I think. However there might be some scale issues.. :/

application/classes/Ushahidi/Console/Notification/SavedSearch.php
```diff
@@ -0,0 +1,180 @@
...
+
+		// @todo Might need to limit the number of saved searches retrieved at a time
+		$savedSearches = $this->setRepository->getSearchResults();
+
+		$config = Kohana::$config->load('saved_search');
+		$lastSearchTimestamp = $config->get('last_search_timestamp') ?: 0;
+		$lastSearchTimestamp = date('Y-m-d H:i:s', $lastSearchTimestamp);
+
+		// Start transaction
+		$this->db->begin();
```
rjmackay: why the transaction around this?

application/classes/Ushahidi/Console/Notification/SavedSearch.php
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	protected function executeQueue(InputInterface $input, OutputInterface $output)
+	{
+		$count = 0;
+
+		// Search data
+		$search = service('factory.data')->get('search');
+		
+		$this->db = service('kohana.db');
```
rjmackay: Need to inject this. Though ideally we shouldn't be talking directly to the DB from a console command at all

application/classes/Ushahidi/Console/Notification/SavedSearch.php
```diff
@@ -0,0 +1,180 @@
...
+			]
+		];
+	}
+
+	protected function executeQueue(InputInterface $input, OutputInterface $output)
+	{
+		$count = 0;
+
+		// Search data
+		$search = service('factory.data')->get('search');
```
rjmackay: Inject this too?

application/classes/Ushahidi/Console/Notification/SavedSearch.php
```diff
@@ -0,0 +1,180 @@
...
+		$this->db->commit();
+
+		return [
+			[
+				'Message' => sprintf('%d messages queued for sending', $count)
+			]
+		];
+	}
+
+	private function generateMessages($savedSearch, $postCount)
```
rjmackay: is this mostly copy/paste from notifications:collection ? Could you share it via a base class or trait?

jasonmule:
<sup>**[application/classes/Ushahidi/Console/Notification/SavedSearch.php, line 97 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/756#-K0XcvWAMaAtZzSwVhhM-r1-97)** ([raw file](https://github.com/ushahidi/platform/blob/6ddeeefa4d908b55178fa8d4b9dd7e8f0c5e6067/application/classes/Ushahidi/Console/Notification/SavedSearch.php#L97)):</sup>
For saved searches, I switched this to just sending the count of messages instead of sending the actual posts as messages, so it could be as simple as sending the count diff between searches?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/756)

jasonmule:
`notification_queue` acts as a FIFO queue and the entries are deleted. The plan was to eventually move this to an actual queue instead of saving to the DB.


---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/756)

rjmackay:
Makes sense. Laravel has a queue with adapters for DB, or real queues.. we can probably move to something like that eventually :)


---

Review status: 0 of 7 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

<sup>**[application/classes/Ushahidi/Console/Notification/SavedSearch.php, line 97 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/756#-K0XcvWAMaAtZzSwVhhM-r1-97)** ([raw file](https://github.com/ushahidi/platform/blob/6ddeeefa4d908b55178fa8d4b9dd7e8f0c5e6067/application/classes/Ushahidi/Console/Notification/SavedSearch.php#L97)):</sup>
I think we need to be sending the actual messages, not just a count. Otherwise theres no easy way for a user to see what changed in their search.

@jshorland can you add your thinking / confirm I'm not making this overcomplicated?
For saved searches to work for managing workflows, we need to be able to send notifications whenever a post is changed and now matches that search. Since a post might change tags, or current stage after its created we can't rely on just sending notifications for 'new' posts which match a search, but have to regularly check if existing posts now match. Does that make sense?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/756)

rjmackay:
Reviewed 8 of 8 files at r1.
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/756)

jasonmule:
<sup>**[application/classes/Ushahidi/Console/Notification/SavedSearch.php, line 97 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/756#-K0XcvWAMaAtZzSwVhhM-r1-97)** ([raw file](https://github.com/ushahidi/platform/blob/6ddeeefa4d908b55178fa8d4b9dd7e8f0c5e6067/application/classes/Ushahidi/Console/Notification/SavedSearch.php#L97)):</sup>
Agreed that we should not rely on the post creation time but if we send out the post content instead, do we limit the number of posts sent out in a notification? E.g. what happens when a saved search has 50 new posts.  This is why I was thinking of sending out the count + url to the saved search as the notification message.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/756)

application/classes/Ushahidi/Console/Notification/Collection.php
jasonmule: It makes more sense to override `execute`

application/classes/Ushahidi/Repository/Contact.php
jasonmule: So that you can fetch all contacts that can be notified but also pass a set_id to filter for multiple contacts subscribed to that set.

application/classes/Ushahidi/Console/Notification/Collection.php
jasonmule: Yup. Added this before.

application/classes/Ushahidi/Console/Notification/SavedSearch.php
jasonmule: Using posts_sets should simplify this for now. Also, not sure about running a diff as adding a new post to a saved search will give you the same behaviour as adding a new post to a collection? If the post doesn't exist, it's added a notification queued.

application/classes/Ushahidi/Repository/Contact.php
rjmackay: Right but its never used *without* the set_id .. could we just make that param required?

rjmackay:
Reviewed 9 of 10 files at r2.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/756)

rjmackay:
Reviewed 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

<sup>**[application/classes/Ushahidi/Console/SavedSearch.php, line 90 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform/756#-K2JnEQAatOgpxH-VXpy)** ([raw file](https://github.com/ushahidi/platform/blob/158cca3158b79bf7a0fcb1bdee97bc20d349eec8/application/classes/Ushahidi/Console/SavedSearch.php#L90)):</sup>
Am I right that this will automatically get queue for notifications when added?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/756)

rjmackay:
I think this is probably good to go now? Maybe get Jess to test once its pushed to the dev site?


---

Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/756)

jasonmule:
<sup>**[application/classes/Ushahidi/Console/SavedSearch.php, line 90 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform/756#-K2JnEQAatOgpxH-VXpy)** ([raw file](https://github.com/ushahidi/platform/blob/158cca3158b79bf7a0fcb1bdee97bc20d349eec8/application/classes/Ushahidi/Console/SavedSearch.php#L90)):</sup>
Yes. An event is triggered when this is added.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/756)

jasonmule:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

<sup>**[application/classes/Ushahidi/Console/SavedSearch.php, line 90 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform/756#-K2JnEQAatOgpxH-VXpy)** ([raw file](https://github.com/ushahidi/platform/blob/158cca3158b79bf7a0fcb1bdee97bc20d349eec8/application/classes/Ushahidi/Console/SavedSearch.php#L90)):</sup>
Done.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/756)

rjmackay:
:+1: land it.


---

Reviewed 2 of 2 files at r4.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/756)

PR #757, opened 2015-10-09
Add twitter to config groups

jasonmule:
We need this to load twitter configs from the database.

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform/757)


rjmackay:
I don't know if this makes sense. This only effects loading twitter params into the API, and I think we're only using that group for values like the last twitter message id.
Why did we need to make this change?

jasonmule:
Don't we still need to allow the twitter group through the config repo when we attach the database config in `attached_db_config()` so that we can load previously saved keys?

PR #784, opened 2015-10-20
#668 list all form attributes and stages

### LINKED ISSUE #668: Add endpoints to fetch ALL attributes and stages

jasonmule:
Fixes #668

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform/784)


jasonmule:
@rjmackay When you get a chance, could you take a quick look?

application/classes/Ushahidi/Formatter/Formattribute.php
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @author     Ushahidi Team <team@ushahidi.com>
+ * @package    Ushahidi\Application
+ * @copyright  2014 Ushahidi
+ * @license    https://www.gnu.org/licenses/agpl-3.0.html GNU Affero General Public License Version 3 (AGPL3)
+ */
+
+use Ushahidi\Core\Traits\FormatterAuthorizerMetadata;
+
+class Ushahidi_Formatter_Formattribute extends Ushahidi_Formatter_API
```
rjmackay: Should use CamelCase on the `FormAttribute` part .. and the file name.

application/classes/Ushahidi/Formatter/Formstage.php
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @author     Ushahidi Team <team@ushahidi.com>
+ * @package    Ushahidi\Application
+ * @copyright  2014 Ushahidi
+ * @license    https://www.gnu.org/licenses/agpl-3.0.html GNU Affero General Public License Version 3 (AGPL3)
+ */
+
+use Ushahidi\Core\Traits\FormatterAuthorizerMetadata;
+
+class Ushahidi_Formatter_Formstage extends Ushahidi_Formatter_API
```
rjmackay: Same here re: Formstage -> FormStage

application/routes/default.php
```diff
@@ -215,6 +203,20 @@
 	));
 
 /**
+ * Forms API SubRoute
+ */
+Route::set('forms', $apiBase . 'forms(/<form_id>)/<controller>(/<id>)',
```
rjmackay: Why did this change at all?

rjmackay: nm. I see the change. I'm not 100% why this moved down the file though? vs just making the 1 line change.

src/Core/Usecase/Form/VerifyFormLoaded.php
```diff
@@ -45,7 +45,10 @@ protected function verifyFormExists()
 	// Usecase
 	public function interact()
 	{
-		$this->verifyFormExists();
+		if ($this->getRequiredIdentifier('form_id')) {
```
rjmackay: Doesn't `getRequiredIdentifier` throw an error if `form_id` hasn't been set?
I think you just need `getIdentifier` ?

rjmackay: This trait is used in many places: i.e create / edit / delete form stages .. the form_id should only be optional for SearchFormStage and SearchFormAttribute Usecases. Not sure what the best way to handle that is? Maybe just override `verifyFormExists` on those Usecases instead?

application/routes/default.php
jasonmule: so that the route above is matched first for requests like api/v3/forms/1

rjmackay:
Good to land. Minor tweak to improve the usecase.. but if test pass ship it


---

Reviewed 2 of 10 files at r1, 1 of 7 files at r2, 5 of 5 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

<sup>**[src/Core/Usecase/Form/SearchFormAttribute.php, line 31 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform/784#-K2paHljbD8n6-d2C_HO)** ([raw file](https://github.com/ushahidi/platform/blob/3a14a223bb7655efa77f45ad25677c49d5fb0186/src/Core/Usecase/Form/SearchFormAttribute.php#L31)):</sup>
I think we still need to verify it *if* an ID is passed..
Something like
```
// Ensure that the form exists.
if ($form = $this->form_repo->get($this->getIdentifier('form_id'))) {
		$this->verifyEntityLoaded($form, $this->identifiers);
}
```


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/784)

Issue #501, opened 2015-08-23
User password can be bruteforced easily through oauth password auth


ushbot:
Because we expose a client and secret (ushahidiui client) with access to the password auth type, it would be trivial to use these detail to dictionary or brute force attack user passwords.

Add rate limiting oauth/token requests (or check if phpleague/oauth already does that?)
 Look into any other mitigation

From the oauth spec:

> Since this client authentication method involves a password, the
>  authorization server MUST protect any endpoint utilizing it against
>  brute force attacks.

*Created by @rjmackay on 2014-11-26 16:30:01.*

*Imported from https://phabricator.ushahidi.com/T1188*

PR #807, opened 2015-11-02
Add rate limiting for login attempts

jasonmule:
#501

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform/807)

### DIFF
--- a/application/classes/Ushahidi/Core.php
+++ b/application/classes/Ushahidi/Core.php
@@ -553,7 +553,7 @@ abstract class Ushahidi_Core {
 		$di->set('ratelimiter.login.flap', $di->lazyNew('BehEh\Flaps\Flap'));
 
 		$di->params['BehEh\Flaps\Flap'] = [
-			'storage' => $di->lazyGet('ratelimiter.storage'),
+			'storage' => $di->lazyNew('BehEh\Flaps\Storage\DoctrineCacheAdapter'),
 			'name' => 'login'
 		];
 
@@ -572,15 +572,12 @@ abstract class Ushahidi_Core {
 			'throttlingStrategy' => $di->lazyGet('ratelimiter.login.strategy'),
 		];
 
-		// Rate limiter storage
-		$di->set('ratelimiter.storage', $di->lazyNew('BehEh\Flaps\Storage\DoctrineCacheAdapter'));
-
 		$di->params['BehEh\Flaps\Storage\DoctrineCacheAdapter'] = [
-			'cache' => $di->lazyGet('ratelimiter.storage.cache')
+			'cache' => $di->lazyGet('ratelimiter.cache')
 		];
 
 		// Rate limit storage cache
-		$di->set('ratelimiter.storage.cache', function() use ($di) {
+		$di->set('ratelimiter.cache', function() use ($di) {
 			$cache = $di->lazyGet('ratelimiter.config.cache');
 
 			if ($cache === 'memcache') {
--- a/application/classes/Ushahidi/Core.php
+++ b/application/classes/Ushahidi/Core.php
@@ -553,8 +553,9 @@ abstract class Ushahidi_Core {
 		// Rate limiter storage
 		$di->set('ratelimiter.storage', $di->lazyNew('BehEh\Flaps\Storage\DoctrineCacheAdapter'));
 
+		// @todo should use persistent storage in production
 		$di->params['BehEh\Flaps\Storage\DoctrineCacheAdapter'] = [
-			'cache' => $di->lazyNew('Doctrine\Common\Cache\ApcCache')
+			'cache' => $di->lazyNew('Doctrine\Common\Cache\ArrayCache')
 		];
 
 		// Rate limiter violation handler
--- a/application/classes/Ushahidi/Core.php
+++ b/application/classes/Ushahidi/Core.php
@@ -542,25 +542,13 @@ abstract class Ushahidi_Core {
 		$di->setter['Ushahidi_Listener_PostSetListener']['setRepo'] =
 			$di->lazyGet('repository.notification.queue');
 
-		// Login rate limiter
-		$di->set('ratelimiter.login', $di->lazyNew('BehEh\Flaps\Flap'));
+		// Set up login rate limiter
+		$di->set('ratelimiter.login.flap', $di->lazyNew('BehEh\Flaps\Flap'));
 
 		$di->params['BehEh\Flaps\Flap'] = [
 			'storage' => $di->lazyGet('ratelimiter.storage'),
 			'name' => 'login'
 		];
-
-		// Rate limiter storage
-		$di->set('ratelimiter.storage', $di->lazyNew('BehEh\Flaps\Storage\DoctrineCacheAdapter'));
-
-		// @todo should use persistent storage in production
-		$di->params['BehEh\Flaps\Storage\DoctrineCacheAdapter'] = [
-			'cache' => $di->lazyNew('Doctrine\Common\Cache\ArrayCache')
-		];
-
-		// Rate limiter violation handler
-		$di->setter['BehEh\Flaps\Flap']['setViolationHandler'] =
-			$di->lazyNew('BehEh\Flaps\Violation\ExceptionViolationHandler');
 
 		$di->set('ratelimiter.login.strategy', $di->lazyNew('BehEh\Flaps\Throttling\LeakyBucketStrategy'));
 
@@ -569,6 +557,46 @@ abstract class Ushahidi_Core {
 			'requests' => 3,
 			'timeSpan' => '1m'
 		];
+
+		$di->set('ratelimiter.login', $di->lazyNew('Ushahidi_RateLimiter'));
+
+		$di->params['Ushahidi_RateLimiter'] = [
+			'flap' => $di->lazyGet('ratelimiter.login.flap'),
+			'throttlingStrategy' => $di->lazyGet('ratelimiter.login.strategy'),
+		];
+
+		// Rate limiter storage
+		$di->set('ratelimiter.storage', $di->lazyNew('BehEh\Flaps\Storage\DoctrineCacheAdapter'));
+
+		$di->params['BehEh\Flaps\Storage\DoctrineCacheAdapter'] = [
+			'cache' => $di->lazyGet('ratelimiter.storage.cache')
+		];
+
+		// Rate limit storage cache
+		$di->set('ratelimiter.storage.cache', function() use ($di) {
+			$cache = Kohana::$config->load('ratelimiter.cache');
+			
+			if ($cache === 'memcache') {
+				$di->setter['Doctrine\Common\Cache\MemcachedCache']['setMemcached'] =
+					$di->LazyNew('\Memcached');
+
+				return $di->newInstance('Doctrine\Common\Cache\MemcachedCache');
+			}
+			elseif ($cache === 'filesystem') {
+				$di->params['Doctrine\Common\Cache\FilesystemCache'] = [
+					'directory' => Kohana::$config->load('ratelimiter.filesystem.directory'),
+				];
+
+				return $di->newInstance('Doctrine\Common\Cache\FilesystemCache');
+			}
+
+			// Use in-memory cache if none is configured
+			return $di->newInstance('Doctrine\Common\Cache\ArrayCache');
+		});
+
+		// Rate limiter violation handler
+		$di->setter['BehEh\Flaps\Flap']['setViolationHandler'] =
+			$di->lazyNew('Ushahidi_ThrottlingViolationHandler');
 
 		/**
 		 * 1. Load the plugins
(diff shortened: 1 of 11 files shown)


src/Core/Usecase/User/LoginUser.php
```diff
@@ -33,12 +46,32 @@ public function setAuthenticator(PasswordAuthenticator $authenticator)
...
+	}
+
 	// Usecase
 	public function interact()
 	{
 		// Fetch the entity, using provided identifiers...
 		$entity = $this->getEntity();
 
+		// Rate limit login attempts
+		$this->rateLimiter->pushThrottlingStrategy($this->throttlingStrategy);
```
rjmackay: Why not do this directly on in `setThrottlingStrategy`? or even better outside of this class entirely (in the DI layer or a wrapper)

src/Core/Usecase/User/LoginUser.php
```diff
@@ -33,12 +46,32 @@ public function setAuthenticator(PasswordAuthenticator $authenticator)
...
+		$this->rateLimiter->limit($entity->email);
```
rjmackay: What happens when this fails? is there an exception thrown?

rjmackay: Also why is the identifier email not ID?

src/Core/Usecase/User/LoginUser.php
```diff
@@ -16,6 +16,9 @@
 use Ushahidi\Core\Usecase\ReadRepository;
 use Ushahidi\Core\Usecase\ReadUsecase;
 
+use BehEh\Flaps\Flap;
```
rjmackay: Core Usecases shouldn't rely on anything outside Ushahidi\Core directly.
We can still use these, but they should be wrapped in a Core interface / object.

Ideally something super simple like an `Ushahidi\Core\Tools\RateLimiter` interface with just a `limit(Entity $entity)` method. Then extend Flap to implement that interface and forward the entity id to Flap.

application/classes/Ushahidi/Core.php
```diff
@@ -542,6 +542,34 @@ public static function init()
...
+		// Login rate limiter
+		$di->set('ratelimiter.login', $di->lazyNew('BehEh\Flaps\Flap'));
+
+		$di->params['BehEh\Flaps\Flap'] = [
+			'storage' => $di->lazyGet('ratelimiter.storage'),
+			'name' => 'login'
+		];
+
+		// Rate limiter storage
+		$di->set('ratelimiter.storage', $di->lazyNew('BehEh\Flaps\Storage\DoctrineCacheAdapter'));
```
rjmackay: For prod:
- Maybe use something like the FilesystemCache Adapter or something that'll work without extra tooling by default 
- We did have field for storing failed login counts etc in the DB.. but maybe that doesn't make sense anymore?
- Should ship an option for using memcache since we can deploy that easily to ushahidi.io

rjmackay:
@willdoran can you review too?

src/Core/Usecase/User/LoginUser.php
jasonmule: Got it.

src/Core/Usecase/User/LoginUser.php
jasonmule: We use an exception handler to handle violations and we can use IDs as the identifier.

application/classes/Ushahidi/Core.php
jasonmule: Not sure if we want to to track login counts in the DB. We could implement a DB adapter but I think something like memcache makes more sense for a production environment. I suppose we can make memcache and the FilesystemCache available as config options?

rjmackay: - Skip DB support.. sounds like extra effort.
- yup memcache and filesystem as config options would be good.

src/Core/Usecase/User/LoginUser.php
rjmackay: OK. Similar to the comment above we should be defining our own exception class and throwing that.. otherwise we end up with a dependency on Flaps exception

rjmackay:
Reviewed 9 of 9 files at r2.
Review status: 9 of 11 files reviewed at latest revision, 7 unresolved discussions.

---

<sup>**[application/classes/Ushahidi/Core.php, line 569 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/807#-K2j5IESRv9ZTFd4AJt9)** ([raw file](https://github.com/ushahidi/platform/blob/83cf20b1baaed3c62e659f0b79f85c7aeee1d914/application/classes/Ushahidi/Core.php#L569)):</sup>
should this be switching in the same way as ratelimiter.storage.cache?


---

<sup>**[application/classes/Ushahidi/ThrottlingViolationHandler.php, line 19 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/807#-K2j6UJoBinaQz1Fd0LP)** ([raw file](https://github.com/ushahidi/platform/blob/83cf20b1baaed3c62e659f0b79f85c7aeee1d914/application/classes/Ushahidi/ThrottlingViolationHandler.php#L19)):</sup>
where is this error handled? I think we need to catch it in the controller to transform it into an HTTP exception still


---

<sup>**[application/config/ratelimiter.php, line 14 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/807#-K2j6dIAaB3LMKFdiYdP)** ([raw file](https://github.com/ushahidi/platform/blob/83cf20b1baaed3c62e659f0b79f85c7aeee1d914/application/config/ratelimiter.php#L14)):</sup>
what happens when this is set to false?


---

<sup>**[src/Core/Usecase/User/LoginUser.php, line 60 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/807#-K2j5fqTwxJPj5AOSDxw)** ([raw file](https://github.com/ushahidi/platform/blob/83cf20b1baaed3c62e659f0b79f85c7aeee1d914/src/Core/Usecase/User/LoginUser.php#L60)):</sup>
Think we can drop this todo entirely now. rehashing isn't really needed right now.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/807)

rjmackay:
Review status: 9 of 11 files reviewed at latest revision, 8 unresolved discussions.

---

<sup>**[application/classes/Ushahidi/Core.php, line 577 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/807#-K2j7H85Eepz6yignxvj)** ([raw file](https://github.com/ushahidi/platform/blob/83cf20b1baaed3c62e659f0b79f85c7aeee1d914/application/classes/Ushahidi/Core.php#L577)):</sup>
ideally this should be defined elsewhere like
```
		$di->set('ratelimiter.config', function() use ($di) {
			return Kohana::$config->load('ratelimiter')->as_array();
		});
```
The loaded with
`$di->lazyGet('ratelimiter.config')` here.

See cdn.config for an example.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/807)

jasonmule:
<sup>**[application/classes/Ushahidi/Core.php, line 569 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/807#-K2j5IESRv9ZTFd4AJt9)** ([raw file](https://github.com/ushahidi/platform/blob/83cf20b1baaed3c62e659f0b79f85c7aeee1d914/application/classes/Ushahidi/Core.php#L569)):</sup>
Yeah. We should provide the Redis option as well.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/807)

jasonmule:
<sup>**[application/classes/Ushahidi/Core.php, line 569 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/807#-K2j5IESRv9ZTFd4AJt9)** ([raw file](https://github.com/ushahidi/platform/blob/83cf20b1baaed3c62e659f0b79f85c7aeee1d914/application/classes/Ushahidi/Core.php#L569)):</sup>
Actually , we discussed providing Memcached for production so I think we can stick with cache storage?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/807)

jasonmule:
Review status: 9 of 11 files reviewed at latest revision, 4 unresolved discussions.

---

<sup>**[application/classes/Ushahidi/Core.php, line 577 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/807#-K2j7H85Eepz6yignxvj)** ([raw file](https://github.com/ushahidi/platform/blob/83cf20b1baaed3c62e659f0b79f85c7aeee1d914/application/classes/Ushahidi/Core.php#L577)):</sup>
Done.


---

<sup>**[application/classes/Ushahidi/ThrottlingViolationHandler.php, line 19 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/807#-K2j6UJoBinaQz1Fd0LP)** ([raw file](https://github.com/ushahidi/platform/blob/83cf20b1baaed3c62e659f0b79f85c7aeee1d914/application/classes/Ushahidi/ThrottlingViolationHandler.php#L19)):</sup>
This will be caught in the oauth controller and treated as an invalid request error. Not sure if we want to send out a specific rate limiting error for password requests...


---

<sup>**[application/config/ratelimiter.php, line 14 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/807#-K2j6dIAaB3LMKFdiYdP)** ([raw file](https://github.com/ushahidi/platform/blob/83cf20b1baaed3c62e659f0b79f85c7aeee1d914/application/config/ratelimiter.php#L14)):</sup>
Added a comment to make this clear.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/807)

jasonmule:
Review status: 9 of 11 files reviewed at latest revision, 5 unresolved discussions.

---

<sup>**[application/classes/Ushahidi/Core.php, line 554 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/807#-K28hrTIW0LR7coshk5H-r1-554)** ([raw file](https://github.com/ushahidi/platform/blob/88626636ddf39265e58eaf0f0e5c4564b1cf009f/application/classes/Ushahidi/Core.php#L554)):</sup>
Done.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/807)

rjmackay:
Review status: 6 of 11 files reviewed at latest revision, 3 unresolved discussions.

---

<sup>**[application/classes/Ushahidi/Core.php, line 569 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/807#-K2j5IESRv9ZTFd4AJt9)** ([raw file](https://github.com/ushahidi/platform/blob/83cf20b1baaed3c62e659f0b79f85c7aeee1d914/application/classes/Ushahidi/Core.php#L569)):</sup>
I guess I'm not clear what the difference is between ratelimiter.storage and ratelimiter.storage.cache ? what are they each used for? why is one using fs or memcached storage while they other is always in memory?


---

<sup>**[application/config/ratelimiter.php, line 19 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform/807#-K2nqXV0BDG1b1plojZc)** ([raw file](https://github.com/ushahidi/platform/blob/ed5d870b3076576de223013168000b7163c33ee7/application/config/ratelimiter.php#L19)):</sup>
in that case, can we default to filesystem based cache? 


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/807)

jasonmule:
Review status: 6 of 11 files reviewed at latest revision, 3 unresolved discussions.

---

<sup>**[application/classes/Ushahidi/Core.php, line 569 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/807#-K2j5IESRv9ZTFd4AJt9)** ([raw file](https://github.com/ushahidi/platform/blob/83cf20b1baaed3c62e659f0b79f85c7aeee1d914/application/classes/Ushahidi/Core.php#L569)):</sup>
Removing the ratelimiter.storage prefix and using ratelimiter.cache


---

<sup>**[application/config/ratelimiter.php, line 19 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform/807#-K2nqXV0BDG1b1plojZc)** ([raw file](https://github.com/ushahidi/platform/blob/ed5d870b3076576de223013168000b7163c33ee7/application/config/ratelimiter.php#L19)):</sup>
Done.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/807)

rjmackay:
Good to land! :+1: 


---

Review status: 6 of 11 files reviewed at latest revision, 3 unresolved discussions.

---

<sup>**[application/classes/Ushahidi/Core.php, line 569 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/807#-K2j5IESRv9ZTFd4AJt9)** ([raw file](https://github.com/ushahidi/platform/blob/83cf20b1baaed3c62e659f0b79f85c7aeee1d914/application/classes/Ushahidi/Core.php#L569)):</sup>
... finally figure out how this works. I'm super slow sorry!


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/807)

PR #849, opened 2015-12-08
Add user id to messages to record who sent a reply

rjmackay:
Prerequisite for building the post messages list view.

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform/849)

### DIFF
--- a/application/tests/features/api.messages.feature
+++ b/application/tests/features/api.messages.feature
@@ -18,6 +18,7 @@ Feature: Testing the Messages API
         And the type of the "id" property is "numeric"
         And the "message" property equals "Test creating outgoing"
         And the "status" property equals "pending"
+        And the "user.id" property equals "2"
         Then the guzzle status code should be 200
 
     Scenario: Creating an incoming message should fail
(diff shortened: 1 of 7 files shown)


rjmackay:
cc @hollycorbett try this :)

application/tests/features/api.messages.feature
```diff
@@ -224,3 +225,20 @@ Feature: Testing the Messages API
         Then the response is JSON
         And the "id" property equals "110"
         Then the guzzle status code should be 200
+
+    Scenario: Admin can set user id when creating a message
```
rjmackay: TBH I'm not 100% this should be the case. Its what we did for posts. But it might actually be simpler if CreateMessage just **always** set user id to the current user..

src/Core/Usecase/Message/CreateMessage.php
```diff
@@ -0,0 +1,34 @@
...
+class CreateMessage extends CreateUsecase
+{
+
+	protected function getEntity()
+	{
+		$entity = parent::getEntity();
+
+		// If no user information is provided, default to the current session user.
+		if (
+			empty($entity->user_id) &&
```
rjmackay: could skip this condition and ALWAYS set user id.

rjmackay:
I haven't figured out how to block changing the user id within our current validators :/

Issue #844, opened 2015-12-02
Upload posts from a CSV file


jasonmule:
Feature Epic: CSV import
User Need: Bulk data import
Primary users of this feature: Users with many various data sources other than those available within Ushahidi 

## Introduction: 
Users that collect data outside of their Ushahidi deployment should be able to upload that data into the deployment.

## User Stories:
- Logged in users with CSV import permission can import a CSV file and then structure (clean up) and publish any data necessary
- User is notified if errors in either import occur

PR #850, opened 2015-12-08
844 csv api

jasonmule:
#844

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform/850)


rjmackay:
Sorry for the insanely late review on this. I'm concerned that this seems to repeat things that were already handled by the Import usecase. And seems to be very CSV specific. I've added comments inline.. but let me know if you'd rather talk through it.
I'd be very happy to be convinced I'm wrong here too.. since I don't want to be adding more work right now.

I'd also expected to see the CSV upload and the import as different API endpoints. Since the upload could just use the media API, and the Import just need to reference a media ID after that. However I might be missing something.. @willdoran which is easier for the client? Do we do an upload as a first step before import anyway?

---

Reviewed 21 of 26 files at r1, 1 of 1 files at r2, 5 of 6 files at r3.
Review status: 26 of 27 files reviewed at latest revision, 8 unresolved discussions.

---

*[application/classes/Controller/Api/CSV.php, line 11 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform/850#-K7T1EeuK0WU29FmH-2E:-K7T1EeuK0WU29FmH-2F:630948088) ([raw file](https://github.com/ushahidi/platform/blob/26b42a7261f50816106e80e07de03359514b8f9f/application/classes/Controller/Api/CSV.php#L11)):*
this should probably be call import or import/csv since we'll have an export tool later

---

*[application/classes/Ushahidi/Listener/CSVPostListener.php, line 49 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform/850#-K7T3XnxWEBqtnmKTkSP:-K7T3Xnyq0nG2Qfm3qKy:339535824) ([raw file](https://github.com/ushahidi/platform/blob/26b42a7261f50816106e80e07de03359514b8f9f/application/classes/Ushahidi/Listener/CSVPostListener.php#L49)):*
Isn't this almost exactly what the ImportUsecase does? ie. couldn't we just be calling that here?

---

*[application/routes/default.php, line 31 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform/850#-K7T5aKIV4K05iAcX59P:-K7T5aKIV4K05iAcX59Q:-375443819) ([raw file](https://github.com/ushahidi/platform/blob/26b42a7261f50816106e80e07de03359514b8f9f/application/routes/default.php#L31)):*
lets do a generic import/ api please. Or at least import/csv

---

*[application/tests/datasets/ushahidi/sample.csv, line 1 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform/850#-K7T5jGqJiKGXMmBt4XH:-K7T5jGqJiKGXMmBt4XI:462522102) ([raw file](https://github.com/ushahidi/platform/blob/26b42a7261f50816106e80e07de03359514b8f9f/application/tests/datasets/ushahidi/sample.csv#L1)):*
Can we make sure we test for handling quotes fields? and various escaping combos.

---

*[migrations/20151127093600_create_csv.php, line 18 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform/850#-K7T3z3X24XAG0cqsHsr:-K7T3z3X24XAG0cqsHss:-1095296557) ([raw file](https://github.com/ushahidi/platform/blob/26b42a7261f50816106e80e07de03359514b8f9f/migrations/20151127093600_create_csv.php#L18)):*
So this table represents an import? again I really think this should be more generic, ie. called 'import' not csv. 

Why are the post fields (tags, status, published_to, form_id etc) not just part of the mapping? Currently we'll have to update this table any time the post table changes.

---

*[src/Core/Entity/CSV.php, line 16 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform/850#-K7T4NMo1HsrHE9s95w-:-K7T4NMo1HsrHE9s95w0:-672265257) ([raw file](https://github.com/ushahidi/platform/blob/26b42a7261f50816106e80e07de03359514b8f9f/src/Core/Entity/CSV.php#L16)):*
I think this should be 'Import' with type: csv 

---

*[src/Core/Usecase/CSV/CreateCSVUsecase.php, line 55 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform/850#-K7T4jWzjtkpDER5v6TF:-K7T4jX-eI3008BX53vz:668176549) ([raw file](https://github.com/ushahidi/platform/blob/26b42a7261f50816106e80e07de03359514b8f9f/src/Core/Usecase/CSV/CreateCSVUsecase.php#L55)):*
Is this the only real difference here? any way we could just split it into a trait? .. even better would be building some kind of metadata hook into the Media api so we can add info about CSVs, images etc without adding extra APIs. But that probably an exercise for the future.

---

*[src/Core/Usecase/CSV/DeleteCSVUsecase.php, line 29 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform/850#-K7T5CIEV0sRhg7m-n31:-K7T5CIEV0sRhg7m-n32:-2100250104) ([raw file](https://github.com/ushahidi/platform/blob/26b42a7261f50816106e80e07de03359514b8f9f/src/Core/Usecase/CSV/DeleteCSVUsecase.php#L29)):*
what differs here vs standard media upload?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/850#-:-K7T26uvIpDqtReWZNtN:1565576849)*

willdoran:
From the client's side, we pass up the file, and get back the entity id. We pass that onto the import mapper. We pull down the import entity with the columns parsed from the file and then allow the users to map fields to appropriate columns.

Based on Jason's suggestion we also have an automapper, if the column name matches an existing field we auto set the mapping - so title to title. This treats the names as case insensitive. If the name doesn't appear its default is set to "Don't Import" and the user must explicitly set it.

From the clients view point, it treats data file type as generic - we can easily add a field sent to the platform defining the file type. At the moment, we have a single endpoint on the client - data-importer. This connects to /csv but can become /import as required.

Using the media api is possible but I'd like to logically separate them as the media should store the data long term where as the uploaded data file is just a temp file. But that's just a preference. Having only the media upload would be slightly less code to maintain if the upload system changes for some reason.

---

Review status: 26 of 27 files reviewed at latest revision, 8 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/850#-:-K7TFOldJXksJKbsHJHo:-1545570667)*

willdoran:
Review status: 26 of 27 files reviewed at latest revision, 8 unresolved discussions.

---

*[application/tests/datasets/ushahidi/sample.csv, line 1 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform/850#-K7T5jGqJiKGXMmBt4XH:-K7TK5k3VZU3BacTpqkm:794913930) ([raw file](https://github.com/ushahidi/platform/blob/26b42a7261f50816106e80e07de03359514b8f9f/application/tests/datasets/ushahidi/sample.csv#L1)):*
We could use this
https://www.mockaroo.com/
it even generates broken data using NaughtyStrings

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/850)*

rjmackay:
:100: on the automapper :)

> Using the media api is possible but I'd like to logically separate them as the media should store the data long term where as the uploaded data file is just a temp file

Good point. Even if we do batched imports we don't want to leave the source CSV around forever or publicly available.
That's a good case having CSV upload separate. Similarly we need the extra API to read columns from a CSV.

What I'm still figuring out is: should the actual import be a different API call though?

Right now you're doing something like this right?
POST /csv with file
GET /csv/1 to get columns
PUT /csv/1 with mappings

It feels like setting the mappings and triggering an import should be a seperate API call. Not sure about the name yet but something like ie.
POST /csv with file
GET /csv/1 to get columns
POST /csv/1/import with mappings + CSV id.

Semantically thats a bit better using POST vs PUT. It'd also allow triggering an import multiple times.. and it seems more extensible too, in future we could point to a URL instead of a CSV id, or delay the actual import to later or set a schedule for importing too. Does that make sense? or am I just making it too complicated?

---

Review status: 26 of 27 files reviewed at latest revision, 10 unresolved discussions.

---

*[application/classes/Ushahidi/Repository/CSV.php, line 58 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform/850#-K7TTglhveZS2KXcbUaa:-K7TTgliIzG-VqTraoQ0:1518067560) ([raw file](https://github.com/ushahidi/platform/blob/26b42a7261f50816106e80e07de03359514b8f9f/application/classes/Ushahidi/Repository/CSV.php#L58)):*
I don't think we should be emitting events in a repo.
Repo's are for storage/persistence. This seems like something that should happen in the usecase itself. 

---

*[src/Core/Entity/CSV.php, line 31 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform/850#-K7TTvDtEGntltEMntHg:-K7TTvDuEFmwZkFNW_nR:444139586) ([raw file](https://github.com/ushahidi/platform/blob/26b42a7261f50816106e80e07de03359514b8f9f/src/Core/Entity/CSV.php#L31)):*
So this is the flag to trigger an import?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/850#-:-K7TNjuLsHIXkPXLa4-x:888708169)*

rjmackay:
Review status: 26 of 27 files reviewed at latest revision, 11 unresolved discussions.

---

*[application/classes/Ushahidi/Transformer/CSVPostTransformer.php, line 15 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform/850#-K7TVoEh3tR-RXObsB22:-K7TVoEigxk_0pYs7hMw:-883277545) ([raw file](https://github.com/ushahidi/platform/blob/26b42a7261f50816106e80e07de03359514b8f9f/application/classes/Ushahidi/Transformer/CSVPostTransformer.php#L15)):*
Whats the key difference here that couldn't be done by the existing `Ushahidi_Transformer_MappingTransformer` ? I don't mind changing the implementation but we really shouldn't need to implementations of the same thing.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/850)*

rjmackay:
Reviewed 1 of 6 files at r3.
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/850)*

rjmackay:
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[application/tests/datasets/ushahidi/sample.csv, line 1 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform/850#-K7T5jGqJiKGXMmBt4XH:-K7TwVknZs7OmLN1cIMB:643441085) ([raw file](https://github.com/ushahidi/platform/blob/26b42a7261f50816106e80e07de03359514b8f9f/application/tests/datasets/ushahidi/sample.csv#L1)):*
:+1: I would love to see more tests which intentionally try to break everything :)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/850)*

jasonmule:
It actually makes more sense to trigger an import through a separate API call.

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/850#-:-K7fsF-eRQsQzcUUQbwP:1440727407)*

rjmackay:
> It actually makes more sense to trigger an import through a separate API call.

Cool. Any idea how much work is left to address the review feedback and split the API calls? Do you want to do all the before it lands or are the some bits you think could just happen in a follow up task?

---

Review status: all files reviewed at latest revision, 10 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/850#-:-K7mzjAz7aeG1NCMLYGj:-1171667539)*

jasonmule:
*[application/classes/Ushahidi/Transformer/CSVPostTransformer.php, line 15 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform/850#-K7TVoEh3tR-RXObsB22:-K7lVsalq6ZNZ27b4SDr:-458013056) ([raw file](https://github.com/ushahidi/platform/blob/26b42a7261f50816106e80e07de03359514b8f9f/application/classes/Ushahidi/Transformer/CSVPostTransformer.php#L15)):*
- the ability to filter post fields from values e.g. title
- mapping the columns and taking into account that not all columns will be mapped
- Finally creating form values from the record

It's possible that I missed something as well about how `Ushahidi_Transformer_MappingTransformer` works

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/850)*

rjmackay:
Can we close this PR now its landed to develop??

PR #861, opened 2016-01-04
Make sure dependency installation completes before checking for phinx

tobiasziegler:
Hi there,

My attempts at installing a brand new development server [following the Vagrant setup instructions](https://www.ushahidi.com/support/install-ushahidi#installing-with-vagrant-a-nodejs-dev-server) failed with a "!!! Migration tool phinx not installed, unable to run migrations" message. Looking at `/bin/update` I saw that the check for whether phinx exists comes before Composer has installed the dependencies (which includes phinx). As a workaround I used `vagrant ssh` and then ran the bin-update script once with `--no-migrate` (letting Composer install the dependencies) and then again without the `--no-migrate` flag to get the database and platform fully working.

As a proposed fix, I've moved all of the phinx checks to occur after the Composer install process has been finished. I can't see any adverse consequences from making the changes but if there's a better way to address the issue please feel free to let me know.

When testing the `/bin/update` changes I found that Puppet's default timeout of 300 seconds kicked in while it was downloading the dependencies. At this point I disabled the timeout, but the alternative would be to set a longer timeout that's appropriate (on my current dodgy connection it took 1400-1500 seconds - maybe 30 minutes/1800 seconds could work?).

Thanks for all the work that has gone into this project - after getting around this issue I now have a local setup working for development and testing.

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform/861)

### DIFF
--- a/bin/update
+++ b/bin/update
@@ -84,17 +84,11 @@ if [ $? -gt 0 ]; then
 		exit 1
 	fi
 fi
-# Check phinx
-if [[ ! -f "$phinx" && "$migrate" = true ]]; then
-	echo "!!! Migration tool phinx not installed, unable to run migrations"
-	exit 1
-fi
 
 # Disable interactive questions
 if [ "$nointeraction" = true ]; then
 	echo ">>> Disable interactive questions"
 	composer="$composer --no-interaction"
-	phinx="$phinx --no-interaction"
 fi
 
 # Deployment mode, disable all developer dependencies :)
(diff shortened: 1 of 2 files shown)


tobiasziegler:
[clabot:check]

aoduor:
Hey @tobiasziegler thanks for this. Someone on our team will review your pull request and be in touch in the course of this and next week!

willdoran:
Hey thanks so much for your time on this! It looks good and passes tests for me. I'll merge now :)

Issue #859, opened 2015-12-15
Private Deployments


rjmackay:
Feature Epic: Private instance
User Need: Privacy
Triage: Pending
Primary users of this feature: Responders who need to triage and respond to sensitive and private data.
User Testing: Pending
Point estimate: 21

Introduction:
Private instance will require login before a user can see any data. Once logged in a private instance has identical functionality and design to any other instance

User Stories:
- Admin user should be able to login and view / manage deployment
- Member users should be able to login and view data in the deployment, as usual
- Custom roles should be able to login and view data in the deployment plus access any permissions they have
- Non logged in users arrive on homepage and directed to login
- Non members can request access to a private deployment (*nice to have)
- Non members cannot self-register for private deployment
- When publishing posts on a private deployment (even when ‘published to everyone’) the posts should only be visible to logged in users.

PR #866, opened 2016-01-11
Create private deployment

jasonmule:
#859

- [ ] Add API tests

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform/866)

### DIFF
--- a/application/classes/Ushahidi/Core.php
+++ b/application/classes/Ushahidi/Core.php
@@ -57,7 +57,8 @@ abstract class Ushahidi_Core {
 
 		// Private deployment config settings
 		$di->set('site.private', function() use ($di) {
-			return Kohana::$config->load('site.private');
+			return Kohana::$config->load('site.private')
+				and Kohana::$config->load('features.private.enabled');
 		});
 
 		$di->set('tool.uploader.prefix', function() use ($di) {
--- a/application/classes/Ushahidi/Core.php
+++ b/application/classes/Ushahidi/Core.php
@@ -51,11 +51,14 @@ abstract class Ushahidi_Core {
 		});
 
 		// Ratelimiter config settings
-
 		$di->set('ratelimiter.config', function() use ($di) {
 			return Kohana::$config->load('ratelimiter')->as_array();
 		});
 
+		// Private deployment config settings
+		$di->set('site.private', function() use ($di) {
+			return Kohana::$config->load('site.private');
+		});
 
 		$di->set('tool.uploader.prefix', function() use ($di) {
 			// Is this a multisite install?
--- a/src/Core/Tool/Authorizer/ContactAuthorizer.php
+++ b/src/Core/Tool/Authorizer/ContactAuthorizer.php
@@ -42,7 +42,7 @@ class ContactAuthorizer implements Authorizer
 		$user = $this->getUser();
 
 		// Only logged in users have access if the deployment is private
-		if ($this->isPrivate() and !$user->getId()) {
+		if (!$this->hasAccess()) {
 			return false;
 		}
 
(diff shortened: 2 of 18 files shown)


application/classes/Ushahidi/Core.php
```diff
@@ -51,11 +51,14 @@ public static function init()
 		});
 
 		// Ratelimiter config settings
-
 		$di->set('ratelimiter.config', function() use ($di) {
 			return Kohana::$config->load('ratelimiter')->as_array();
 		});
 
+		// Private deployment config settings
+		$di->set('site.private', function() use ($di) {
```
rjmackay: :+1:

src/Core/Tool/Authorizer/ContactAuthorizer.php
```diff
@@ -31,12 +32,20 @@ class ContactAuthorizer implements Authorizer
...
+	use PrivateDeployment;
+
 	/* Authorizer */
 	public function isAllowed(Entity $entity, $privilege)
 	{
 		// These checks are run within the user context.
 		$user = $this->getUser();
 
+		// Only logged in users have access if the deployment is private
+		if ($this->isPrivate() and !$user->getId()) {
```
rjmackay: I think you can do `this->getUserId` but I think we should probably add something like `->hasUser` to UserContext so we're not manually checking the id (though we already do that elsewhere)

rjmackay: In fact, given you're doing this check everywhere, maybe alias it into the PrivateDeployment trait too. ie.
Add `canAccessDeployment` (or a better name) and do
```
if (! $this->canAccessDeployment()) {
  return false;
}
```
or maybe thats overkill?

src/Init.php
```diff
@@ -301,6 +301,7 @@ function feature($name)
 $di->setter['Ushahidi\Core\Usecase\Form\VerifyStageLoaded']['setStageRepository']
 	= $di->lazyGet('repository.form_stage');
 $di->setter['Ushahidi\Core\Traits\Event']['setEmitter'] = $di->lazyNew('League\Event\Emitter');
+$di->setter['Ushahidi\Core\Traits\PrivateDeployment']['setPrivate'] = $di->lazyGet('site.private');
```
rjmackay: Nice and clean :+1:

rjmackay:
Authorizer could be a little clean but this looks good.
However we really need some tests for this.

src/Core/Tool/Authorizer/ContactAuthorizer.php
jasonmule: Much nicer.

rjmackay:
Changes look good. Need some test then good to land.

rjmackay:
Reviewed 4 of 15 files at r1, 11 of 11 files at r2.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/Core/Tool/Authorizer/UserAuthorizer.php, line 85 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/866#-K7xAr6aanRIbpyZ1m0O:-K7xAr6aanRIbpyZ1m0P:1575756833) ([raw file](https://github.com/ushahidi/platform/blob/ea7a2b4c019e3aadcc5f414de605835566f9f155/src/Core/Tool/Authorizer/UserAuthorizer.php#L85)):*
We should add tests to make sure users can't register on private deployments.
Otherwise it'd be too easy to break this in future updates to this Authorizer.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/866)*

rjmackay:
Reviewed 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/866)*

rjmackay:
Looks good. But make sure to merge to develop, not master

---

Reviewed 3 of 3 files at r4.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform/866#-:-KAahrYhMea_s4OFtIAP:755787280)*

PR #911, opened 2016-02-17
Support for location points

jasonmule:
[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform/911)

### DIFF
--- a/application/classes/Ushahidi/FileReader/CSV.php
+++ b/application/classes/Ushahidi/FileReader/CSV.php
@@ -38,6 +38,7 @@ class Ushahidi_FileReader_CSV implements FileReader
 	public function process($file)
 	{
 		$reader = $this->reader_factory->createReader($file);
+		$reader->setEnclosure('\'');
 
 		// Filter out empty rows
 		$nbColumns = count($reader->fetchOne());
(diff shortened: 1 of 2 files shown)


application/classes/Ushahidi/FileReader/CSV.php
```diff
@@ -38,6 +38,7 @@ public function setReaderFactory(ReaderFactory $reader_factory)
 	public function process($file)
 	{
 		$reader = $this->reader_factory->createReader($file);
+		$reader->setEnclosure('\'');
```
rjmackay: Does this mean we ONLY support single quoted CSVs??

rjmackay:
@willdoran @jasonmule is this documented somewhere in the UI ? ie. locations can be in JSON format in the CSV...

I'm not 100% happy with this since its a weird way to have locations in a CSV.. we should look at handling simple lat,lon and lon,lat formats later.

willdoran:
On the data-import page, there is a description of the format with an example but it might be off-putting/complex. I'm not sure if I described it clearly enough.

rjmackay:
Alright. That'll do for release at least.. is it possible to import lat and lon from separate columns at the moment?

aoduor:
Perhaps this is the source of my failed data uploads. Are the lat/lon values supposed to be in separate columns? or the same column? And what should the column header be?

PR #922, opened 2016-02-17
Provide error messages for unhandled exceptions

### LINKED ISSUE #912: CSV Upload Error Message is blank
image

Happened when I uploaded a CSV file that doesn't meet the guidelines. It would be helpful if the alert box explained why the error occurred.

jasonmule:
Fixes #912

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform/922)

### DIFF
--- a/application/classes/Ushahidi/Rest.php
+++ b/application/classes/Ushahidi/Rest.php
@@ -462,6 +462,14 @@ abstract class Ushahidi_Rest extends Controller {
 				[':error' => $e->getMessage()]
 			);
 		}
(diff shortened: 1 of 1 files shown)


rjmackay:
What was happening prior to this? We should have already been throwing a 500 with a stacktrace (in dev mode anyway).

rjmackay:
This was swallowing other exceptions ie. validation errors. I'm just reverting it since we already format 500 errors as JSON. If anything we should just fix that formatting.

Issue #961, opened 2016-03-01
Admin can view all notifications and contacts


jasonmule:
Fix this so that users can view only the notifications they are subscribed to on the current UI.

rjmackay:
I think we're double point scoring this, by scoring this issue AND #954

jasonmule:
Closing this as it's most likely a duplicate of #954

Issue #954, opened 2016-02-26
Duplicate notification options in Notifications view


sebm90:
- On krcs.ushahidi.com, I login and go to "Notificatons"

- Expected behaviour: I see one notificaiton option for my saved search "Fire Reports"

- Actual behaviour: I see 18 "Fire Report" notification options:
![image](https://cloud.githubusercontent.com/assets/5349387/13349474/59d7edf2-dc7a-11e5-9bda-9c152777f653.png)

PR #962, opened 2016-03-02
961 notifications view

jasonmule:
Closes ushahidi/platform#961. Should fix ushahidi/platform#954 if duplicates are as a result of an admin seeing multiple subscriptions from different users.

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform/962)


rjmackay:
1. Looks like you've pulled a bunch of other changes with your PR
2. I don't think this should get fixed in the API. Its correct that the API should let an admin manage others notifications.. but the "My Notifications" UI should be limited to just their notifications.
We should pass a param like `?user=me` to the API and grab jusst their notifications.

Make sense??

jasonmule:
Noted. Going ahead and closing this.

Issue #953, opened 2016-02-24
Notifications are never queued


rjmackay:
1. Create a saved search for post within 100km of Auckland
2. Set up notifications: email + SMS
3. Add a new post with location Auckland.
- Confirm post shows up in saved search.
Expected: messages queued to send
Actual: No messages are ever queued

rjmackay:
@jasonmule can you look into this? I haven't tried to reproduce locally yet.
This is a problem for the KRC deployment so needs to get fixed asap

jasonmule:
I tried to reproduced this locally and it's possible that this fails because of the following reasons:

1. The user has not subscribed to notifications. When the user adds a notification for the first time, it's probably a good idea to direct them to manage their notifications and add at least one contact.
2. The post has not been published. At the moment we only send out notifications for published posts.

The following commands queue notifications for saved searches:

1 . Find posts from saved searches
```
bin/ushahidi savedsearch
```

2 . Queue notifications
```
bin/ushahidi notification queue
```

sebm90:
When I tried at krcs.ushahidi.io, I checked that I was subscribed to notifications before publishing a post, and made sure that the post had been published (and all post steps completed). The notification was still not sent.

jasonmule:
Thanks @sebm90. We should probably still remind you to add a contact if you don't have one when you subscribe.

sebm90:
@jasonmule Could you explain what adding a contact involves? Aren't my own contact details (my email, which is already in the system) enough? Does this need to be a different person?

jasonmule:
@rjmackay Did you ran `bin/ushahidi savedsearch` ?

jasonmule:
@sebm90 Could you go to Manage notifications and check that your email contact is turned on to receive notifications?

ushbot:
@jasonmule I was testing on staging. I haven't tried to reproduce locally yet. I'll do that today.

> The post has not been published. At the moment we only send out notifications for published posts.

That *might* have been my issue.. though I thought I tested both. I definitely enabled notifications in the UI.

rjmackay:
oops that was me not a bot.

sebm90:
@jasonmule My email contact is turned on and was turned on when I tested notifications:
![image](https://cloud.githubusercontent.com/assets/5349387/13348856/3d597640-dc76-11e5-9185-b95a90b30d00.png)

rjmackay:
Just reproduced the issue locally for myself. 
[...]

PR #965, opened 2016-03-04
Fix issue updating posts_sets with new posts on saved searches

rjmackay:
SavedSearch console command was using a single SearchData - this meant values from the previous search were left in the state.

Updated to use the DataFactory and get a fresh data object each time.

Refs ushahidi/platform#953

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform/965)


willdoran:
This looks good, it appears to be working in my local.

PR #968, opened 2016-03-04
Use user param to filter notifications and contacts

jasonmule:
- Filter contacts and notifications using currently logged in user

Test these changes by:
- See api.notifications.feature and api.contacts.feature

Fixes ushahidi/platform#954 .

Ping @ushahidi/platform

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform/968)

### DIFF
--- a/application/classes/Ushahidi/Repository/Contact.php
+++ b/application/classes/Ushahidi/Repository/Contact.php
@@ -33,7 +33,7 @@ class Ushahidi_Repository_Contact extends Ushahidi_Repository implements
 			->execute($this->db);
 		return $result->get('id', 0);
 	}
-	
+
 	// Ushahidi_Repository
 	protected function getTable()
 	{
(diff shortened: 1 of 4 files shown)


application/classes/Ushahidi/Repository/Contact.php
```diff
@@ -60,11 +60,10 @@ protected function setSearchConditions(SearchData $search)
 	{
 		$query = $this->search_query;
 
-		$user = $this->getUser();
-
-		// Limit search to user's records unless they are admin
-		if (! $this->isUserAdmin($user)) {
```
rjmackay: I don't think we should be removing this. Though maybe we should update it to check a permission as well as isUserAdmin()

application/classes/Ushahidi/Repository/Notification.php
```diff
@@ -39,12 +39,11 @@ protected function getTable()
 	public function setSearchConditions(SearchData $search)
 	{
 		$query = $this->search_query;
-				
-		$user = $this->getUser();
 
-		// Limit search to user's records unless they are admin
-		if (! $this->isUserAdmin($user)) {
```
rjmackay: I don't think we should be removing this. Though maybe we should update it to check a permission as well as isUserAdmin()

PR #971, opened 2016-03-09
adding view enable for plans

willdoran:
- Minor change adding plans view as configurable

Test these changes by:
- Using the client branch feature/cloud-reintegration, test to check that the plan setting item is available or not as dependent on the setting in application/configs/features

Fixes ushahidi/platform# .

Ping @ushahidi/platform

---
This change is [Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform/971)

### DIFF
--- a/application/config/features.php
+++ b/application/config/features.php
@@ -14,12 +14,12 @@ return [
 
 	// Post views
 	'views' => [
-		'map' => TRUE,
+		'map' => FALSE,
 		'list' => TRUE,
 		'chart' => TRUE,
 		'timeline' => TRUE,
 		'activity' => TRUE,
-    'plans' => FALSE,
+    'plan' => TRUE,
 	],
 
 	// Data sources
--- a/application/config/features.php
+++ b/application/config/features.php
@@ -19,7 +19,7 @@ return [
 		'chart' => TRUE,
 		'timeline' => TRUE,
 		'activity' => TRUE,
-    'plan' => TRUE,
+        'plan' => TRUE,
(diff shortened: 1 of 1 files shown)


application/config/features.php
```diff
@@ -14,11 +14,12 @@
 
 	// Post views
 	'views' => [
-		'map' => TRUE,
+		'map' => FALSE,
 		'list' => TRUE,
 		'chart' => TRUE,
 		'timeline' => TRUE,
 		'activity' => TRUE,
+    'plan' => TRUE,
```
jasonmule: Might need to check indentation.

application/config/features.php
```diff
@@ -14,11 +14,12 @@
 
 	// Post views
 	'views' => [
-		'map' => TRUE,
+		'map' => FALSE,
```
rjmackay: Whats up with this? map views should be enabled by default - otherwise OSS users are going to find they have no map view.

application/config/features.php
```diff
@@ -14,11 +14,12 @@
...
 		'list' => TRUE,
 		'chart' => TRUE,
 		'timeline' => TRUE,
 		'activity' => TRUE,
+        'plan' => TRUE,
```
rjmackay: If this is cloud only - it should be disabled by default. The default state here should represent the OSS settings.

willdoran:
For both of those, I had inverted them, I reverted it for develop so it's correct in the develop branch.

Issue #443, opened 2015-08-22
View thread of messages and replies associated with a post


ushbot:
This was discussed but never had any designs or patterns produced.

-   Any post that was created from an SMS, email or twitter message should have a way for an admin to see the message it was created from, and any replies etc.
-   My understanding was this would live under the 'View as...' menu on a post detail page. Along with revision history etc.
    -   However we have no pattern for what that view/list would look like.

*Created by @rjmackay on 2015-07-31 13:49:41.*

*Imported from https://phabricator.ushahidi.com/T1754*

sophshep:
• Add "View messages associated with the post" functionality to post detail page.

rjmackay:
+ ability to reply to message from that page.

sophshep:
@rjmackay - Two new pages to show this:

1. [Post detail with messages associated with it](http://github.ushahidi.org/platform-pattern-library/assets/html/5_layouts/post-detail-with-messages.html) - The link to view messages is tied to the "posted by" credit.

2. [Post messages thread](http://github.ushahidi.org/platform-pattern-library/assets/html/5_layouts/post-messages.html) - Reusing the timeline style, but added an "alt" color to show the difference between who is speaking, and added a form for a reply at the bottom.

rjmackay:
Awesome!
- What would 'Verify this number' do?

rjmackay:
This is already built for msheria.. we just need to clean up and merge to core.

PR #999, opened 2016-03-29
Show and reply to messages associated with a post

jasonmule:
- Shows email/ SMS messages related to a post.
- Allow a user to reply to an incoming email/ SMS.

Test these changes by:
- Use with https://github.com/ushahidi/platform-client/pull/146

Fixes ushahidi/platform#443 .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/999)


application/classes/Ushahidi/DataProvider.php
```diff
@@ -43,7 +43,7 @@ public function receive($type, $from, $message, $to = NULL, $title = NULL, $data
 		$usecase = service('factory.usecase')->get('messages', 'receive');
 		try
 		{
-			$usecase->setPayload(compact(['type', 'from', 'message', 'to', 'title', 'data_provider_message_id', 'data_provider', 'contact_type']))
+			$message_id = $usecase->setPayload(compact(['type', 'from', 'message', 'to', 'title', 'data_provider_message_id', 'data_provider', 'contact_type']))
```
rjmackay: Do we need this change? if so can we update the docblock to with a correct @return

plugins/smssync/classes/Controller/Sms/Smssync.php
```diff
@@ -91,7 +98,7 @@ private function _incoming()
 		// Remove Non-Numeric characters because that's what the DB has
 		$to = preg_replace("/[^0-9,.]/", "", $this->request->post('sent_to'));
 
-		$from = preg_replace("/[^0-9,.]/", "", $from);
+		$from = preg_replace("/[^0-9A-Za-z,.]/", "", $from);
```
rjmackay: Might be best to make this even more forgiving.. If we can get alpha chars in phone numbers, we can probably get all of unicode letters.

src/Core/Usecase/Message/CreateMessage.php
```diff
@@ -21,6 +21,12 @@ protected function getEntity()
 	{
 		$entity = parent::getEntity();
 
+		if (! empty($this->payload['parent_id'])) {
```
rjmackay: Needs a comment.

rjmackay:
Looks good. Nice and simple mostly :)

application/classes/Ushahidi/Validator/Contact/Update.php
```diff
@@ -69,18 +69,8 @@ public function valid_contact($contact, $data, $validation)
...
-		{
-			// Remove all non-digit characters from the number
-			$number = preg_replace('/\D+/', '', $contact);
-
-			if (strlen($number) == 0)
-			{
-				$validation->error('contact', 'invalid_phone', [$contact]);
-			}
-		}
+		// There's no easy to validate phone numbers. Even just accepting numbers
```
rjmackay: @jasonmule moved this over from msheria too. We hit an issue where SMSSync was forwarding messages with a phone number like "Safaricom" .. however this probably removes the validation we wanted creating contacts from the UI.. Any ideas for a solution to both?

jasonmule: We probably shouldn't be using a regex to validate phone numbers. Maybe use https://github.com/googlei18n/libphonenumber but that still won't validate "Safaricom". For now, we could use a separate validator for the receive use case?

plugins/smssync/classes/Controller/Sms/Smssync.php
jasonmule: Maybe we should limit this to alphanumeric senders? i.e 1 - 11 alphanumeric characters including a space.

application/classes/Ushahidi/DataProvider.php
jasonmule: Proabably not.

rjmackay:
Reviewed 7 of 11 files at r1, 1 of 1 files at r2, 6 of 6 files at r4, 1 of 1 files at r5.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/ushahidi/platform/999)*

rjmackay:
Good to go :shipit: 

---

Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/ushahidi/platform/999#-:-KFuZqmJ8y4k0NYMN8Te:1178353558)*

Issue #1003, opened 2016-04-05
Create migration runner for shared servers


rjmackay:
**Introduction:**
Running migrations currently requires commandline access but shared servers don't normally allow this. Create a web based migration runner to run phinx migrations

**User Stories:**
- Deployers should be able to visit a URL (ie. mydomain.com/migrate) to run migrations

rjmackay:
@zhalloran could you look into this? it should be fairly straight forward..
ie. a stand alone PHP file ie. `httpdocs/migrate.php`
You could base in on something like https://github.com/robmorgan/phinx/blob/0.5.x-dev/app/web.php

It just needs to run phinx migrations. Though ideally we might also add an enable/disable flag in our .env file so that this is disabled by default.. its not quite a security risk. But not something you want to accidentally run on your deployment.

PR #1066, opened 2016-05-04
Feature/migration

zhalloran:
- Added migrate, migrate/status. migrate/migrate & migrate/rollback end points to the API (Do we want to have a rollback option?).
- Added Migrate controller class. Calls to this controller throw a 403 error if the user is not an admin.
- Added 'migrate' record to oauth_scopes tables

Test these changes by:
- execute signed api calls to end points
- tests still need to be written

Fixes ushahidi/platform#1003

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1066)

### DIFF
--- a/application/classes/Controller/Api/Migrate.php
+++ b/application/classes/Controller/Api/Migrate.php
@@ -25,6 +25,12 @@ class Controller_Api_Migrate extends Ushahidi_Rest {
 	public function action_get_index_collection($command = 'status')
 	{
 		
+		$user = service('session.user');
+		
+		if ('admin' != $user->role) {
+			throw new HTTP_Exception_403('Must be an admin to access this service');
+		}
+		
 		$commands = [
 		    'status'   => 'getStatus',
 		    'migrate'  => 'getMigrate',
--- a/application/classes/Controller/Api/Migrate.php
+++ b/application/classes/Controller/Api/Migrate.php
@@ -19,42 +19,26 @@ class Controller_Api_Migrate extends Ushahidi_Rest {
 
 	protected function _scope()
 	{
-		return 'config';
+		return 'migrate';
 	}
 
-	public function action_get_migrate()
+	public function action_get_index_collection()
 	{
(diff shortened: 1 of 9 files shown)


rjmackay:
Not following our typical structure - but I think migration would always be a special case.
Main thing we need to handle is the initial setup: when we have no tables at all.

---

Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[application/classes/Controller/Api/Migrate.php, line 27 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/1066#-KGxajL8CXw26U9bof-k:-KGxajL8CXw26U9bof-l:-1532934950) ([raw file](https://github.com/ushahidi/platform/blob/631442a55bba2a6e77925a5c2163c1b9cdab388b/application/classes/Controller/Api/Migrate.php#L27)):*
I don't know if it matters here, but in general controller in platform will usually just call out to a usecase, which then handles everything else. 

---

*[application/classes/Controller/Api/Migrate.php, line 30 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/1066#-KGxfmHX-p3Wwq53E88G:-KGxfmHX-p3Wwq53E88H:1052805060) ([raw file](https://github.com/ushahidi/platform/blob/631442a55bba2a6e77925a5c2163c1b9cdab388b/application/classes/Controller/Api/Migrate.php#L30)):*
We need handling for a special case where no tables exist at all - in which case oauth checks will fail, as will checking role.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ushahidi/platform/1066#-:-KGxgRBRccwwou8SN3F7:-564140620)*

zhalloran:
Awesome. Thanks for the feedback.

zhalloran:
@rjmackay Broke out migration/migrate into it's own class to allow for unauthenticated requests. Tests are...lacking.

rjmackay:
Added some barebones testing. Landing this thing

PR #1070, opened 2016-05-09
Feature/display collections

yuchenglin:
- Output collection names into entities from database. (To display them in the popup windows of map view)

Test these changes by:
- Update both ``/platform`` and ``/platform-client`` according to this pull request. There would be collection names in each popup windows.

Fixes ushahidi/platform# .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1070)

### DIFF
--- a/application/classes/Ushahidi/Formatter/Post/GeoJSONCollection.php
+++ b/application/classes/Ushahidi/Formatter/Post/GeoJSONCollection.php
@@ -47,17 +47,20 @@ class Ushahidi_Formatter_Post_GeoJSONCollection implements Formatter
 
 			if (! empty($geometries))
 			{
-				$set_name = [];
-				$set_ids = $entity->sets;
-
-				if ($set_ids != null) {
-					$query = DB::select('sets.name')
+				$collections = [];
+				$collection_ids = $entity->sets;
+				if ($collection_ids != null) {
+					$query = DB::select('sets.name','sets.id')
 								->from('sets')
-								->where('id', 'IN', $set_ids);
-					$set_names = $query->execute();
-					
-					foreach ($set_names as $tmp_set_name) {
-						$set_name[] = $tmp_set_name['name'];
+								->where('id', 'IN', $collection_ids)
+								->where('search', '=', 0);
+					$tmp_collections = $query->execute();
+					foreach ($tmp_collections as $tmp_collection) {
+						$collections [] = [
+							'id' => $tmp_collection['id'],
+							'name' => $tmp_collection['name'],
+							'url' => URL::site(Ushahidi_Rest::url('collections', $tmp_collection['id']), Request::current())
+						];
 					}
 				}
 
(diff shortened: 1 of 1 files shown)


yuchenglin:
[clabot:check]

application/classes/Ushahidi/Formatter/Post/GeoJSONCollection.php
```diff
@@ -47,6 +47,20 @@ public function __invoke($entities)
 
 			if (! empty($geometries))
 			{
+				$set_name = [];
+				$set_ids = $entity->sets;
+
+				if ($set_ids != null) {
+					$query = DB::select('sets.name')
```
rjmackay: I think this needs to differentiate between collections and saved searches. Both use the sets table in the backend - but the public api always references `collections` or `savedsearches`. Could you add them as 2 separate fields?

application/classes/Ushahidi/Formatter/Post/GeoJSONCollection.php
```diff
@@ -58,6 +72,7 @@ public function __invoke($entities)
 						'description' => $entity->content,
 						'id' => $entity->id,
 						'url' => URL::site(Ushahidi_Rest::url($entity->getResource(), $entity->id), Request::current()),
+						'set_name' => $set_name,
```
rjmackay: as above. Split this and rename to `collections` and `savedsearches`

rjmackay:
Hi @yuchenglin !
Thanks for the pull request. I've added some comments on the code for things I think need changing before this lands.
Any chance you could also add `collections` to the posts api (not just the geojson) too? 
Thanks!

application/classes/Ushahidi/Formatter/Post/GeoJSONCollection.php
rjmackay: Also - we'd normally only include the `id` for relations, not the name as well. Definitely not only the name.
Could you make your client changes work using just the collection `id`?

yuchenglin: I would revise part of the geojson output as below (which was `set_name`).
`"collections": [
                    {
                      "id": 1,
                      "name": "Taipei Fire"
                    },
                    {
                      "id": 2,
                      "name": "Taipei Flood"
                    }
                ]`
Is this format fit the requirement? Thanks

rjmackay: That'd be great. If you can also include URL that'd be perfect
We use ```URL::site(Ushahidi_Rest::url($resource, $id), Request::current())``` elsewhere to generate URLs.

yuchenglin:
Hi,
I have rename the misleading variables and include collection url. Please check it. Thanks!

yuchenglin:
Hi,
If anything is missing or incomplete, please inform me. Thank you!

rjmackay:
Sorry about the delay on this. Lost track of it.
The problem with this is that its doing a DB lookup in a formatter. We should only ever be doing DB lookups in the repository. If you want to do that and then format the data in the formatter I could accept that.

Issue #1048, opened 2016-05-04
Mustang: User filters


rjmackay:
- Create updated user filters:
  - Filter by roles
  - Filter by keyword (ie. query param `q`)
- Create a service to maintain state of users (or maybe a generic factory that can be reused for roles, categories etc?)

Each filter could be its own directive? or a single directive for 'user-filters'? However whatever is created here should ideally be reusable for filtering elsewhere.

Refer to mustang design for users view http://localhost:8000/assets/html/5_layouts/settings-users.html

rjmackay:
Complete alongside #1069

PR #1077, opened 2016-05-17
Updated user to allow role as array for filtering

rjmackay:
- Changes user repo to allow filtering by multiple roles

Test these changes by:
- On the setting/users page filter by roles

Fixes ushahidi/platform#1048

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1077)

### DIFF
--- a/application/classes/Ushahidi/Repository/User.php
+++ b/application/classes/Ushahidi/Repository/User.php
@@ -79,7 +79,7 @@ class Ushahidi_Repository_User extends Ushahidi_Repository implements
 	// SearchRepository
 	public function getSearchFields()
 	{
-		return ['email', 'role', 'q' /* LIKE realname, email */];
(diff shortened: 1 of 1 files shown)


application/classes/Ushahidi/Repository/User.php
```diff
@@ -79,7 +79,7 @@ public function update(Entity $entity)
 	// SearchRepository
 	public function getSearchFields()
 	{
-		return ['email', 'role', 'q' /* LIKE realname, email */];
+		return ['email', 'roles', 'q' /* LIKE realname, email */];
```
rjmackay: I don't think the param name should change. Our param names are still singular even when they allow multiple values.

rjmackay:
@willdoran assigned to you to revert the param name change. Then good to land.

PR #1276, opened 2016-06-28
Fix survey field validation.

### LINKED ISSUE #1271: when i add a single text field to the post - it asks for lat + long
### Expected behavior
showing up the text i right
### Actual behaviour
screen shot 2016-06-25 at 5 01 14 pm
### Steps to reproduce the behaviour/error
add a sentence field to a post and then just put text in it when you create a point on the map

rjmackay:
Validate values against the correct field. We were accidentally
validating everything against a default message_location field

Fixes #1271

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1276)

### DIFF
--- a/application/classes/Ushahidi/Validator/Post/Decimal.php
+++ b/application/classes/Ushahidi/Validator/Post/Decimal.php
@@ -13,7 +13,7 @@ class Ushahidi_Validator_Post_Decimal extends Ushahidi_Validator_Post_ValueValid
 {
 	protected function validate($value)
 	{
-		if (!Valid::decimal($value)) {
+		if (!Valid::numeric($value)) {
 			return 'decimal';
 		}
 	}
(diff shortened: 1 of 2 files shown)


application/classes/Ushahidi/Validator/Post/Decimal.php
```diff
@@ -13,7 +13,7 @@ class Ushahidi_Validator_Post_Decimal extends Ushahidi_Validator_Post_ValueValid
 {
 	protected function validate($value)
 	{
-		if (!Valid::decimal($value)) {
+		if (!Valid::numeric($value)) {
```
willdoran: @rjmackay I'm not getting an error when I enter 1, it should tell  me that I have to enter a decimal value but it's just passing, is that desired behaviour now, that a decimal can also not have a number of decimal places?

The field is also no longer displaying on the post detail.

rjmackay: Switching to accept `1` was intentional since `1.0 == 1` so I don't think users care. We really just have decimal to accept any number at all while Integer only accepts integers.

Issue #1282, opened 2016-07-04
Feature: Add require approval and who can edit controls on surveys settings


rjmackay:
Changes:
- [x] Add toggle for 'require posts be reviewed before they're published' to survey settings
- [x] Add control for 'who can add to this survey' to survey settings
- [x] API: Add `require_approval` field to Form entity
- [x] API: Add `createable_by` field to Form entity (ie. which roles can add to a survey)

Background: 
https://github.com/ushahidi/platform/issues/1263#issuecomment-228464926
https://github.com/ushahidi/platform/issues/1262#issuecomment-228815980

---

## [Survey settings](http://github.ushahidi.org/platform-pattern-library/assets/html/5_layouts/settings-survey-edit.html)

Part of a survey's definition is who can add posts to it and whether posts requires review before they're published:

![survey settings](https://cloud.githubusercontent.com/assets/1136279/16395394/3b003582-3c7f-11e6-84a7-c530a5e6eab1.png)

zhalloran:
@rjmackay I'd prefer to set the default value for require_approval to false, but...for existing deployments, the default is to require approval (status => draft in posts). Any preference for how to approach this?

zhalloran:
@rjmackay @brandonrosage For who can add to a survey the example has the logged in user, Everyone, or Specific Roles as options. Do we intend to allow the ability to select individual users as well? Also, do we have screens for the role selection action?

rjmackay:
My take is that require_approval should default to `true`. Because thats what we already have and I think also whats expected. I worry we surprise deployers otherwise by letter anonymous posts go public without review.
@jshorland any thought?

Who can add to a survey should only allow roles, not users to be selected. You can look at http://preview.ushahidi.com/platform-pattern-library/develop/assets/html/5_layouts/settings-survey-edit.html to see how the UI should work.

zhalloran:
@rjmackay Perfect. I must have been looking in the wrong branch for the layout. Which branch is this?

rjmackay:
Ah right. We're just sorting out branches in the PL.. there's a `master` and `develop` branch now. This is still in `develop`.

rjmackay:
From discussion with @jshorland. The UI has updated slightly to include a ‘who can edit posts in this survey’ too. But for this iteration we can continue with just the ‘who can add to this survey’ control for this iteration and add who can edit later (probably along with changes to tasks). Make sense?

zhalloran:
[...]

PR #1318, opened 2016-07-29
Feature/add review workflow toggle 1282

zhalloran:
- Add require_approval Boolean field to Form
- Add all_roles Boolean field to Form
- Add API end-point for adding & retrieving a list of Roles that are allowed to post to a specific Form (i.e. /forms/:form_id/roles)

Test these changes by:
- Creating a Form, updating a Form, calling POST /forms/:form_id/roles with payload `{ 'roles': [:role_id]}

Fixes ushahidi/platform#1282 .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1318)


rjmackay:
Reviewed 21 of 21 files at r1.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks broke.

---

*[application/classes/Controller/Api/Forms/Roles.php, line 27 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KOC0KyKO1DBpK5qlHv3:-KOC0KyLcGWHyv5mbn8E:-1286030326) ([raw file](https://github.com/ushahidi/platform/blob/4df5537ee4560281d6ecd40627ee4ca72c9cecac/application/classes/Controller/Api/Forms/Roles.php#L27)):*
> ```PHP
> 	protected function _identifiers()
> 	{
> 		return parent::_identifiers() + [
> ```

The `parent::_identifiers()` call should pass ALL params through already.

---

*[application/classes/Ushahidi/Formatter/Form/Role.php, line 14 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KOC0bNTxZY4v40agFPL:-KOC0bNUY3SgwUjeqmX1:-1133346421) ([raw file](https://github.com/ushahidi/platform/blob/4df5537ee4560281d6ecd40627ee4ca72c9cecac/application/classes/Ushahidi/Formatter/Form/Role.php#L14)):*
> ```PHP
> use Ushahidi\Core\Traits\FormatterAuthorizerMetadata;
> 
> class Ushahidi_Formatter_Form_Role extends Ushahidi_Formatter_API
> ```

If you can't get this to work with Entities you can just implement that `Formatter` interface instead of extending and accept an array instead.

---

*[migrations/20160720230341_add_require_approval_to_forms.php, line 7 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KOC1UY_6fH1BTndZ0zO:-KOC1UYayl2q9RWuO5uE:-712744445) ([raw file](https://github.com/ushahidi/platform/blob/4df5537ee4560281d6ecd40627ee4ca72c9cecac/migrations/20160720230341_add_require_approval_to_forms.php#L7)):*
> ```PHP
> class AddRequireApprovalToForms extends AbstractMigration
> {
>     
> ```

indenting in this files is odd. Mixed spaces and tabs.

---

*[migrations/20160722231016_create_form_roles_table.php, line 12 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KOC1_gaF9q5ANkbqsMU:-KOC1_gaF9q5ANkbqsMV:-970801787) ([raw file](https://github.com/ushahidi/platform/blob/4df5537ee4560281d6ecd40627ee4ca72c9cecac/migrations/20160722231016_create_form_roles_table.php#L12)):*
> ```PHP
> 			->addColumn('form_id', 'integer', ['null' => false])
> 			->addColumn('role_id', 'integer', ['null' => false])
> 			->addForeignKey('form_id', 'forms', 'id', array('delete'=> 'CASCADE', 'update'=> 'NO_ACTION'))
> ```

Shouldn't update's cascade too?

---

*[migrations/20160726225434_add_form_all_roles.php, line 13 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KOC29FuKL5oL5tbPN2W:-KOC29FuKL5oL5tbPN2X:-1172439899) ([raw file](https://github.com/ushahidi/platform/blob/4df5537ee4560281d6ecd40627ee4ca72c9cecac/migrations/20160726225434_add_form_all_roles.php#L13)):*
> ```PHP
>     {
> 		$this->table('forms')
>             ->addColumn('all_roles', 'boolean', [
> ```

This might be better with a more explicit name ie. `everyone_can_create` .. minor though

---

*[src/Core/Entity/Form.php, line 28 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KOC1vyQqnOeOPOSOnaw:-KOC1vyQqnOeOPOSOnax:-1698223237) ([raw file](https://github.com/ushahidi/platform/blob/4df5537ee4560281d6ecd40627ee4ca72c9cecac/src/Core/Entity/Form.php#L28)):*
> ```PHP
> 	protected $updated;
> 	protected $require_approval;
> 	protected $all_rolls;
> ```

`roles` not `rolls`

---

*[src/Core/Entity/Form.php, line 49 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KOC1yWn2Zsvv5FMRXVt:-KOC1yWn2Zsvv5FMRXVu:-1698223237) ([raw file](https://github.com/ushahidi/platform/blob/4df5537ee4560281d6ecd40627ee4ca72c9cecac/src/Core/Entity/Form.php#L49)):*
> ```PHP
> 			'updated'     => 'int',
> 			'require_approval'  => 'bool',
> 			'all_rolls'   => 'bool',
> ```

`roles` not `rolls`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ushahidi/platform/1318)*

rjmackay:
Reviewed 6 of 11 files at r2.
Review status: 20 of 24 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[src/Core/Tool/Authorizer/PostAuthorizer.php, line 203 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KOm4GsFbYIVFN752cUK:-KOm4GsGT_itj567CeDd:bfashar) ([raw file](https://github.com/ushahidi/platform/blob/46572b158769a9187977788e7a4ee9b2e91a6149/src/Core/Tool/Authorizer/PostAuthorizer.php#L203)):*
> ```PHP
> 		// we check to see if the Form & Role Join exists in the `FormRoleRepository`
> 		if ($entity->form_id) {
> 			$form = $this->form_repo->get($entity->form_id);
> ```

This should be a single repo call - the rest should be handed off to the repo itself.
I would suggest:
- Only inject the Form Repo
- Define a method in the Form Repo interface of `canRoleCreate($roleName)`
- That method should be able to handle the lookup in a single query. Currently this is running 3 queries and putting too much logic into the authorizer.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ushahidi/platform/1318)*

zhalloran:
Review status: 13 of 24 files reviewed at latest revision, 6 unresolved discussions.

---

*[application/classes/Controller/Api/Forms/Roles.php, line 27 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KOC0KyKO1DBpK5qlHv3:-KOmiGNYJ02RuI1ZSFYl:br8zpyr) ([raw file](https://github.com/ushahidi/platform/blob/4df5537ee4560281d6ecd40627ee4ca72c9cecac/application/classes/Controller/Api/Forms/Roles.php#L27)):*
<details><summary><i>Previously, rjmackay (Robbie Mackay) wrote…</i></summary>
> The `parent::_identifiers()` call should pass ALL params through already.
</details>
Probably a holdover from when I was trying to get this to work before I added `roles` to the FormRole Entity. It's been removed.

---

*[application/classes/Ushahidi/Formatter/Form/Role.php, line 14 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KOC0bNTxZY4v40agFPL:-KOmik4gF9AvfOkex2HQ:b8hzafn) ([raw file](https://github.com/ushahidi/platform/blob/4df5537ee4560281d6ecd40627ee4ca72c9cecac/application/classes/Ushahidi/Formatter/Form/Role.php#L14)):*
<details><summary><i>Previously, rjmackay (Robbie Mackay) wrote…</i></summary>
> If you can't get this to work with Entities you can just implement that `Formatter` interface instead of extending and accept an array instead.
</details>
Resolved last week.

---

*[migrations/20160720230341_add_require_approval_to_forms.php, line 7 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KOC1UY_6fH1BTndZ0zO:-KOmhdD1dHt5wI1vYelU:bq6uw19) ([raw file](https://github.com/ushahidi/platform/blob/4df5537ee4560281d6ecd40627ee4ca72c9cecac/migrations/20160720230341_add_require_approval_to_forms.php#L7)):*
<details><summary><i>Previously, rjmackay (Robbie Mackay) wrote…</i></summary>
> indenting in this files is odd. Mixed spaces and tabs.
</details>
Fixed in new commit.

---

*[migrations/20160722231016_create_form_roles_table.php, line 12 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KOC1_gaF9q5ANkbqsMU:-KOmdFzlSnvSK4wNtPte:b-1u4xf4) ([raw file](https://github.com/ushahidi/platform/blob/4df5537ee4560281d6ecd40627ee4ca72c9cecac/migrations/20160722231016_create_form_roles_table.php#L12)):*
<details><summary><i>Previously, rjmackay (Robbie Mackay) wrote…</i></summary>
> Shouldn't update's cascade too?
</details>
Cascades for updates to the `form` record are not required. We don't care about any changes to the `form` record except for deletion.

---

*[src/Core/Tool/Authorizer/PostAuthorizer.php, line 203 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KOm4GsFbYIVFN752cUK:-KOmix-IYo23IgcZktOW:bp3ni34) ([raw file](https://github.com/ushahidi/platform/blob/46572b158769a9187977788e7a4ee9b2e91a6149/src/Core/Tool/Authorizer/PostAuthorizer.php#L203)):*
<details><summary><i>Previously, rjmackay (Robbie Mackay) wrote…</i></summary>
> This should be a single repo call - the rest should be handed off to the repo itself.
> I would suggest:
> - Only inject the Form Repo
> - Define a method in the Form Repo interface of `canRoleCreate($roleName)`
> - That method should be able to handle the lookup in a single query. Currently this is running 3 queries and putting too much logic into the authorizer.
</details>
I'll look at this tomorrow.

---

*[migrations/20160726225434_add_form_all_roles.php, line 13 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KOC29FuKL5oL5tbPN2W:-KOmeISEgHBU8iHvrQMo:bsyqcgs) ([raw file](https://github.com/ushahidi/platform/blob/4df5537ee4560281d6ecd40627ee4ca72c9cecac/migrations/20160726225434_add_form_all_roles.php#L13)):*
<details><summary><i>Previously, rjmackay (Robbie Mackay) wrote…</i></summary>
> This might be better with a more explicit name ie. `everyone_can_create` .. minor though
</details>
Seems reasonable. If this is a change we want to make, might as well make it now. The commit has been pushed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ushahidi/platform/1318)*

rjmackay:
Reviewed 1 of 11 files at r2, 2 of 12 files at r3.
Review status: 15 of 25 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/ushahidi/platform/1318)*

rjmackay:
The more I dig into this the more issues I have.. the FormRole entity is doubling as a collection rather than just an individual resource. Even then its not really an entity at all, its just a join table.. and a property of the form itself anyway.
This causes issues throughout as the semantics just don't work.
TBH I still think this should just be saved from the Form endpoint + repo. It should definitely still save into a join table and not use the horrible schema we've used elsewhere (ie. JSON array of roles). But the additional endpoint just adds problems and tech debt.

We *can* make the endpoint work with less hacky-ness if we use a collection formatter, that takes away much of the uglyness. I'm just not sure its worth it.

---

Reviewed 1 of 11 files at r2, 7 of 12 files at r3, 5 of 7 files at r4.
Review status: 24 of 26 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[application/classes/Ushahidi/Formatter/Form/Role.php, line 21 \[r4\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KQYh9W_FkEo9VKZA_OP:-KQYh9W_FkEo9VKZA_OQ:b-2oxbva) ([raw file](https://github.com/ushahidi/platform/blob/ca1f07bcd7eff129ce2aea7573d9afe3f5c7a0fb/application/classes/Ushahidi/Formatter/Form/Role.php#L21)):*
> ```PHP
> 	{
> 		$data = [
> 			'id'  => $entity->id,
> ```

I don't think these should have their own ID, they're just a join between forms and roles. Elsewhere we use a structure like forms/:formid/roles/:roleid .. we should match that here (if we even provide that url at all)

---

*[application/classes/Ushahidi/Repository/Form.php, line 30 \[r4\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KQYukVCMllHAzAeKR15:-KQYukVCMllHAzAeKR16:b-l1ugap) ([raw file](https://github.com/ushahidi/platform/blob/ca1f07bcd7eff129ce2aea7573d9afe3f5c7a0fb/application/classes/Ushahidi/Repository/Form.php#L30)):*
> ```PHP
>     public function getEntity(Array $data = null)
>     {
> 	    if (isset($data["id"])) {
> ```

This is kinda messy - ideally this should just use a method which returns exactly the data needed here, not assign to 2 temp variables on the way.

---

*[application/classes/Ushahidi/Repository/Form.php, line 108 \[r4\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KQYxl_4vemIpyZK6dVv:-KQYxl_4vemIpyZK6dVw:bt3x6pv) ([raw file](https://github.com/ushahidi/platform/blob/ca1f07bcd7eff129ce2aea7573d9afe3f5c7a0fb/application/classes/Ushahidi/Repository/Form.php#L108)):*
> ```PHP
> 
>         foreach($results as $role) {
>             if (!is_null($role['name'])) {
> ```

it seems weird that we're returning role names here and attaching that to `can_add` but we use role_id on the form roles endpoint

---

*[application/classes/Ushahidi/Repository/Post.php, line 818 \[r4\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KQYfRsXFELyyEkP195d:-KQYfRsXFELyyEkP195e:b-88v4u1) ([raw file](https://github.com/ushahidi/platform/blob/ca1f07bcd7eff129ce2aea7573d9afe3f5c7a0fb/application/classes/Ushahidi/Repository/Post.php#L818)):*
> ```PHP
> 		unset($post['values'], $post['tags'], $post['completed_stages'], $post['sets'], $post['source'], $post['color']);
> 
> 		// Check if form_id is p
> ```

Truncated comment?

---

*[application/classes/Ushahidi/Repository/Post.php, line 824 \[r4\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KQYfpaOX_K-ze8nzFbK:-KQYfpaOX_K-ze8nzFbL:b-wqddp5) ([raw file](https://github.com/ushahidi/platform/blob/ca1f07bcd7eff129ce2aea7573d9afe3f5c7a0fb/application/classes/Ushahidi/Repository/Post.php#L824)):*
> ```PHP
> 			if (!$form->require_approval)
> 			{
> 				$post['status'] = 'published';
> ```

This doesn't work in every instance: if an admin creates a post they should see the UI to set published/archived/draft during creation, but this would then overwrite their choice.
We should probably only set this here if its not set already.

Also need to set it to `draft` in an else condition.

We should also be updating the validator 

---

*[migrations/20160722231016_create_form_roles_table.php, line 12 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KOC1_gaF9q5ANkbqsMU:-KQYgClHE7jj3Ps36ktA:b-ozf0xu) ([raw file](https://github.com/ushahidi/platform/blob/4df5537ee4560281d6ecd40627ee4ca72c9cecac/migrations/20160722231016_create_form_roles_table.php#L12)):*
<details><summary><i>Previously, zhalloran (Zack Halloran) wrote…</i></summary>
> Cascades for updates to the `form` record are not required. We don't care about any changes to the `form` record except for deletion.
</details>
While its unlikely we need to care about updates, we *might* care in future (ie. when changing IDs to migrate from multi db to single db). Adding a cascade here is simple and matches our choices elsewhere.

---

*[migrations/20160722231016_create_form_roles_table.php, line 9 \[r4\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KQYkrrMj_z2KVuqqC7l:-KQYkrrNffkeYlqfgN35:b-bzsdco) ([raw file](https://github.com/ushahidi/platform/blob/ca1f07bcd7eff129ce2aea7573d9afe3f5c7a0fb/migrations/20160722231016_create_form_roles_table.php#L9)):*
> ```PHP
>     public function change()
>     {
>         $this->table('form_roles')
> ```

I don't think this table needs its own ID (at least not right now) we should add
```
        $this->table('form_roles', [
                'id' => false,
                'primary_key' => ['form_id', 'role_id'],
            ])
```

---

*[src/Core/Entity/Form.php, line 51 \[r4\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KQYm91DBCbOhQK9Hv2O:-KQYm91DBCbOhQK9Hv2P:b2z17f9) ([raw file](https://github.com/ushahidi/platform/blob/ca1f07bcd7eff129ce2aea7573d9afe3f5c7a0fb/src/Core/Entity/Form.php#L51)):*
> ```PHP
> 			'require_approval'  => 'bool',
> 			'everyone_can_create'   => 'bool',
> 			'can_add'     => 'array',
> ```

Consistency: we have can_add but then we have everyone_can_create.
Lets make can_add into can_create

---

*[src/Core/Entity/FormRole.php, line 27 \[r4\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KQYjrhfZd15229vSrb9:-KQYjrhfZd15229vSrbA:b-4uq4bu) ([raw file](https://github.com/ushahidi/platform/blob/ca1f07bcd7eff129ce2aea7573d9afe3f5c7a0fb/src/Core/Entity/FormRole.php#L27)):*
> ```PHP
> 	{
> 		return [
> 			'id'       => 'int',
> ```

Does this table actually have an ID?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ushahidi/platform/1318#-:-KQYtwImJdXZEw0U29fZ:brnav54)*

rjmackay:
Hacked my way through some changes to how we save form roles - primarily to avoid abusing the existing $repo->update() pattern and associated validators etc, by having them save and validate multiple items. Now we have a validator process a single entity, and use a distinct interface to save the entire collection at once.

rjmackay:
Review status: 8 of 30 files reviewed at latest revision, 10 unresolved discussions.

---

*[application/classes/Ushahidi/Formatter/Form/Role.php, line 21 \[r4\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KQYh9W_FkEo9VKZA_OP:-KQbgPH_McbV4nXs2itV:b-h868uc) ([raw file](https://github.com/ushahidi/platform/blob/ca1f07bcd7eff129ce2aea7573d9afe3f5c7a0fb/application/classes/Ushahidi/Formatter/Form/Role.php#L21)):*
<details><summary><i>Previously, rjmackay (Robbie Mackay) wrote…</i></summary>
> I don't think these should have their own ID, they're just a join between forms and roles. Elsewhere we use a structure like forms/:formid/roles/:roleid .. we should match that here (if we even provide that url at all)
</details>
turned out to be hard to get rid of. Ignoring

---

*[application/classes/Ushahidi/Repository/Form.php, line 30 \[r4\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KQYukVCMllHAzAeKR15:-KQbgSTsC-08WkUkxjG2:b-nyxpe6) ([raw file](https://github.com/ushahidi/platform/blob/ca1f07bcd7eff129ce2aea7573d9afe3f5c7a0fb/application/classes/Ushahidi/Repository/Form.php#L30)):*
<details><summary><i>Previously, rjmackay (Robbie Mackay) wrote…</i></summary>
> This is kinda messy - ideally this should just use a method which returns exactly the data needed here, not assign to 2 temp variables on the way.
</details>
Cleaned a little

---

*[application/classes/Ushahidi/Validator/Post/Create.php, line 171 \[r7\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KQc0m3I2aQpSgDyuZ48:-KQc0m3I2aQpSgDyuZ49:b7qyvql) ([raw file](https://github.com/ushahidi/platform/blob/a0fd5d87606aa8960d5ac634aa8bb8a6190ed36c/application/classes/Ushahidi/Validator/Post/Create.php#L171)):*
> ```PHP
> 	}
> 
> 	public function checkApprovalRequired (Validation $validation, $status, $data)
> ```

Not super happy with this but its better here than the authorizer.

---

*[src/Core/Tool/Authorizer/PostAuthorizer.php, line 128 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform/1318#-KQc0t0jyHnvKdkS7SeC:-KQc0t0jyHnvKdkS7SeD:btdyg5o) ([raw file](https://github.com/ushahidi/platform/blob/bfabd934c6e0034638d9b20c55543a719cf2c03b/src/Core/Tool/Authorizer/PostAuthorizer.php#L128)):*
> ```PHP
>         }
> 
>         // Non-admin users are not allowed to change post status
> ```

Moved to validation

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ushahidi/platform/1318)*

Issue #1007, opened 2016-04-05
Install: support install into a subdirectory


rjmackay:
- Deployers should be able to install the API into a subdirectory ie. mydomain.com/ushahidi rather than having to be at the root of the domain
- Deployers should be able to install the client into a subdirectory ie. mydomain.com/ushahidiui rather than having to be at the root of the domain

rjmackay:
@jasonmule assigning this to you. Feel free to delegate to anyone else as needed (including myself).

jasonmule:
@tuxpiper Do you want to take a look at this?

tuxpiper:
Sure, I can take a look

tuxpiper:
As far as I can tell, there is nothing in the code restricting this now. I've got single virtual hosts working locally without any issues. This would be purely a matter of documenting this kind of setup.

I guess the question is.. should we update our docs to recommend this setup, or just add it as an option? How should we make that decision? @rjmackay @aoduor 

Copying my nginx/php-fpm setup here so you can see how simple it is:

```
server {
	listen 80 default_server;
	server_name _;

	index index.html;
	root /var/app/server/www;

	location ~ /(api|oauth)/.* {
		# Forward api requests to fastcgi interface to PHP
		rewrite ^(.*)$ /index.php$1;

		fastcgi_split_path_info ^/(.+\.php)(/.+)$;

		include fastcgi_params.conf;

		fastcgi_param SCRIPT_FILENAME           $fastcgi_script_name;
		fastcgi_param PATH_INFO                 $fastcgi_path_info;
		fastcgi_param HTTP_HOST                 $host;
		fastcgi_param SERVER_NAME               $host;
		fastcgi_intercept_errors                on;
		fastcgi_index                           index.php;
		fastcgi_pass                            api_backend;

		break;
	}

	location / {
		try_files $uri $uri/ @missing;
	}

	# Rewrite 404s back to index.html for pushState support
	# All routing is handled by Angular.
	location @missing {
		rewrite ^ /index.html last;
	}

}
```

The apache config wouldn't be more complicated

rjmackay:
I think the recommended set up should be 2 vhosts - but we should document this as a clear alternative.
Deployers on shared hosts are most likely to be successful with a single zip/tarball they can upload, with a .htaccess ready to go.

rjmackay:
@tuxpiper could you document this and other parts of install from bundled release and add that to the ushahidi.com/support pages?

rjmackay:
[...]

PR #1328, opened 2016-08-04
ushahidi/platform#1007 workaround for hosts that mangle the PATH_INFO wierdly (common in php-cgi)

tuxpiper:
-
Add a workaround to be able to set the `$_SERVER` array key where the URI to be parsed by the API should be gotten from.

The need for this came up as I was trying to install platform in a cpanel host with php-cgi (which is known for removing `$_SERVER['PATH_INFO']` and providing `$_SERVER['ORIG_PATH_INFO']` instead).

By setting `SERVER_PATHINFO_VAR=ORIG_PATH_INFO` in  the`.env` this would be worked around.

Test these changes by:
-
Installing platform a cpanel/php-cgi host :)

Fixes ushahidi/platform#1007 . Making single folder installation in shared hosting kind of environments more possible.

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1328)

### DIFF
--- a/application/bootstrap.php
+++ b/application/bootstrap.php
@@ -129,6 +129,8 @@ if (getenv("RAVEN_URL"))
 	$error_handler->registerExceptionHandler();
 	$error_handler->registerErrorHandler();
 	$error_handler->registerShutdownFunction();
+
+	Kohana::$log->attach(new Log_Raven($client));
 }
 
 /**
--- a/application/classes/Log/Raven.php
+++ b/application/classes/Log/Raven.php
@@ -1,0 +1,79 @@
+<?php defined('SYSPATH') or die('No direct script access.');
+/**
+ * Raven log writer. Writes out messages and stores them in Sentry.
+ *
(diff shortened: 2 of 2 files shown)


httpdocs/index.php
```diff
@@ -20,12 +20,21 @@ class_exists('Minion_Task') OR die('Please enable the Minion module for CLI supp
 	Minion_Task::factory(Minion_CLI::options())->execute();
 }
 else
-{
+{	
+	/* You may set SERVER_PATHINFO_VAR in your .env file with the name of the $_SERVER array
```
rjmackay: Can we include an example htaccess line here too?

rjmackay:
LGTM ..
however if we know `ORIG_PATH_INFO` is a common case, could we just check that envvar directly?

tuxpiper:
Rebasing against develop in a new PR (#1331)

Issue #1366, opened 2016-09-08
CSV import improvements


rjmackay:
- We removed the configure posts step from CSV import
- We need to re-instate this but ensure the UX is more inline with the pattern library.

rjmackay:
Assigned to myself to figure out if there are any issues with the current PL designs.

rjmackay:
Suggested minimal version of this:
- Add global toggle for published, or option to map to CSV column
- Add missing mapping for CSV column to categories, post status, etc
- Switch mapping columns so we list all post fields, and dropdown of CSV columns

brandonrosage:
@jshorland, @rjmackay: I've staged the three layouts for the revised CSV import flow in the `sprint-11` branch. You can sample them like this:

* [Step 1](http://preview.ushahidi.com/platform-pattern-library/sprint-11/assets/html/5_layouts/settings-import.html): The screen you see when you first visit "Import."
* [Step 2](http://preview.ushahidi.com/platform-pattern-library/sprint-11/assets/html/5_layouts/settings-import-configure.html): The screen you use to configure things like post status, field values, and categories.
* [Step 3](http://preview.ushahidi.com/platform-pattern-library/sprint-11/assets/html/5_layouts/settings-import-complete.html): The screen that, if the import is still processing, tells you as much. The layout includes "nearly complete" and "complete" states. In any case, the idea is that, when the process is finished, you're greeted with a message that includes an optional link to a Timeline view of the posts that were just imported.

jshorland:
@brandonrosage What's the best way to provide feedback on the designs

brandonrosage:
No. 1, if there are needs and requirements that are not addressed, please call those out in the comments here.

No. 2, if you have questions (about how it's supposed to work in various situations, or if something doesn't appear to be working), please toss those in the comments, too.

No. 3, if you have feedback about the visual design (e.g. it's not appealing or too messy), please make a comment. 

In any of those cases, It helps me a lot to have them captured in this thread (rather than something more abstract, like InVision comments).

jshorland:
got it. 

On screen for step 2, if the user chooses to map a column to set status and/or categories, is it possible to have them complete this at the same time they map the other fields? I believe this would require some back end complications of populating the mapping fields after the user selects the methods of configuring these two. But to me, it's confusing/redundant to do it separately rather than just all at once in the mapping step.

jshorland:
[...]

PR #1443, opened 2016-10-13
Return created entities ids from import

willdoran:
- Adding return of created entity ids to import use case

Test these changes by:
-

Fixes ushahidi/platform#1366

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1443)


rjmackay:
LGTM .. my only suggestion is that we create the collection on the backend too and return the collection id

Issue #1469, opened 2016-10-31
Survey fields: Don't assume Title & Description


brandonrosage:
Surveys should no longer assume "Title" and "Description" fields. Users should be able to name and configure all fields -- no matter which ordered position they're in -- however they like.

In an effort to lay the foundation for [a larger "Free-form surveys" solution](https://github.com/ushahidi/platform/issues/1335#issuecomment-256446871), we need to scope this one of two ways:

**1. Make surveys fully customizable**
This would call for new surveys to be populated with only one field -- an "Untitled" short text field -- that can be turned into any of the available field types. 

It would also call for any existing surveys with Title and Description fields to treat them as equal citizens: They can be re-ordered and changed into different field types (naturally requiring us to re-format existing values appropriately, and warning deployers of that implication before confirming such a change).

It would also call for the addition of a single toggle switch/checkbox attached to each field that indicated whether that field should be included in "postcards" for that survey.

Finally, it'd call for postcards to honor this flexibility by adopting the logic [outlined in the full "free-form surveys" solution](https://github.com/ushahidi/platform/issues/1335#issuecomment-256446871), where they display the values for fields that are configured to be shown in postcards (using the same order directed by the survey overall).

**2. Just allow users to rename the 'Title' & 'Description' labels**
Their respective field types would remain fixed, as short- and long-text fields. And they'd continue to be fields you couldn't re-order.

-----

Demonstrations of each of the aforementioned field controls aren't yet staged in the PL, but can be very quickly depending on the path the development team chooses to take here.

rjmackay:
@willdoran do you think option 1 is doable in the next 2 weeks? Would option 2 be much better?

rjmackay:
Complications:
- SMS messages have no title
- Data sources
     - Where do we save sms message/email title and subject?
- Searching ?q=blah now needs to search all text/varchar fields
- Converting unknown post to a survey
    - No longer guaranteed to have title + description field
    - Need to map fields between SMS and survey etc. And do this bulk
- Post export
    - No common fields between surveys anymore, but should be fine
- Can I just save a totally empty post now? (ie. not required title + description)

rjmackay:
@brandonrosage @jshorland I talked with @willdoran and it looks like Option 2 is the only thing feasible in our timeline. As we went through it we realised the huge complication is how where does data from SMS/Email go if we don't have fixed title and description fields..

brandonrosage:
[...]

PR #1500, opened 2016-11-10
Allow edit of default survey fields

willdoran:
- Adds two new post value types covering Title and Description
- Adds associated repos and validators
- Next: add migration

Fixes ushahidi/platform#1469

Ping @ushahidi/platform

…ble post title and descriptions

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1500)

### DIFF
--- a/application/classes/Ushahidi/Core.php
+++ b/application/classes/Ushahidi/Core.php
@@ -635,7 +635,7 @@ abstract class Ushahidi_Core {
 					'point'    => $di->lazyGet('validator.post.point'),
 					'relation' => $di->lazyGet('validator.post.relation'),
 					'varchar'  => $di->lazyGet('validator.post.varchar'),
-					'title'  => $di->lazyGet('validator.post.title'),
+					'title'    => $di->lazyGet('validator.post.title'),
 					'media'    => $di->lazyGet('validator.post.media'),
 				],
 			];
--- a/application/classes/Ushahidi/Core.php
+++ b/application/classes/Ushahidi/Core.php
@@ -467,7 +467,9 @@ abstract class Ushahidi_Core {
 		$di->set('repository.post.point', $di->lazyNew('Ushahidi_Repository_Post_Point'));
 		$di->set('repository.post.relation', $di->lazyNew('Ushahidi_Repository_Post_Relation'));
 		$di->set('repository.post.text', $di->lazyNew('Ushahidi_Repository_Post_Text'));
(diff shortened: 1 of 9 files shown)


application/classes/Ushahidi/Core.php
```diff
@@ -467,7 +467,9 @@ public static function init()
 		$di->set('repository.post.point', $di->lazyNew('Ushahidi_Repository_Post_Point'));
 		$di->set('repository.post.relation', $di->lazyNew('Ushahidi_Repository_Post_Relation'));
 		$di->set('repository.post.text', $di->lazyNew('Ushahidi_Repository_Post_Text'));
+		$di->set('repository.post.description', $di->lazyNew('Ushahidi_Repository_Post_Description'));
```
rjmackay: I thought your plan was to still save title and description into the posts table? or did that not work as well as expected?

migrations/20161110175958_update_post_title_desc.php
```diff
@@ -0,0 +1,69 @@
...
+
+        $insert_title = $pdo->prepare(
+            "INSERT into
+                form_attributes
+                (`label`, `type`, `required`, `priority`, `cardinality`, `input`, `key`, `form_stage_id`)
+              VALUES
+                ('Title', 'title', 1, 0, 0, 'varchar', :key, :form_stage_id)"
+        );
+
+        $insert_desc = $pdo->prepare(
```
rjmackay: This could be combined into a single insert with title.. just need to use `:title_key` and `:desc_key`

migrations/20161110175958_update_post_title_desc.php
```diff
@@ -0,0 +1,69 @@
...
+     * Migrate Up.
+     */
+    public function up()
+    {
+        $pdo = $this->getAdapter()->getConnection();
+
+        $rows = $this->fetchAll(
+            "SELECT id
+                FROM form_stages
+                WHERE priority = 0"
```
rjmackay: This should be type = 'post'

rjmackay:
- Need to block deletion of title/description attributes in UI
- Need to throw an error/ignore data when someone sends a values for the title/description attribute.

migrations/20161110175958_update_post_title_desc.php
willdoran: Done

migrations/20161110175958_update_post_title_desc.php
willdoran: Done

Issue #1509, opened 2016-11-16
Duplicate emails being pulled into platform


jshorland:
### Expected behaviour
Each unique email should only be pulled into platform once. 

### Actual behaviour
User xenowatch.ushahidi.io has thousands of duplicate emails being pulled into their platform.

@willdoran  can you write notes as appropriate?

willdoran:
Yep, the main issue with this is the way we are using IMAP, unfortunately at present we end up redownloading what is already in the mail box IF the mail provider doesn't treat access as having been read. So we need to change this, so that we store the UID of the messages we have retrieved, we then each time ask for all the messages with UID > last_uid. We are not a full mail client so we don't take account of any intelligent behaviours such as awareness of deletion.

jshorland:
Still happening. The test deployment I created keeps bringing in the most recent email over and over and over and over

jshorland:
@willdoran can we hand this to @Angamanga to finish up?

willdoran:
Ah yes! I got entirely drawn into rollcall and this fell out of my head. I'll take @Angamanga through the issue and the needed fix.

rjmackay:
@jshorland is it *just* the most recent email? or many recent emails?
I wonder if this is just an off-by-one error ie. we're saving the ID of the last email we brought in, and rather than doing `id > lastid` we're doing `id >= lastid` or something like that.

willdoran:
@rjmackay This is actually an issue with the mail fetcher, if there are no new emails and you ask for emails from last_uid to last_uid + N, then it just returns the last mail again :( Always. So I suggested to @Angamanga that the best way to handle it is to add a check that what's returned doesn't equal the last uid.

rjmackay:
I think @Angamanga fixed this. @jshorland can you retest?

Angamanga:
@rjmackay @jshorland Yes, this was fixed in https://github.com/ushahidi/platform/pull/1536.

jshorland:
Great! I' going to check in with the user who reported this and let them know. I'll close after they confirm fixed. And to double check @Angamanga since this is fixed, their duplicate emails within their deployment should be cleared out, correct?

jshorland:
Fixed. Closing.

PR #1513, opened 2016-11-17
Email receive improvement

willdoran:
- Add function to retrieve most recent email uid
- Updated email provider to request emails based on a range

Test these changes by:
- [ ] Setup an email provider
- [ ] Attempt to import emails, it should import the first X number
- [ ] Attempt to import emails, it should import the next X number

Fixes ushahidi/platform#1509

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1513)

### DIFF
--- a/application/classes/Ushahidi/Repository/Message.php
+++ b/application/classes/Ushahidi/Repository/Message.php
@@ -198,7 +198,8 @@ class Ushahidi_Repository_Message extends Ushahidi_Repository implements
 			->order_by(
 				'uid',
 				'desc'
-			);
+			)
+			->limit(1);
 		$result =	$query->execute($this->db);
 
 		$last_uid = $result->get('uid', 0) ? $result->get('uid', 0) : null;
--- a/application/classes/Ushahidi/Repository/Message.php
+++ b/application/classes/Ushahidi/Repository/Message.php
@@ -189,6 +189,23 @@ class Ushahidi_Repository_Message extends Ushahidi_Repository implements
 		return false;
 	}
 
+	public function getLastUID($data_provider_type)
+	{
+		$last_uid = null;
+		$query = DB::select([DB::expr('ABS(' . $this->getTable() . '.' . 'data_provider_message_id' . ')'), 'uid'])
+			->from($this->getTable())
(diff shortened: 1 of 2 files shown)


application/classes/Ushahidi/Repository/Message.php
```diff
@@ -189,6 +189,23 @@ public function checkStatus($status, $direction)
 		return false;
 	}
 
+	public function getLastUID($data_provider_type)
+	{
+		$last_uid = null;
+		$query = DB::select([DB::expr('ABS(' . $this->getTable() . '.' . 'data_provider_message_id' . ')'), 'uid'])
+			->from($this->getTable())
+			->where('data_provider', '=', $data_provider_type)
+			->order_by(
```
rjmackay: Can you add a limit clause to this? unless I'm missing something we're loading *all* messages just to get the lastest one?

plugins/data-provider-email/classes/DataProvider/Email.php
```diff
@@ -149,32 +151,32 @@ public function fetch($limit = FALSE)
...
 					elseif (! empty($message))
 					{
 						$message = imap_qprint($message);
-						$this->_process_incoming($overview[0], $message);
+						$this->_process_incoming($email, $message);
 					}
 
 					// After processing, delete!
-					imap_delete($connection, $email_number);
+					//imap_delete($connection, $email_number);
```
rjmackay: Did we never succeed in deleting message before? it seems like this should have prevented fetching message multiple times anyway.

Are we marking pulled message read or anything now?

application/classes/Ushahidi/Repository/Message.php
willdoran: Done

plugins/data-provider-email/classes/DataProvider/Email.php
willdoran: So this is dependent on imap setup, for gmail it doesn't work since they block delete you have to do a convoluted move operation to trash. For the others it just doesn't seem to work.

No we're not changing anything about the messages, we can mark them read but I kind of wanted to leave it as much as possible.

Issue #1412, opened 2016-09-25
Change category visibility to match pattern library


jshorland:
From same user having issues with images. @rjmackay, can you help ID the issue when you're back online today?:

Another problem keeps arising as well. While you've restored the pictures, our categories seem to keep going missing! I have taken a snip of the drop down filters bar and this is what happens,

The category not only dissapears here in the filters bar but also in the settings bar, where the categories were initially set. Besides, before this, when I select a certain categories used to appear in the task bar, below the deployment name, as shown in the right side of the below image but unfortunately now it does not. This is as shown in the picture below. 

Why does this happen? Can you help us fix it?
Now that the pictures issues have been partly resolved this last one day,at least this map is presentable. However I need it in its full working condition for the state goverment to view it and if they are in favour and support, perhaps subscribe to one of your paid versions of Ushahidi. Please treat this as an urgent request as well.

jshorland:
https://app.intercom.io/a/apps/hl5rfiga/inbox/all/conversations/6060959927

rjmackay:
This is because 1 category was only visible to logged in users

This is how it was set
<img width="907" alt="screenshot 2016-09-26 15 46 53" src="https://cloud.githubusercontent.com/assets/7965/18821263/23e0e396-8401-11e6-8b58-014400672134.png">

This is what it needed to be to be seen by everyone.
<img width="860" alt="screenshot 2016-09-26 15 46 59" src="https://cloud.githubusercontent.com/assets/7965/18821264/23e1e638-8401-11e6-9f10-7d0fe6474399.png">

I think we should update this UI and probably use the same style of role selector as seen elsewhere:
https://www.dropbox.com/s/dbnbt9ba8ofhdrq/Screenshot%202016-09-26%2015.52.56.png?dl=0

rjmackay:
Lower priority since this is fixed for the reporting user

rjmackay:
Correct UI for this is present on http://preview.ushahidi.com/platform-pattern-library/develop/assets/html/5_layouts/settings-category-edit.html

Angamanga:
I might mistake what the behaviour of the roles for the categories should do but I think I found an error. The roles selected in categories does not seem to do anything with the categories. They are visible to all users independent of what roles are set. A member can choose between all categories when creating a post and if a category that is only available for admins is set, it is still visible in posts for not logged in. Is this really correct behaviour?

jshorland:
[...]

PR #1523, opened 2016-12-02
1412 role ui

Angamanga:
- Adding the column 'everyone_can_create' (boolean) to the tags-table
- Adding a migration that adds the column with the default-value `false` to existing tags

Test these changes by:
- Not any visible change yet (is in pr for changing the ui)
- Make sure each field in the table 'tags' has the everyone_can_create value of `false`

Fixes ushahidi/platform#1412 (together with pr in the client) .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1523)

### DIFF
--- a/migrations/20161202144619_add_tag_everyone_can_create.php
+++ b/migrations/20161202144619_add_tag_everyone_can_create.php
@@ -1,0 +1,41 @@
+<?php
+
+use Phinx\Migration\AbstractMigration;
+
+class AddTagEveryoneCanCreate extends AbstractMigration
+{
+    /**
+     * Change Method.
+     *
+     * More information on this method is available here:
+     * http://docs.phinx.org/en/latest/migrations.html#the-change-method
+     *
+     * Uncomment this method if you would like to use it.
+     *
+    public function change()
+    {
+    }
+    */
+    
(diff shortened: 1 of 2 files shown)


migrations/20161202144619_add_tag_everyone_can_create.php
```diff
@@ -0,0 +1,41 @@
...
+    }
+    */
+    
+    /**
+     * Migrate Up.
+     */
+    public function up()
+    {
+        $this->table('tags')
+            ->addColumn('everyone_can_create', 'boolean', [
```
willdoran: sorry, my fault, this should probably be 'everyone_can_see'

Angamanga: Hmm, everyone_can_create comes from the survey-form, where that is used. I could change it to 'everyone' in the directive instead, to make it more general?

willdoran: It would be better to name it in terms of its purpose of in this case everyone can see the tag. Let's discuss and see if the field is actually need for tags.

Issue #1486, opened 2016-11-06
Allow multiple selection of roles when creating a saved search


aoduor:
### Expected behaviour
I'd like to be able to limit the audience of saved searches to more than one user role

### Actual behaviour
ATM, i can only select a single role in a dropdown

### Steps to reproduce the behaviour/error
* Set search criteria
* Click on "Save search"
* Add saved search details
* Select audience of your saved search

rjmackay:
@jshorland I don't think this is P0.. nor a bug.

rjmackay:
- This should apply to both saved searches and collections
- The collection view in the pattern library shows how this should look:
http://preview.ushahidi.com/platform-pattern-library/develop/assets/html/5_layouts/map-collection.html
(click on edit in the sidebar)
- This roles control exists in a few places already, so hopefully it can be reused from there

https://github.com/ushahidi/platform/issues/1412 needs almost exactly the same change.

Angamanga:
I have been looking through the code and starting to find my way. There is no separate directives for this in the survey-editor but the code for the functionality is there, so I started writing a directive. That in turn led me to questions about where to put files, and how everything is put together so a client-code-Q&A-skype-session would be good. Is it you @willdoran, @rjmackay or someone else who can guide me through the code?

willdoran:
@Angamanga We can do a complete code walk through tomorrow together during the one on one

PR #1526, opened 2016-12-09
Exchanging column 'visible_to' to 'role' in for sets-table

Angamanga:
- renames the column `visible_to` to `role` in `sets`-table
- exchanges `visible_to` to `role` where used

Test these changes by:
- [ ] make sure that the column `visible_to` is renamed to `role` in sets-table
- [ ] make sure that `visible_to` is not used within the platform-code
 
Fixes ushahidi/platform#1486 .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1526)

### DIFF
--- a/src/Core/Entity/Set.php
+++ b/src/Core/Entity/Set.php
@@ -22,7 +22,7 @@ class Set extends StaticEntity
 	protected $url;
 	protected $view;
 	protected $view_options;
-	protected $visible_to;
+	protected $role;
 	protected $featured;
 	protected $created;
 	protected $updated;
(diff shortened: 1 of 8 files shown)


src/Core/Entity/Set.php
```diff
@@ -38,7 +38,7 @@ protected function getDefinition()
 			'url'          => '*url',
 			'view'         => 'string',
 			'view_options' => '*json',
-			'visible_to'   => '*json',
+			'role'   => '*json',
```
willdoran: I think the spacing needs to be fixed here to matching surrounding lines

Issue #1503, opened 2016-11-14
After changing location of a twitter post, the post appears in two locations, not the updated one


jshorland:
### Expected behaviour
If I change the location of a tweet when structuring it as a post, it should update to that location. NOt the original location of the tweet
### Actual behaviour
The post appears with two locations

### Steps to reproduce the behaviour/error
1. login to usaelectionmonitor.ushahidi.io 
2. grab an unstructured tweet
3. try to change the location
4. click save
![screen shot 2016-11-14 at 2 26 46 pm](https://cloud.githubusercontent.com/assets/3224856/20285478/1edbd954-aa90-11e6-9801-4034ae9c4657.png)

rjmackay:
Probably the same as  #1508

rjmackay:
@Angamanga I think you were already looking at this?

Angamanga:
@jshorland Corresponding pr with list for testing: https://github.com/ushahidi/platform-client/pull/461

jshorland:
Again. Sorry this has taken me so long :-/ fail on my part. Passes QA. Let's get this deployed!

Issue #1508, opened 2016-11-16
Two map pins after changing location


natmanning:
### Expected behaviour

When I change the location of a report that has location on it, there is only 1 location (the one i changed it it)

### Actual behaviour

If a post comes in with location (Twitter w/ location turned on for instance), and then I change the location to the location of the content, not where the person was when they submitted, it ends up showing two pins on the Map, both locations.

### Steps to reproduce the behaviour/error

![screen shot 2016-11-16 at 12 27 26 pm](https://cloud.githubusercontent.com/assets/2200987/20364467/1d90ee04-abf8-11e6-876a-6be45911d687.png)

willdoran:
@natmanning are the two pins both there after a page refresh or indeed trying the same view from a different browser? I'm trying to work out if we don't clear the map properly.

jshorland:
@natmanning @willdoran both pins are there after refresh. I tried in a different browser (firefox vs. chrome) and it's there for me as well. @natmanning could you double check these?

natmanning:
yes still there. It seems to be the case when someone submits a post from Twitter w/ their Geolocation turned on, but the post itself is about a different place. After the post structurer/admin moves the point to a new place, it now shows up as both places. 

If you load usaelectionmonitor.ushahdi.io as a viewer (dojn't login) and look at the posts in Sweden, and then open up that post page. You will see it is posted as boht places: 

https://usaelectionmonitor.ushahidi.io/posts/437

rjmackay:
@Angamanga I think you were already looking at this?

Angamanga:
@jshorland Same as #1503, fixed by pr: https://github.com/ushahidi/platform-client/pull/461

Angamanga:
@jshorland You can test this on this post: http://qa.ushahididev.com/posts/5027

jshorland:
Fixed by #1503

PR #1530, opened 2016-12-14
Filter out message_location-values from twitter-posts

Angamanga:
- Filtering out the values for `message_location` when a post/message has another location assigned to it.

Test these changes by:

Fixes ushahidi/platform#1503, #1508  .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1530)

### DIFF
--- a/application/classes/Ushahidi/Repository/Post.php
+++ b/application/classes/Ushahidi/Repository/Post.php
@@ -138,10 +138,19 @@ class Ushahidi_Repository_Post extends Ushahidi_Repository implements
 		// Get all the values for the post. These are the EAV values.
 		$values = $this->post_value_factory
 			->proxy($this->include_value_types)
-			->getAllForPost($id, $this->include_attributes);
-
+				->getAllForPost($id, $this->include_attributes);
+		
+		$messageLocation = 'message_location';
+        $found_location = false;
 		$output = [];
+
 		foreach ($values as $value) {
+			/*  check to see if there are other locations set than message_location */
+			if($this->checkForLocation($value, $messageLocation)) 
+			{
+				$found_location = true;
+			}
+
 			if (empty($output[$value->key])) {
 				$output[$value->key] = [];
 			}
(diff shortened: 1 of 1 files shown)


application/classes/Ushahidi/Repository/Post.php
```diff
@@ -138,20 +138,42 @@ protected function getPostValues($id)
 		// Get all the values for the post. These are the EAV values.
 		$values = $this->post_value_factory
 			->proxy($this->include_value_types)
-			->getAllForPost($id, $this->include_attributes);
-
+				->getAllForPost($id, $this->include_attributes);
+		
+		$messageLocation = 'message_location';
+        $found_location = false;
```
willdoran: I think there is an indentation problem here

application/classes/Ushahidi/Repository/Post.php
```diff
@@ -138,20 +138,42 @@ protected function getPostValues($id)
...
 		$output = [];
+
 		foreach ($values as $value) {
+			/*  check to see if there are other locations set than message_location */
+			if($this->checkForLocation($value, $messageLocation)) 
+			{
```
willdoran: I think this looks reasonable but it would be good to get a second opinion from @rjmackay and @jasonmule

application/classes/Ushahidi/Repository/Post.php
Angamanga: Yep, definitely...

rjmackay:
Had a quick chat with @willdoran about this.. I would take an entirely different approach. I don't think hiding a piece of data make senses, once something isn't being returned in the API it might as well not exist. We keep the twitter location in the messages table already (although in JSON). When we copy the message location into the location field in the client, we should also delete the twitter location.

Will can help you make the tweaks to the client to ensure that happens.

Angamanga:
Makes sense, I found where to tweak the client so added the change this pr: https://github.com/ushahidi/platform-client/pull/461.

Issue #1585, opened 2017-02-16
Unencrypted IMAP doesn't work


tuxpiper:
### Expected behaviour
When configuring an IMAP server with no encryption as datasource, platform should be able to retrieve emails from it

### Actual behaviour
Platform doesn't attempt to connect to the IMAP server because of a malformed connection string. In the logs can be seen:

```
ERROR: imap_open(): Couldn't open stream {imap.mail.server.com:143/IMAP/None}INBOX. Errors: Can't open mailbox {lin05.hkdns.co.za:143/IMAP/None}INBOX: invalid remote specification in /var/www/html/platform/application/classes/Ushahidi/Console/Dataprovider.php:80
```

The remote specification is indeed malformed (notice the "None" after "IMAP/").

PR #1586, opened 2017-02-16
fixes check for unencrypted IMAP #1585

tuxpiper:
The check was against 'none' , but the setting comes as 'None'. Just made the check case insensitive.

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1586)


tuxpiper:
Fixes #1585

Issue #1582, opened 2017-02-09
No tweets being pulled in despite correct data source integration


aoduor:
### Expected behaviour
* We should be seeing posts from tweets under #resist and #equality on https://equality.ushahidi.io under the Unknown posts saved search

### Actual behaviour
* We're getting no tweets corresponding to hashtags set under twitter configuration

### Steps to reproduce the behaviour/error
* Log into your deployment
* Set up twitter config
* Set up hashtags #equality #resist 
* Check the Unknown posts saved search

cc @natmanning 

Worth noting that I tested with my own twitter credentials just to be sure that Nat's weren't the problem.

aoduor:
Got another one of our users complaining about the same issue on intercom.

See below. 

"I have gone through all of the support materials, and I just need a little help with troubleshooting if you have a moment.

As far as I can tell, I have set up the twitter and email data sources correctly.

If the twitter feed is supposed to forward hashtagged tweets into the installation as unknown posts, it is not happening. I did test posts with the appropriate hashtags and nothing came through. I also sent emails to the exclusive email account set up for receiving posts, and nothing came through.

I imagine that my assumptions about possible use might be wrong, or I might have made an error in settings setup. Can you please offer some simple examples on how these two should work when used correctly? There is great information on setup, but I could use some guidance on putting them into practice and understanding the flow of data.

Thanks so much for your help!

Best,

Frank"

willdoran:
There appears to be a bug with hashtags, it works correctly when the # is dropped. I'll take this now.

Angamanga:
I dont manage to reproduce it, for me, both # and no-# works locally, but only with a `,` between the search-terms. Where did you reproduce it @willdoran ? I am going through the requests we do anyway and check that we are doing correct queries to twitter...

willdoran:
@Angamanga Just to clarify here, I was able to reproduce on staging and prod, but did it succeed for you? I'll try again to see if I can still reproduce.

Angamanga:
@willdoran no, i could not reproduce it, tried on staging (qa.ushahididev.com) and locally with different variations of search-terms and they comes through.

jshorland:
On QA.ushahididev.com, I could not reproduce this. cc: @willdoran

Angamanga:
[...]

PR #1594, opened 2017-03-03
Resetting twitter since_id if search-terms are changed

Angamanga:
- resetting since_id if the twitter-search-terms are changed

Test these changes by:
- add a twitter-search-term and pull in tweets
- change the search term. Tweets older/same age than the ones pulled in from the previous search-term should be present

Fixes ushahidi/platform#1582 .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1594)

### DIFF
--- a/application/classes/Ushahidi/Repository/Config.php
+++ b/application/classes/Ushahidi/Repository/Config.php
@@ -52,14 +52,15 @@ class Ushahidi_Repository_Config implements
 		$immutable = $entity->getImmutable();
 		foreach ($entity->getChanged() as $key => $val) {	
 			if (! in_array($key, $immutable)) {				
-				if($key === 'twitter')
+				
+				/* Below is to reset the twitter-since_id when the search-terms are updated. This should be revised when the data-source tech-debt is addressed*/
+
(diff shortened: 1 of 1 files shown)


application/classes/Ushahidi/Repository/Config.php
```diff
@@ -48,10 +48,19 @@ public function update(Entity $entity)
...
 
-		foreach ($entity->getChanged() as $key => $val) {
-			if (! in_array($key, $immutable)) {
+		$immutable = $entity->getImmutable();
+		foreach ($entity->getChanged() as $key => $val) {	
+			if (! in_array($key, $immutable)) {				
+				
+				/* Below is to reset the twitter-since_id when the search-terms are updated. This should be revised when the data-source tech-debt is addressed*/
+
+				if($key === 'twitter' && $val['twitter_search_terms'] !== $config['twitter']['twitter_search_terms'])
```
willdoran: I think for now this is an OK place to put it, but we should think about adding a better abstraction - however - if datasources are in need of a rewrite, it's worth noting it for then

Issue #1550, opened 2017-01-20
Review Yodie for potential COMRADES Integration


Shadrock:
**Task:** We are planning a "test integration" to ensure that tech developed by COMRADES consortium partners can be integrated with Ushahidi's work flow. To that end, please review [YODIE](https://gate.ac.uk/applications/yodie.html) as an example. Identify any potential issues that need to be addressed such as differences in programming language, UI/UX needs, general compatibility with platform. 

The test integration will also require revisiting the API documentation Robbie recently created: [Draft in Ushahidi Drive](https://docs.google.com/document/d/1Nxd5s443MyttCqWGMhKIlZVfHz1i04gFbyKqOwjvTlc/edit) and [submitted final doc in COMRADES Drive](https://drive.google.com/file/d/0B0fInrbWAUMsNl94Wm5qSzZ4RzA/view?usp=sharing). Most of the consortium were uncertain of how to use this documentation with regards to the products they are developing. 

**Due Date:** February 17th

**Note:** There is working meeting for COMRADES scheduled for March 29-30 in the Netherlands. If there's anything specific that should be done in person (e.g. training them on how to use Github) or if you'd like to attend and present something that will make your life easier, this would be a good time.

rjmackay:
@Shadrock is the test integration due by Feb 17th? or just review for scoping?

Shadrock:
Robbie: that date is for scoping. To test the integration is going to be a bit of a logistics lift... we'll need to get the devs from the COMRADES consortium to enter it via Github to start with... so they'll be a call, etc. But I want to make it as easy for them as possible so that they'll actually do it the way we want to.

Shadrock:
@rjmackay @jshorland I just love the great whooshing sound a deadline makes as it goes by... I'm setting up a call with Jess this week to talk integration: it would be nice to have this done. Robbie: would it be easier for you to just start by sitting in on that call? If so, I'll arrange a time.

rjmackay:
Yikes. Sorry!

### Initial thoughts:
- Ideally we wouldn't run or host YODIE ourselves, but use it through an API. Rather the partner that built it should run it.
- The existing YODIE api appears to be very simplistic: http://services.gate.ac.uk/yodie/api.html
  - **There is no spec for the expected response format. We'll need this to start anything**
  - It looks like the API returns links to any terms it could match in its database. We could probably store these as a list appended to a post, and maybe linked within the post body. Other tools might have more complex data which could be trickier for us to store in a deployment
[...]

PR #1637, opened 2017-03-13
1550 comrades yodie integration

willdoran:
- [x] Adding webhook integration
- [x] Adding Markdown field type
- [x] Add Markdown validation

Test these changes by:
-

Fixes ushahidi/platform#1550

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1637)


src/Core/Entity/Webhook.php
```diff
@@ -0,0 +1,60 @@
...
+
+namespace Ushahidi\Core\Entity;
+
+use Ushahidi\Core\StaticEntity;
+
+class Webhook extends StaticEntity
+{
+	protected $id;
+	protected $user_id;
+	protected $post_id;
```
rjmackay: Why does a webhook have a post id? Shouldn't a webhook just be a URL, secrets and the user id that registered it (maybe)?

src/Core/Entity/WebhookQueue.php
```diff
@@ -0,0 +1,58 @@
...
+ * @package    Ushahidi\Platform
+ * @copyright  2014 Ushahidi
+ * @license    https://www.gnu.org/licenses/agpl-3.0.html GNU Affero General Public License Version 3 (AGPL3)
+ */
+
+namespace Ushahidi\Core\Entity;
+
+use Ushahidi\Core\StaticEntity;
+
+class WebhookQueue extends StaticEntity
```
rjmackay: Could we call the entity a 'job' rather than queue? That seems more consistent with usual terminology

src/Core/Tool/Authorizer/WebhookAuthorizer.php
```diff
@@ -0,0 +1,71 @@
...
+		// These checks are run within the user context.
+		$user = $this->getUser();
+
+		// Only logged in users have access if the deployment is private
+		if (!$this->hasAccess()) {
+			return false;
+		}
+
+		// Admin is allowed access to everything
+		if ($this->isUserAdmin($user)) {
```
rjmackay: I think this endpoint could probably be admin only? ie. no owner or other access

src/Core/Tool/Signer.php
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * @author     Ushahidi Team <team@ushahidi.com>
+ * @package    Ushahidi\Platform
+ * @copyright  2014 Ushahidi
+ * @license    https://www.gnu.org/licenses/agpl-3.0.html GNU Affero General Public License Version 3 (AGPL3)
+ */
+
+namespace Ushahidi\Core\Tool;
+
+class Signer
```
rjmackay: Indenting is all over the place in this file

migrations/20170311004159_create_webhook_queue.php
```diff
@@ -0,0 +1,32 @@
...
+{
+
+    /**
+     * Migrate Up.
+     */
+    public function up()
+    {
+      $this->table('webhook_queue')
+        ->addColumn('user_id', 'integer', ['null' => false])
+		    ->addColumn('post_id', 'integer', ['null' => false])
```
rjmackay: indenting weirdness.

application/classes/Ushahidi/Repository/Webhook/Queue.php
```diff
@@ -0,0 +1,75 @@
...
+	{
+		$state = [
+			'created' => time(),
+		];
+
+		return parent::create($entity->setState($state));
+	}
+
+	// WebhookQueueRepository
+	public function getWebhooks($limit)
```
rjmackay: This naming is weird. Ideally I think we're call them Jobs and this would be getJobs

application/classes/Ushahidi/Console/Webhook.php
```diff
@@ -0,0 +1,135 @@
...
+		// Get Queued webhook requests
+		$webhook_requests = $this->webhookQueueRepository->getWebhooks($limit);
+
+		// Start transaction
+		$this->db->begin();
+
+		foreach ($webhook_requests as $webhook_request) {
+			// Generate request
+			$request = $this->generateRequest($webhook_request);
+			$this->client->sendAsync($equest);
```
rjmackay: Could we just move the send inside `generateRequest` ie. change it to processRequest() or something

application/classes/Ushahidi/Listener/PostListener.php
```diff
@@ -0,0 +1,51 @@
...
+use Ushahidi\Core\Entity\WebhookQueueRepository;
+use Ushahidi\Core\Entity\WebhookRepository;
+
+class Ushahidi_Listener_PostListener extends AbstractListener
+{
+	protected $repo;
+
+	protected $webhook_repo;
+
+	public function setRepo(WebhookQueueRepository $repo)
```
rjmackay: These could probably just be done through constructor injection.. matter of opinion though

application/classes/Ushahidi/Repository/Post.php
```diff
@@ -867,6 +875,11 @@ public function create(Entity $entity)
 			$this->updatePostStages($id, $entity->form_id, $entity->completed_stages);
 		}
 
+		// TODO: Revist post-Kohana
+		// This might be better placed in the usecase but
+		// given Kohana's future I've put it here
+		$this->emit($this->event, $id, 'create');
```
rjmackay: I kinda think we could embed these directly in the CRUDS usecases so they're universal.
But yea, we're likely to need to refactor anyway

application/classes/Ushahidi/Core.php
```diff
@@ -670,6 +689,19 @@ public static function init()
 		$di->setter['Ushahidi_Listener_PostSetListener']['setRepo'] =
 			$di->lazyGet('repository.notification.queue');
 
+		// Event listener for the Post repo
+		$di->setter['Ushahidi_Repository_Post']['setEvent'] = 'PostCreateEvent';
+		$di->setter['Ushahidi_Repository_Post']['setListener'] =
```
rjmackay: The weirdest thing about this event/listener set up is that there can only be 1 listener.
Probably something we can address when integrating laravel, but there should really be 1 central event bus with unlimited listeners.

src/Core/Entity/Webhook.php
willdoran: doesn't, missed removing it from original version I made.
Done.

src/Core/Entity/WebhookQueue.php
willdoran: Done

application/classes/Ushahidi/Console/Webhook.php
willdoran: done

Issue #1651, opened 2017-03-21
Allow visitors of a deployment to choose language to view from homepage


aoduor:
### Expected behaviour
When visiting a deployment made available in another language, It would help to be able to access it in English as well.

Something like what we had on v2 deployments here
![screenshot](https://drops.azureedge.net/drops/files/acc_531939/AXpH?rscd=inline%3B%20filename%3DScreen%2520Capture%2520on%25202017-03-21%2520at%252011-37-39.gif&rsct=image%2Fgif&se=2017-03-21T09%3A08%3A17Z&sig=EAy5UrELV%2FK8f0xarlw%2BYePZqacQN%2Bq1cF1RZPyRTFk%3D&sp=r&sr=b&st=2017-03-21T08%3A08%3A17Z&sv=2013-08-15)

Context: The SBTF is informally helping the deployer for https://peru.ushahidi.io get volunteers to help with managing data around the floods in Peru. 

Comment from Per Aarvik: "Can this instance be multilingual? English/Spanish? If so - it would dramatically expand the number of volunteers that could contribute..."

### Actual behaviour
There's no option for me to switch languages on the home page

cc @brandonrosage @jshorland

aoduor:
Flagging coz this does sound like an urgent request.

rjmackay:
Should this be a per user setting? or just changable for the current session?

Angamanga:
@brandonrosage Do you have any suggestions where to put it this kind of dropdown?

aoduor:
This should be per session I would think?

Handheld Sent,

Angela Oduor Lungati,
Director of Community Engagement,
Ushahidi | www.ushahidi.com

jshorland:
@Angamanga I'm going to move this back to design column and assign @brandonrosage

brandonrosage:
My recommendation is to include this control in the mode context. After the mode bar (menu), It contains the most global content. The toolbar, by contrast, is intended to display controls for manipulating the content immediately in view -- not the whole deployment.

So we get this:

![medium screen](https://cloud.githubusercontent.com/assets/1136279/24297150/a97d0412-106f-11e7-9a81-af8ebb2aa0f3.png)

And on small screens: 

![small screen](https://cloud.githubusercontent.com/assets/1136279/24297163/b9093450-106f-11e7-9aeb-e2b93d7c0f12.png)

[...]

PR #1659, opened 2017-03-26
Adds language to user

Angamanga:
- Adding column 'language' to user-table. This is to make the translation adjust the language based on the user-settings.

Test these changes by:
- When requesting a user, language should be set or null.

Fixes ushahidi/platform#1651 .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1659)


migrations/20170325110408_add_language_to_user.php
```diff
@@ -0,0 +1,41 @@
...
+{
+    /**
+     * Change Method.
+     *
+     * More information on this method is available here:
+     * http://docs.phinx.org/en/latest/migrations.html#the-change-method
+     *
+     * Uncomment this method if you would like to use it.
+     *
+    public function change()
```
willdoran: Just remove this

Issue #1372, opened 2016-09-08
Migrate from kohana to laravel / lumen


rjmackay:
This will probably be a multi step process

1. Install laravel, and migrate routing to laravel. All validation, db queries still use Kohana. Use both auto loaders side-by-side. The big lift here is just migrating the Rest controller.
2. Start migrating away from any helpers etc we're using. Try to avoid auto loading anything from kohana. DB and validation code could still use kohana.. but we should try to rely on *just* those libraries and nothing else
3. Migrate validators
4. Finally migrate DB code.

rjmackay:
@jshorland This got slightly more urgent now Kohana is taking everything offline as of July http://discourse.kohanaframework.org/t/kohana-retirement-2017-07-01/1277

rjmackay:
**Progress**

- [x] Add lumen in to codebase #1663
- [ ] Migrate API controllers to Lumen #1674 
- [ ] Move Repos, Validators, etc out of Kohana
- [ ] Move DI set up out of Kohana (into a service provider)
- [ ] Remove Kohana dependencies from everything other than validators and repos
- [ ] Stop bootstrapping kohana
- [ ] Fix Dataprovider callbacks
- [ ] Migrate event handling/listeners to lumen
- [ ] Migrate console to artisan
- [ ] Make `artisan migrate` call phinx
- [ ] Migrate validators to use lumen
  - Add test coverage
  - Use Illuminate\Validation
- [ ] Migrate repos
  - Do we refactor too?
  - Or just move to Illuminate\Database

rjmackay:
Looking at the kohana code. I think it might be possible to extract just the Kohana Validation libs into their own library while we transition, and removing the Kohana dependencies from the database lib should be fairly simple too.

rjmackay:
@willdoran here's where this is at:

- I'm working on splitting Kohana's Validation lib into a standalone (https://github.com/ushahidi/kohana-validation)...
- Next things are
   - Move dataproviders into namespaces ie. src/App/DataProviders. I suggesting *just* moving the code and namespacing it. *Don't* refactor, we will do that, but first we want to drop kohana. The fastest way to do that is to get the dataprovider code out of modules.
   - Remove kohana dependencies from ohanzee/database & namespace it. You can see Woody started namespacing it over here: https://github.com/ushahidi/ohanzee-db/compare/master...namespaces
[...]

Issue #712, opened 2015-09-18
Replace kohana db with a standalone DB library


rjmackay:
Needs to support:
- Mysql
- Postgis
- Mssql (nice to have)
- SQLite (for testing) + spatialite?

Need a query builder, and dbal.. but migrations stay with phinx, and no ORM.

Aha! Link: https://ushahiditeam.aha.io/features/PROD-727

rowasc:
Pending move to Lumen DB https://github.com/ushahidi/platform/pull/1663

Issue #710, opened 2015-09-18
Replace Kohana Validation with better validation lib


rjmackay:
- Potentially Laravel Validation lib, or some other standalone validator.
- Implement existing validation interfaces
- Generate errors that can be tied to form fields in the client
- Need to be able to handle validating nested arrays better, particularly reporting errors for things like post values.

Aha! Link: https://ushahiditeam.aha.io/features/PROD-728

rowasc:
Note:  The heavier work of getting validators out of kohana is done in the v4-lumen version since Robbie added the ability to use multiple validator engines, so if I'm not mistaken, we need to start using Laravel validation instead in new work? tagging tech-debt to review

PR #1663, opened 2017-03-30
Migrate to Lumen

rjmackay:
- Adds lumen boilerplate app
- Moves tests out of application/
- Updates a bunch of packages
- Updates some tests
- Make httpdocs/index.php just point to lumen

**Progress**
- [x] Add lumen in to codebase (This PR)
- [x] Migrate API controllers to Lumen #1674 
- [x] Move Repos, Validators, etc out of Kohana #1741 
- [x] Move DI set up out of Kohana (into a service provider?) #1741 
- [x] Migrate dependencies on kohana config to laravel config or config repo #1746
- [x] Stop bootstrapping kohana #1841 
- [x] Fix Dataprovider callbacks #2009 
- [x] Migrate console to artisan #2202 
- [x] Remove Kohana dependencies from everything other than validators and repos #1841 
- [x] Clean up DI layer
- [x] Make `artisan migrate` call phinx
- [x] Improve dataproviders
- [x] Migrate file handling to use laravel's filesystem

**Post merge follow ups**
- Migrate event handling/listeners to lumen (started) #2891 
- Set up queues and move over scheduled tasks #2890
- Move to lumen db #712 
- Move to lumen validation #710 

Test these changes by:
- See if CI build passes

Depends on #1666 and ushahidi/platform-cloud-interface#136 (updating to php 5.6)
Refs #1372 
Refs ushahidi/platform-cloud-interface#271

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1663)


app/Exceptions/ValidationException.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+namespace Ushahidi\App\Exceptions;
+
+use Symfony\Component\HttpKernel\Exception\UnprocessableEntityHttpException;
+
+/**
+ * Validation Exception
+ */
+class ValidationException extends UnprocessableEntityHttpException
```
jasonmule: I suppose that we are doing this to prevent error responses from breaking? Are we testing this using API tests?

app/Http/Middleware/ExampleMiddleware.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Ushahidi\App\Http\Middleware;
+
+use Closure;
+
+class ExampleMiddleware
```
jasonmule: :+1: For the example here and the one below as well...

app/Exceptions/ValidationException.php
rjmackay: I needed a way to catch the errors from the Validators and return them in the HTTP response - we did something similar in Kohana. Hopefully we can come up with a cleaner way later on, this was a bit ugly.

src/Core/Usecase/Form/VerifyFormLoaded.php
```diff
@@ -58,26 +58,34 @@ protected function verifyFormExists()
 	 * Checks that the form exists.
 	 * @param  Data $input
 	 * @return void
+	 * @todo  maybe move to validator
 	 */
 	protected function verifyFormDoesNoExistInTargetedSurveyState()
 	{
 		// Ensure that the form exists.
 		if ($this->form_contact_repo->formExistsInPostStateRepo($this->getRequiredIdentifier('form_id'))) {
-			throw new \HTTP_Exception_400('The form already has a set of contacts');
```
rjmackay: @rowasc FYI we should pretty much never be using HTTP errors directly in src/Core code.
The idea is that outer layers (controllers, commands etc) depend on inner layers (Usecases) but not the reverse.
So Controller -> Usecases -> Entities etc

Issue #1654, opened 2017-03-23
Build failure "undefined function Symfony\Component\Config\ctype_alpha()"


chrisveness:
I am attempting to build Ushahidi platform as per [Install Ushahidi on Linux](https://www.ushahidi.com/support/install-ushahidi#installing-on-linux) instructions.

After the following steps:

    $ git clone https://github.com/ushahidi/platform.git
    $ cd platform
    $ [setup .env]
    $ mv httpdocs/template.htaccess httpdocs/.htaccess
    $ chmod 0777 application/logs application/cache application/media/uploads
    $ bin/update

I get an error reported:

    ...
    >>> Running migrations
    PHP Warning:  Module 'dom' already loaded in Unknown on line 0
    PHP Warning:  Module 'gd' already loaded in Unknown on line 0
    PHP Warning:  Module 'imap' already loaded in Unknown on line 0
    PHP Warning:  Module 'mcrypt' already loaded in Unknown on line 0
    PHP Warning:  Module 'SimpleXML' already loaded in Unknown on line 0
    Phinx by Rob Morgan - https://phinx.org. version 0.4.3
    
    PHP Fatal error:  Uncaught Error: Call to undefined function Symfony\Component\Config\ctype_alpha() in /home/[...]/platform/vendor/symfony/config/Symfony/Component/Config/FileLocator.php:85
    Stack trace:
    #0 /home/[...]/platform/vendor/symfony/config/Symfony/Component/Config/FileLocator.php(42): Symfony\Component\Config\FileLocator->isAbsolutePath('application/phi...')
    #1 /home/[...]/platform/vendor/robmorgan/phinx/src/Phinx/Console/Command/AbstractCommand.php(187): Symfony\Component\Config\FileLocator->locate('application/phi...', '/home/movable-t...', true)
    #2 /home/[...]/platform/vendor/robmorgan/phinx/src/Phinx/Console/Command/AbstractCommand.php(212): Phinx\Console\Command\AbstractCommand->locateConfigFile(Object(Symfony\Component\Console\Input\ArgvInput))
    #3 /home/[...]/platform/vendor/robmorgan/phinx/ in /home/[...]/platform/vendor/symfony/config/Symfony/Component/Config/FileLocator.php on line 85

tuxpiper:
Thank you for your report @chrisveness . We are looking into this.

tuxpiper:
@chrisveness looking at your command sequence,  did you run `composer install`?

[...]

Issue #1340, opened 2016-08-14
Check outdated composer packages.


fegul:
When running `bin/update` the following messages appear:

`Package guzzle/guzzle is abandoned, you should avoid using it. Use guzzlehttp/guzzle instead.`

`Package league/url is abandoned, you should avoid using it. Use league/uri instead.`

rjmackay:
We need to do a check of all out of date composer packages.
Some are easy updates, but others will require refactoring.

rjmackay:
Outstanding packages are league/url and guzzle/guzzle, most other packages have been updated.

Aspah:
Thank you! But, how do these packages affect the Ushahidi v.3x installation?

rjmackay:
Installation works just fine with the current set of packages

rjmackay:
We still have guzzle/guzzle because the rackspace flysystem driver requires it. But we no longer depend on it ourselves

PR #1666, opened 2017-03-31
Update composer dependencies

rjmackay:
- Move tests out of application/ into tests/
- Update composer packages to recent versions
  - Update PHPUnit
  - Update Behat
  - Update PHPSpec
  - Update aura/di to remove custom fork
  - Switch flysystem to 1.0 not 1.1 dev
  - Update phinx
- Update phpunit tests for new version
- Update behat tests for new version
- Disable auto resolve in aura/di
- Fix DI setup to work without auto resolve
- Fix linting
- Fix bin/ushahidi to work with newer symfony/console

Future work
- Update guzzle
- Update league/uri (or replace with laravel package)
- Update phpcs

Test these changes by:
- Run tests

Refs #1372 #1340 #1654 

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1666)


application/classes/Ushahidi/Core.php
```diff
@@ -149,7 +149,6 @@ public static function init()
 		$di->setter['Ushahidi\Console\Application']['injectCommands'][] = $di->lazyNew('Ushahidi_Console_Webhook');
 		$di->setter['Ushahidi_Console_Webhook']['setDatabase'] = $di->lazyGet('kohana.db');
 		$di->setter['Ushahidi_Console_Webhook']['setPostRepo'] = $di->lazyGet('repository.post');
-		$di->setter['Ushahidi_Console_Webhook']['setSigner'] = $di->lazyGet('tool.signer');
```
jasonmule: Why can't this be injected here?

jasonmule:
Updates and tests seem to work. Why do we disable auto resolve in aura/di?

rjmackay:
> Updates and tests work ok. Why do we disable auto resolve in aura/di?

Aura/di docs recommend you turn it off because its hard to use/debug. As it happens I turned it off and found some of our DI set up was working by chance, because we had typos in dependencies etc

application/classes/Ushahidi/Core.php
rjmackay: Its unneeded, but also the signer needs an auth token param thats only set during executing of the webhook command

rjmackay:
This is ready to land once ushahidi/platform-cloud-interface#136 is done

rjmackay:
Ah.. So coverage drops with this because we're no longer include `vendor/` in the coverage. It shouldn't have been included in the first place so...

willdoran:
@rjmackay @tuxpiper was saying that staging is already at 5.6 and prod should move over this week. I'd like to merge this branch as I need the changes to be able to add intercom, as such I think it should be ok for develop to go to 5.6 while master is waiting. Does that work for you? 
Should we lower the code coverage requirement since vendor caused a drop?

rjmackay:
I think its OK to merge. Code coverage will probably fluctuate with lumen integration

PR #1670, opened 2017-04-03
Improve categories

Angamanga:
- Adds a forms_tags table to add the relation between forms and tags
- Returns tags and forms when requesting tags and forms from endpoints
- Updates the forms_tags table when adding/changing tags connected to a form/forms connected to a tag
- Adds tags as a field_attribute
- Updates the post_tags-table when a new post with the field_attribute of type tags is created

Test these changes by:
- 

Fixes ushahidi/platform#1397 .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1670)


application/classes/Ushahidi/FormsTagsTrait.php
```diff
@@ -0,0 +1,120 @@
...
+    }
+    
+    // updating/adding tags to a form
+    private function updateFormsTags($form_id, $tags) {
+        if(!$tags) {
+            DB::delete('forms_tags')
+                ->where('form_id', '=', $form_id)
+                ->execute($this->db);
+        } else {
+            if($tags){
```
willdoran: I'm not sure this if is needed, or it could be rolled into the else at least

migrations/20170417090621_move_post_tag_values_to_post_varchar.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    /**
+     * Change Method.
+     *
+     * More information on this method is available here:
+     * http://docs.phinx.org/en/latest/migrations.html#the-change-method
+     *
+     * Uncomment this method if you would like to use it.
+     *
+    public function change()
```
willdoran: Can remove this

PR #1674, opened 2017-04-06
Migrate API controllers to lumen

### LINKED ISSUE #1372: Migrate from kohana to laravel / lumen

rjmackay:
- Define Config Controller
- Define parent REST controller
- [x] Connect up laravel passport for oauth
  - [x] Create ushahidiui client
  - [x] Define scopes
  - [x] Set up test data
  - [x] Populate user in aura DI (or something)
- [x] Set CORS headers and respond to OPTIONS requests
- [x] Format error responses same as before conversion
- [x] Convert API controllers
- [ ] Wire up data provider controllers to call back to Kohana (for now)
- [x] Remove old Kohana controllers

Test these changes by:
- Run tests

Fixes #1372 

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1674)


rjmackay:
@willdoran given this is kinda huge heres the minimum that I think you need to review:
- Double check my changes to the tests
- Review the code in app/ outside of Http Controllers
  - Particularly check the wiring around Passport/auth
- Review 1 or 2 controllers. Maybe look at the special cases for Post and Forms
- Review the base RESTController

willdoran:
@rjmackay It all looks like it's coming along, I don't see anything wrong, the tests look good generally. I like the directory structure generally, though I wonder about whether we could adopt the approach of moving the tests right beside the code. I found that structure very effective in the rollcall js client. Though it may not be strictly kosher in the lumen structure. 

My main concern looking at the Repos is that our data model needs to be restructured specifically for forms, stages, attributes and posts. If nothing else we should take the chance to rename them to match the client. That is probably true for Tags as well. 

May be in Ireland we can updatet the data model. I think involving Dale in that would be good. By then we'll have basically 3/3.5 clients(mobile, facebot, comrades, client) and that should give us a good idea of common needs. Mobile particularly seems like it should force us down the road of lighter requests.

rjmackay:
> My main concern looking at the Repos is that our data model needs to be restructured specifically for forms, stages, attributes and posts. If nothing else we should take the chance to rename them to match the client. That is probably true for Tags as well.

Agreed. I'm trying to just get rid of Kohana first, without rewriting the world in the process. So I'll get through the bulk of this, and the data provider refactor before I tackle the data model shift...

We'll have to come up with a plan for managing old clients if we shift the data model.

app/Http/Controllers/RESTController.php
```diff
@@ -0,0 +1,290 @@
...
+                if ($type === 'jsonp') {
+                    $response->withCallback($request->input('callback'));
+                    // Prevent Opera and Chrome from executing the response as anything
+                    // other than JSONP, see T455.
+                    $response->headers->set('X-Content-Type-Options', 'nosniff');
+                }
+            }
+
+            // Should we prevent this request from being cached?
+            if (! in_array($request->method(), $this->cacheableMethods)) {
```
rjmackay: This is straight from the Kohana version, but in short no. The web server knows nothing about the content its serving, so it doesn't know if its safe to cache or not.
I think originally this was used to make sure GET responses were cachable, but POST/PUT/DELETE were not. I suspect its not actually working now

tests/integration/oauth.feature
```diff
@@ -45,7 +45,7 @@ Feature: Testing OAuth2 endpoints
     Scenario: Authorized Posts Request
         Given that I want to update a "Post"
         And that its "id" is "95"
-        And that the request "Authorization" header is "Bearer testingtoken"
+        And that the oauth token is "testingtoken"
```
rjmackay: Mostly due to having to generate JWT tokens in RestContext grr.

app/Http/Controllers/RESTController.php
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    /**
+     * Get an API URL for a resource.
+     * @param  string  $resource
+     * @param  mixed   $id
+     * @return string
+     * @todo  move this somewhere sensible
+     */
+    public static function url($resource, $id = null)
```
rjmackay: Yup. See the `@todo  move this somewhere sensible` :)

app/Http/Controllers/API/Posts/PostsController.php
```diff
@@ -0,0 +1,161 @@
...
+	protected function getFilters(Request $request)
+	{
+        $params = $this->getRouteParams($request);
+		return $request->query() + [
+			'type'      => $this->postType,
+			'parent'    => isset($params['parent_id']) ? $params['parent_id'] : null,
+		];
+	}
+
+	protected function getPayload(Request $request)
```
rjmackay: No. Filters are built from the query string, payload is built from the JSON body of a request.

Issue #1595, opened 2017-03-03
Configure six webhooks for Intercom to monitor


brandonrosage:
In order to properly target new deployers for an Intercom-powered on-boarding process (and to send the right message at the right time), we need to be able to monitor a few things in Intercom regarding any given deployment/user:

* User's deployment has a description (true or false)
* User's deployment has a logo (true or false)
* Number of active data sources on user's deployment
* Number of users on user's deployment
* The name of No. 1 survey on user's deployment (we want to see if they've renamed the default survey)
* The date the user's deployment was created

Regarding a deployment's "date created," we really just need some indication that the user's deployment is new, in order to begin a campaign when they first land on their deployment. The assumption with "date created" is that we could compare it to the campaign's start date in order to send the campaign first-thing to all deployments created after that point in time.

rjmackay:
@willdoran you should probably spec this along with the other intercom events issue

willdoran:
This should be done at the same time as [132](https://github.com/ushahidi/platform-client/issues/132)

willdoran:
As opposed to [132](https://github.com/ushahidi/platform-client/issues/132), these should be configured on the API. The following needs to be implemented:

- [ ] Add Intercom Listener
- [ ] Load Intercom Config
- [ ] Add hook at Config update for 
  - [ ] Deployment Logo - send boolean to Intercom
  - [ ] Deployment Description - send boolean to Intercom
- [ ] Add hook at datasource update
  - [ ] Send Number of datasources to Intercom
- [ ] Add survey update/delete hook
  - [ ] If first survey changed - specifically name - send name to Intercom
- [ ] This would ideally come from the cloud creation, but it could equally be a hook on config when first log in is completed. However, the latter might be too late. This requires a little more thought.

jshorland:
@willdoran should i move to dev backlog?

rjmackay:
@will where's this at?
Dale suggested using https://segment.com/ to funnel metrics to various tools.. which could actually work well if we *haven't* already done the intercom work. It'll cost a little but may give us more flexibility for data warehousing later. If we've already started on sending intercom events we can ignore Segment for now.

jshorland:
@will this after mobile banners

willdoran:
@jshorland PR with checklist is here for QA #1718

jshorland:
Passes QA

rjmackay:
Waiting for #1734 to be finished before this is enabled

rjmackay:
[...]

PR #1718, opened 2017-04-25
1595 intercom

willdoran:
- Adding intercom events to platform

Test these changes by:
-

Fixes ushahidi/platform#1595

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1718)


application/tests/features/api.users.feature
```diff
@@ -0,0 +1,192 @@
+@usersFixture @oauth2Skip @users
```
willdoran: I need to move this file to fit the new layout

willdoran:
Notes: 
1. It sometimes take a little while for intercom to update, not sure how long but definitely less than 5 minutes.
2. First time login seems to be disabled, not sure if this is now handled separately.

- [ ] Change Deployment description
  - [ ] You should see HAS_DESCRIPTION var for user in test intercom
- [ ] Change/Add deployment logo
  - [ ] You should see HAS_LOGO var for user in test intercom
- [ ] Delete user
  - [ ] You should see TOTAL_USERS var for user in test intercom
- [ ] Register user
  - [ ] You should see TOTAL_USERS var for user in test intercom
- [ ] Change primary form - form with id 1 - name
 - [ ] You should see primary_survey_name with the name you have given your survey
- [ ] Add/Remove datasource
  - [ ] You should see NUM_DATA_SOURCES var for user in test intercom
- [ ] First time login
 - [ ] Once first time login is complete you should see DEPLOYMENT_CREATED_DATE

jshorland:
Looks good in intercom! passes QA.

application/classes/Ushahidi/Core.php
```diff
@@ -73,6 +73,11 @@ public static function init()
 				and Kohana::$config->load('features.private.enabled');
 		});
 
+		// Intercom config settings
+		$di->set('site.intercomAppToken', function() use ($di) {
+			return Kohana::$config->load('site.intercomAppToken');
```
rjmackay: Grouping this into `site` kind of implies that it can change per deployment... it doesn't.
Site is also user editable... this shouldn't be right?

application/classes/Ushahidi/Core.php
```diff
@@ -672,6 +677,22 @@ public static function init()
 		$di->setter['Ushahidi_Listener_PostListener']['setWebhookRepo'] =
 			$di->lazyGet('repository.webhook');
 
+		// Add Intercom Listener to Config
```
rjmackay: It might be too late for this.. but could we move these into src/App ? (fine if not, I'll move them in the lumen branch instead)

application/classes/Ushahidi/Repository/Config.php
```diff
@@ -49,21 +55,49 @@ public function update(Entity $entity)
...
+				$intercom_data['has_description'] = true;
+			}
+
+			if ($key === 'image_header') {
+				$intercom_data['has_logo'] = true;
+			}
+
+			// New User - set their deployment created date
+			if ($key === 'first_login') {
+				$intercom_data['deployment_created_date'] = date("Y-m-d H:i:s");
```
rjmackay: This seems likely to be error prone. Could we do something like inject a real 'deployment_created_date' via migrations? ... or we push deployment creation date from the `platform-cloud-interface` codebase instead since it actually knows when deployments are created.

application/classes/Ushahidi/Repository/Config.php
```diff
@@ -49,21 +55,49 @@ public function update(Entity $entity)
 
 		$config = \Kohana::$config->load($group);
 
+		// Intercom count datasources
+		if ($group === 'data-provider') {
+			$intercomData['num_data_sources'] = 0;
```
rjmackay: This seems very closely coupled for something that is supposedly loosely coupled...

application/classes/Ushahidi/Repository/Form.php
```diff
@@ -77,6 +83,15 @@ public function create(Entity $entity)
     // UpdateRepository
     public function update(Entity $entity)
     {
+
+        // If orignal Form update Intercom if Name changed
+        if ($entity->id === 1) {
```
rjmackay: Whats with the magic number?

application/classes/Ushahidi/Repository/User.php
```diff
@@ -133,6 +138,9 @@ public function isUniqueEmail($email)
 	// RegisterRepository
 	public function register(Entity $entity)
 	{
+
+		$this->updateIntercomUserCount(1);
```
rjmackay: I think it'd make more sense to push the user info itself into Intercom when a new user is added, particularly admins. Then we can actually message all admins from a deployment via intercom.

application/classes/Ushahidi/Core.php
willdoran: hmmmm, I hadn't thought of that. I added it to site to allow users to edit it for their deployment not with the intention that they can push updates to it. Would it be better under something like third_party_app_config?

application/classes/Ushahidi/Repository/User.php
willdoran: I would definitely agree with that. The issue asked for user count though but it would be vastly easier to push the user itself and then intercom filter by deployment to get the count.

application/classes/Ushahidi/Repository/Form.php
willdoran: The original form id, not sure how else to identify it.

application/classes/Ushahidi/Core.php
rjmackay: Yea. Its not super clear where it should live. Probably just `thirdparty` or something similar would work. I think you could just call `getenv()` directly here and skip the kohana config entirely

rjmackay:
Hrm. I somehow missed that this had already merged :(

application/classes/Ushahidi/Repository/Config.php
willdoran: In retrospect, I really dislike this the most. I think what we could do is insert the original user into intercom when the deployment is created in cloud interface then update the user when they complete first_login separately. I don't think that is hugely difficult. The app token can be push into cloud interface from the same ansible.

application/classes/Ushahidi/Core.php
willdoran: If i'm going to make the other changes I can move this as well.

willdoran:
It can be improved still :)

willdoran:
So for the user event I would agree that pushing the entire admin object might be better. For the form, my understanding of the required data was just whether or not the Primary Form had changed name. Part of the issue with the required data is that it is individual piece rather than whole objects. In retrospect, it would have been better to take the required data and abstract it out to some kind of object that can be passed.

There is an issue with pushing each user, which one is the primary? If I push 5 admins but I only push an update that the primary form name has been changed, it's only going to go intercom user that changed it so all the other admins won't have that data and it will be confusing if it's done by a non-admin so there will be a new user created. 

Intercom has Company objects though and it seems better to use that and then users can be associated with companies. Do you think that'd work?

application/classes/Ushahidi/Repository/Config.php
willdoran: Yeah, I'm not sure where else we could intercept the datasources. The ideal would really be to just have an occasional query that gives us stats on deployments. That would work best in the context of the single DB though. Then if that's needed in intercom we could push from our own stats collector.

rjmackay:
> If I push 5 admins but I only push an update that the primary form name has been changed, it's only going to go intercom user that changed it so all the other admins won't have that data and it will be confusing if it's done by a non-admin so there will be a new user created.

That should be a property of the company not the user.

PR #1731, opened 2017-05-08
adding new migration to show_when_published

Angamanga:
- sets show_when_published to true

Test these changes by:
-

Fixes ushahidi/platform# .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1731)

### DIFF
--- a/migrations/20170507060511_make_show_when_published_to_true.php
+++ b/migrations/20170507060511_make_show_when_published_to_true.php
@@ -1,0 +1,32 @@
+<?php
+
+use Phinx\Migration\AbstractMigration;
+
+class MakeShowWhenPublishedToTrue extends AbstractMigration
+{
+    public function up()
+    {
+        $pdo = $this->getAdapter()->getConnection();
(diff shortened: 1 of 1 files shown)


migrations/20170507060511_make_show_when_published_to_true.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+use Phinx\Migration\AbstractMigration;
+
+class MakeShowWhenPublishedToTrue extends AbstractMigration
+{
+    public function up()
+    {
+        $pdo = $this->getAdapter()->getConnection();
+        $rows = $this->fetchAll("SELECT show_when_published from form_stages");
```
tuxpiper: @Angamanga , just a note, since staging ran the original version of the 20170331210813 migration, I think that the table definitions in there will still remain with the default value for the column as 'false' .

It will not be a problem if the value is always explicitly set when making future inserts ... but for things to be 100% in sync, it may make sense to alter the table definition and setting the default of `show_when_published` to true in this migration as well.

PR #1734, opened 2017-05-09
Updating intercom functionality

willdoran:
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1734)


application/classes/Ushahidi/Core.php
```diff
@@ -74,8 +74,8 @@ public static function init()
 		});
 
 		// Intercom config settings
-		$di->set('site.intercomAppToken', function() use ($di) {
-			return Kohana::$config->load('site.intercomAppToken');
+		$di->set('thirdparty.intercomAppToken', function() use ($di) {
+			return Kohana::$config->load('thirdparty.intercomAppToken');
```
rjmackay: I'd suggest just calling getenv() here rather than using Kohana config

application/classes/Ushahidi/Listener/IntercomAdminListener.php
```diff
@@ -0,0 +1,74 @@
...
+
+use GuzzleHttp\Exception\ClientException;
+
+class Ushahidi_Listener_IntercomAdminListener extends AbstractListener
+{
+
+  public function handle(EventInterface $event, $user = null)
+  {
+
+    $config = service('repository.config')->get('thirdparty');
```
rjmackay: I don't think you need to use the repo here, just `service('thirdparty.intercomapptoken')`
The repo is only really necessary for things that are use editable

application/classes/Ushahidi/Listener/IntercomAdminListener.php
```diff
@@ -0,0 +1,74 @@
...
+    $domain = Kohana::$config->load('site.client_url');
```
rjmackay: You might be able to use `service('site')` here. I think it'll give you `domain.ushahidi.io` or similar. But relies on HOST being set or an actual HTTP request

application/classes/Ushahidi/Listener/IntercomCompanyListener.php
```diff
@@ -0,0 +1,65 @@
...
+
+    Kohana::$log->add(Log::ERROR, print_r($config));
+
+		if ($config->intercomAppToken) {
+
+      $client = new IntercomClient($config->intercomAppToken, null);
+
+			try {
+        // Create company with current date if it does not already exist
+        if (!$config->intercomCompanyId) {
```
rjmackay: When would this happen? ie. why wouldn't we have created the company already?

application/classes/Ushahidi/Repository/User.php
```diff
@@ -64,7 +64,11 @@ public function create(Entity $entity)
 			'created'  => time(),
 			'password' => $this->hasher->hash($entity->password),
 		];
-		return parent::create($entity->setState($state));
+		$entity->setState($state);
+		if ($entity->role === 'admin') {
```
rjmackay: Ideally this check happens in the listener too. Then the event is reusable for future integrations that might want *all* users.

application/classes/Ushahidi/Listener/IntercomAdminListener.php
willdoran: What will that do in a non .io setup, it seems to be unset if not multisite

application/classes/Ushahidi/Listener/IntercomAdminListener.php
```diff
@@ -0,0 +1,55 @@
...
+      $domain = service('site');
+      $company = [
+        "id" => $domain
+      ];
+
+  		if ($intercomAppToken) {
+        $client = new IntercomClient($intercomAppToken, null);
+
+  			try {
+          $client->users->update([
```
rjmackay: indenting is weird here?

application/classes/Ushahidi/Listener/IntercomAdminListener.php
```diff
@@ -0,0 +1,55 @@
...
+use GuzzleHttp\Exception\ClientException;
+
+class Ushahidi_Listener_IntercomAdminListener extends AbstractListener
+{
+
+  public function handle(EventInterface $event, $user = null)
+  {
+    if($user && $user->role === 'admin') {
+      $intercomAppToken = getenv('INTERCOM_APP_TOKEN');
+      $domain = service('site');
```
rjmackay: This should return `quakemap.api.ushahidi.io` or just an empty string if we're *not* on an ushahidi.io deployment ... was that what you expected?

application/classes/Ushahidi/Listener/IntercomCompanyListener.php
```diff
@@ -18,21 +18,31 @@
...
-class Ushahidi_Listener_IntercomListener extends AbstractListener
+class Ushahidi_Listener_IntercomCompanyListener extends AbstractListener
 {
-  public function handle(EventInterface $event, $user_email = null, $data = null)
+  protected $config_repo;
+
+  public function setConfigRepo(ConfigRepository $config_repo)
   {
-		$intercomAppToken = service('site.intercomAppToken');
+  		$this->config_repo = $config_repo;
```
rjmackay: this is never used?

application/config/thirdparty.php
```diff
@@ -0,0 +1,30 @@
+  <?php defined('SYSPATH') OR die('No direct access allowed.');
```
rjmackay: Not sure we need this file at all anymore?

application/classes/Ushahidi/Repository/User.php
rjmackay: Any change on this?

PR #1738, opened 2017-05-12
Adjusting unmapped request to incorporate it into stats

willdoran:
- Changed how unmapped is calculated
- Added unmapped to stats response
- Added filter by mapped/unmapped

Test these changes by:
-

Fixes ushahidi/platform# .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1738)

### DIFF
--- a/application/classes/Ushahidi/Repository/Post.php
+++ b/application/classes/Ushahidi/Repository/Post.php
@@ -228,6 +228,7 @@ class Ushahidi_Repository_Post extends Ushahidi_Repository implements
 			'center_point', 'within_km',
 			'published_to',
 			'include_types', 'include_attributes', // Specify values to include
+			'include_unmapped',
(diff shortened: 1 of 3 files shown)


application/classes/Ushahidi/Repository/Post.php
```diff
@@ -689,9 +701,13 @@ public function getGroupedTotals(SearchData $search)
 
 		// Fetch the results and...
 		$results = $this->search_query->execute($this->db);
+		$results = $results->as_array();
+
+		// Append unmapped totals to
+		$results['unmapped'] = $this->getUnmappedTotal($this->getSearchTotal());
```
rjmackay: This is kind of weird adding this to all the stats responses (ie. regardless of what we're grouping by... but I don't have a better solution particularly.

Does this mess with the activity view at all? (hopefully our code was robust and it doesn't)

PR #1746, opened 2017-05-19
Refactor config handling

rjmackay:
- Stop using Kohana Config for site, features and data provider config
- Disable loading kohana config from db
- Move test data provider config into the db

Test these changes by:
- bin/behat
- ... something else?

Refs ushahidi/platform#1372
Depends on #1663 

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1746)


rjmackay:
@willdoran I updated the base of this PR so the diff should be sane now.
Its still a work in progress and I'm trying to figure out some details still
- Previously we used kohana config to define some defaults for 'config' values, we need to handle these in the repo or somewhere else
- Part of me thinks this should really transition to a different structure entirely
  - Separate entities for SiteConfig, FeatureConfig, Dataprovider config
  - Each entity could then define the possible values as needed.
  - That said... its still useful to have a flexible key/value store for config. It just starts to break down when dealing validation or config several levels deep
- For everything outside of site/map/data provider/feature config - we can just use lumen config + dotenv since thats all set at deploy time, not by users.

Any ideas on how to manage this properly would be good.

src/App/Repository/ConfigRepository.php
```diff
@@ -104,6 +119,23 @@ public function update(Entity $entity)
...
 
+	private function insertOrUpdate($group, $key, $value)
+	{
+		$value = json_encode($value);
+
+		try {
+			DB::insert('config', array('group_name', 'config_key', 'config_value'))
+				->values(array($group, $key, $value))
+				->execute($this->db);
+		} catch (\Database_Exception $e) {
```
willdoran: Really not sure about this, but I think it would go away once we have entities to manage certain configs.

tuxpiper: From @willdoran comment:

> I think doing away with the generic json storage for dataproviders would be preferable. It would kind of make sense if we had dozens of dataproviders but in reality we don't and they are never going to increase in number at a particularly dramatic rate. Specifically, for data providers I would think we would be best to add a dataprovider archetype and then extend for that into SMS types, email, twitter etc.

I'm somewhat divided about this. While explicit is generally good 👍 , I think we should also bear in mind the custom integration needs that may appear in individual projects we do for clients. 

Creating specific archetypes for those may make it more difficult to merge in updates from main platform into those project repositories?

rjmackay: > Really not sure about this, but I think it would go away once we have entities to manage certain configs.

Its gross but taken directly from the Kohana Database Config driver. I was going for minimal changes, rather than sensible code. Easy to improve though

rjmackay:
- I'm going to leave the dataprovider config alone until we start reworking them entirely.
- I'll combine map + site config, and
- Feature config should get a major change when we combine DBs and it can actually refer directly to the deployments and tiers tables. Till then any refinement seems a bit wasted

rjmackay:
Ugh. Ok actually: I just want to land this and get to fixing data providers. Then I'll return to this... its messy but it works. And no kohana.

PR #1760, opened 2017-05-26
Fixing error message for tag

willdoran:
- Fixes incorrect error message for category import

Test these changes by:
- [ ] When importing a csv with categories that do not exist in the system you should receive an error message saying `The field :param1 must be a valid category id, Category id: :param2`
where :param1 is the name of the category field and :param2 is the incorrect id/tag

Fixes ushahidi/platform# .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1760)

### DIFF
--- a/application/classes/Ushahidi/Repository/Form.php
+++ b/application/classes/Ushahidi/Repository/Form.php
@@ -83,9 +83,12 @@ class Ushahidi_Repository_Form extends Ushahidi_Repository implements
             $key === 'name' ? $this->emit($this->event, $user->email, ['primary_survey_name' => $val]) : null;
           }
         }
-
+        $form = $entity->getChanged();
+        $form['updated'] = time();
+        // removing tags from form before saving
+        unset($form['tags']);
         // Finally save the form
-        $id = parent::update($entity->setState(['updated' => time()]));
+        $id = $this->executeUpdate(['id'=>$entity->id], $form);
 
         return $id;
     }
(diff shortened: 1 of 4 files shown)


application/messages/post.php
```diff
@@ -27,6 +27,7 @@
 		'digit'         => 'The field :param1 must be a digit, Given: :param2',
 		'email'         => 'The field :param1 must be an email address, Given: :param2',
 		'exists'        => 'The field :param1 must be a valid post id, Post id: :param2',
+    'tagExists'     => 'The field :param1 must be a valid category id, Category id: :param2',
```
Angamanga: Minor issue, the category can be identified with category name too...

application/classes/Ushahidi/Repository/Tag.php
```diff
@@ -116,7 +116,8 @@ public function create(Entity $entity)
 	public function update(Entity $entity)
 	{
 		$tag = $entity->getChanged();
-
+		// removing children before saving tag
+		unset($tag['children']);
```
rjmackay: Is this necessary? I thought I made `children` part of the immutable fields for Tag already.

application/classes/Ushahidi/Repository/Form.php
```diff
@@ -83,9 +83,12 @@ public function update(Entity $entity)
             $key === 'name' ? $this->emit($this->event, $user->email, ['primary_survey_name' => $val]) : null;
           }
         }
-
+        $form = $entity->getChanged();
+        $form['updated'] = time();
+        // removing tags from form before saving
+        unset($form['tags']);
```
rjmackay: I already made `tags` part of the immutable fields on the object. I'm not sure why we still need to do this. I couldn't figure out how to reproduce it locally.

application/classes/Ushahidi/Repository/Tag.php
rjmackay: Ah immutable has no effect, because the previous value was 'empty' so it just sets it :/

PR #1766, opened 2017-05-31
Release 3.7.0

willdoran:
**New Features**
- Added Field Level Locking to Surveys
- Added Task Level Visibility
- Added Webhook Support
- Added Webhook events fro Create, Update of Post
- Added Survey Duplication 
- Added Task Duplication
- Added reply to twitter via Post feature
- Added support for hierarchical categories
- Added Category field type to Survey
- Added Markdown field type to Survey
- Add tags dynamically directly from Posts
- Added language switching for all users

**Minor improvements and fixes**
- Add more details on use case internals
- Fixed keyword twitter fetch
- Display number of unmapped posts to timeline
- Format for PSR2
- Replaced shadowhand/email
- Fix waffle badge
- Fixes check for unencrypted IMAP
- Making sure the email with last-uid is not imported again

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform/1766)


rjmackay:
I don't think we should use PRs for this. We should absolutely never squash-merge develop into master. It makes the commit history impossible to follow - we can't see which commits from develop are now in master

PR #1786, opened 2017-06-08
Add ushahidi/kohana-validation package and use that

rjmackay:
This replaces the dependencies on Kohana Core.

- Update Validators to refer to namespaced Valid and Validation class
- Fix DI for ValidationEngineTrait
- Set up translations
- Move error messages to laravel language files

Test checklist:
- [ ] Check travis build runs with only failures in dataproviders

Refs #1372 

Ping @ushahidi/platform


rjmackay:
@willdoran can you review?
Validation package is over here: https://github.com/ushahidi/kohana-validation
I'm not super happy that I had to include the `Arr` helper but it was the fastest path to done.

src/App/Validator/KohanaValidationEngine.php
```diff
@@ -0,0 +1,81 @@
...
+    public function __construct(Translator $translator)
+    {
+        $this->translator = $translator;
+
+        parent::__construct([], $this->getTranslationCallback());
+    }
+
+    public function getTranslationCallback()
+    {
+        return function ($file, $field, $error = null) {
```
rjmackay: This is probably the ugliest bit. We pass a callback into the KohanaValidation lib for translating messages, and do some wrangling here to connect that to lumen's translator.

src/Core/Tool/ValidationEngineTrait.php
```diff
@@ -27,8 +26,8 @@
      * @param  $validation_factory
      * @return void
      */
-    public function setValidation(InstanceFactory $validation_factory)
+    public function setValidation(ValidationEngine $validation_engine)
```
rjmackay: Not sure *why* this was set up so weirdly. Cleaner now

willdoran:
@rjmackay This is to stablise validation during transition?

rjmackay:
> @rjmackay This is to stablise validation during transition?

Yup. We need to drop Kohana ASAP, so this just keeps the couple of classes we need in the interrim.

PR #1802, opened 2017-06-13
Lumen db without kohana

rjmackay:
- Move to namespaced Ohanzee\Database
- Namespace Ushahidi_Multisite
- Review this with https://github.com/ushahidi/ohanzee-db/pull/3

Test checklist:
- [ ] bin/behat

Refs #1372

Ping @ushahidi/platform

### DIFF
--- a/src/App/Multisite.php
+++ b/src/App/Multisite.php
@@ -1,5 +1,4 @@
-<?php defined('SYSPATH') or die('No direct script access');
-
+<?php
 /**
  * Ushahidi Multisite
  *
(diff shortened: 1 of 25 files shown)


src/App/Multisite.php
```diff
@@ -8,15 +7,18 @@
  * @copyright  2014 Ushahidi
  * @license    https://www.gnu.org/licenses/agpl-3.0.html GNU Affero General Public License Version 3 (AGPL3)
  */
+
+namespace Ushahidi\App;
+
 use League\Url\Url;
 
-class Ushahidi_Multisite
+class Multisite
```
willdoran: This is a much better naming scheme. Dir structure should be a clear, consistent way to traverse the source.

rjmackay:
> Is there a way to avoid \ style I find it jarring and inconsistent.

... I'm not clear what you're talking about? Do you mean the `\Ohanzee\Database` type references? The `\` prefix is necessary because we're in a namespace. The alternative is to insist on `use` statements instead... I can definitely do that, it just requirement more than find and replace.

rjmackay:
Updated to remove `\Ohanzee\DB` and `use Ohanzee\DB` instead.

Issue #1635, opened 2017-03-13
API Config Site Returns Non-Encoded Image_Header URL


dalezak:
### Expected behaviour
The `/api/v3/config/site` endpoint is returning non-encoded `image_header`, which is causing some issues in the mobile app.

### Actual behaviour

Some Deployment `image_header` include whitespace in the URL.

### Steps to reproduce the behaviour/error

With a valid `access_token`, hitting https://USAelectionmonitor.api.ushahidi.io/api/v3/config/site returns:

```
"image_header": "https://74ddb495e3f187ded630-0b4c691ea14dad11384ab079da46af4f.ssl.cf2.rackcdn.com/usaelectionmonitor.api.ushahidi.io/5/8/5810e52688ec4-round simple on light.png"
```

Which includes whitespace in the image URL, but probably should instead return

```
https://74ddb495e3f187ded630-0b4c691ea14dad11384ab079da46af4f.ssl.cf2.rackcdn.com/usaelectionmonitor.api.ushahidi.io/5/8/5810e52688ec4-round%20simple%20on%20light.png
```

Perhaps a solution is for the image uploader should sanitize the file URLs before it is saved?

rjmackay:
@crcommons you could look at this as a first issue in the API. This should be a change needed in the Media Formatter class somewhere.  Have a look around, and then I can give you a run through of *how* that various parts fit together

rjmackay:
@dalezak could you test this against a staging deployment? ie. (qa.ushahididev.com)

PR #1810, opened 2017-06-16
#1635 image urls encode correctly.

crcommons:
-correctly encodes image filenames so that there is no whitespace in the url.

Test checklist:
- [ ]

Fixes ushahidi/platform#1635 .

Ping @ushahidi/platform

### DIFF
--- a/application/classes/Ushahidi/Formatter/Media.php
+++ b/application/classes/Ushahidi/Formatter/Media.php
@@ -40,6 +40,7 @@ class Ushahidi_Formatter_Media extends Ushahidi_Formatter_API
 
 	protected function get_field_name($field)
 	{
+
 		$remap = [
 			'o_filename' => 'original_file_url',
 			'o_size'     => 'original_file_size',
@@ -55,11 +56,20 @@ class Ushahidi_Formatter_Media extends Ushahidi_Formatter_API
 	}
 
 	protected function format_o_filename($value)
-	{
+	{		
+
 		if ($cdnBaseUrl = Kohana::$config->load('cdn.baseurl')) {
-			return $cdnBaseUrl . $value;
+
+			//removes path from image file name, encodes the filename, and joins the path and filename together
+			$url_path = explode("/", $value);
+			$filename = rawurlencode(array_pop($url_path));
+			array_push($url_path, $filename);
+			return $cdnBaseUrl . implode("/", $url_path);
+
 		} else {
-			return URL::site(Media::uri($this->get_relative_path() . $value), Request::current());
+
+			//URL::site / Mesdia::uri encodes the image path properly
+			return URL::site((Media::uri($this->get_relative_path() . $value)), Request::current());
 		}
 	}
 
@@ -78,6 +88,7 @@ class Ushahidi_Formatter_Media extends Ushahidi_Formatter_API
 	 */
 	private function resized_url($width, $height, $filename)
 	{
+
 		// Format demensions appropriately depending on the value of the height
 		if ($height != NULL)
 		{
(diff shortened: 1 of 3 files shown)


application/classes/Ushahidi/Formatter/Media.php
```diff
@@ -55,11 +56,16 @@ protected function get_field_name($field)
...
 		if ($cdnBaseUrl = Kohana::$config->load('cdn.baseurl')) {
-			return $cdnBaseUrl . $value;
+			$url_path = explode("/", $value);
+			$filename = rawurlencode(array_pop($url_path));
+			array_push($url_path, $filename);
+
+			return $cdnBaseUrl . implode("/", $url_path);
 		} else {
-			return URL::site(Media::uri($this->get_relative_path() . $value), Request::current());
+			return URL::site((Media::uri($this->get_relative_path() . $value)), Request::current());
```
rjmackay: It'd be worth adding a comment above this to explain that either URL::site or Media::uri is already encoding the url - so that the next person to look at this knows *why* we're only encoding in 1 case, not the other.

tests/integration/bootstrap/FeatureContext.php
```diff
@@ -80,4 +80,18 @@ public function disableDataImport()
 		$config = \Kohana::$config->load('features');
 		$config->set('data-import.enabled', false);
 	}
+
+	/** @BeforeScenario @cdnEnabled */
+	public function enableCdn()
```
rjmackay: You should remove these before this lands, since they didn't work anyway.
Given we don't have a convenient way to enable/disable the CDN during test runs, we should possibly just run with the CDN enabled for all tests since thats the production state. @willdoran what do you think?

application/classes/Ushahidi/Formatter/API.php
```diff
@@ -20,6 +20,7 @@ class Ushahidi_Formatter_API implements Formatter
 	// Formatter
 	public function __invoke($entity)
 	{
+		Kohana::$log->add(Log::ERROR, print_r($entity, true));
```
rjmackay: Can you remove the log statement before this lands?

crcommons: Will do.

tests/integration/bootstrap/FeatureContext.php
crcommons: @willdoran Let me know if I should remove or not. Thanks!

willdoran: @rjmackay @crcommons If it doesn't work then dropping it seems fine. Out of curiosity why doesn't it work?

crcommons: I can't recall why it didn't work; only that it didn't. @rjmackay may be able to remember and explain.

crcommons: I removed this and now the checks have failed. @rjmackay Can we touch base on this?

.travis.yml
```diff
@@ -39,7 +39,7 @@ before_script:
 - mysql -e 'SET @@GLOBAL.wait_timeout=1800'
 script:
 - if [ "${coverage}" == "" ]; then composer test; fi
-- if [ "${coverage}" != "" ]; then composer run coverage --timeout=0; fi
+- if [ "${coverage}" != "" ]; then composer run coverage --timeout=2000; fi
```
rjmackay: So timeout=0 was supposed to remove the timeout entirely. Maybe that stopped working?

tests/integration/bootstrap/FeatureContext.php
rjmackay: I think the only failure was linting.

rjmackay: > If it doesn't work then dropping it seems fine. Out of curiosity why doesn't it work?

@willdoran CDN config doesn't load from the DB, so any changes we make via behat don't stick.

Issue #1814, opened 2017-06-22
Improve form validation when creating and editing a survey


ryanchristo:
### Expected behaviour
- If you attempt to save a new survey without entering a description, you receive an error message that says "description must not be empty".
- If you remove the name or description when editing a survey and attempt to save, you receive an error message that says "name must not be empty" or "description must not be empty".

### Actual behaviour
- If you attempt to save a new survey without entering a description, you receive an error message that says "Field 'description' doesn't have a default value...".
<img width="626" alt="screen shot 2017-06-21 at 9 07 01 pm" src="https://user-images.githubusercontent.com/12519942/27416975-a252a290-56c5-11e7-8eee-4766fb60bcbe.png">

- If you edit a survey and remove the name and/or description and then save the survey, you receive no error message and the survey is saved with an empty string for the name and/or description.

<img width="291" alt="screen shot 2017-06-21 at 9 08 46 pm" src="https://user-images.githubusercontent.com/12519942/27417014-e6aebc12-56c5-11e7-9a70-7a7b3c40dcc1.png">

- If the name is an empty string, you cannot select and edit the survey from the list of surveys.

<img width="705" alt="screen shot 2017-06-21 at 9 09 46 pm" src="https://user-images.githubusercontent.com/12519942/27417026-05e37e10-56c6-11e7-9012-c5e50f2b8ca4.png">

### Steps to reproduce the behaviour/error
- Settings > Surveys > Create
  - leave description field blank and attempt to save the survey
- Settings > Surveys > (select survey)
  - remove the name and description and save the survey
  - attempt to select the survey without a name from the survey list

PR #1818, opened 2017-06-26
Fixes form validation when creating and editing a survey

ryanchristo:
- Fixes form validation when creating and editing a survey by adding 'not_empty' rule to name and always checking validation of name when editing a survey
- Adds migration allowing null for description value, which prevents the 'no default value' error when attempting to save a survey without a description

Test checklist:
- [x] If you attempt to save a new survey without entering a name, you receive an error message that says "name must not be empty".
- [x] If you attempt to edit a survey and remove the name, you receive an error message that says "name must not be empty".
- [x] If you attempt to save a new survey without entering a description, you do not receive an error stating there is 'no default value' for description.

Fixes ushahidi/platform#1814.

Ping @ushahidi/platform

### DIFF
--- a/application/classes/Ushahidi/Validator/Form/Update.php
+++ b/application/classes/Ushahidi/Validator/Form/Update.php
@@ -33,10 +33,12 @@ class Ushahidi_Validator_Form_Update extends Validator
 	{
 		return [
 			'name' => [
+				['not_empty'],
 				['min_length', [':value', 2]],
 				['regex', [':value', Validator::REGEX_STANDARD_TEXT]], // alpha, number, punctuation, space
 			],
 			'description' => [
+				['not_empty'],
 				['is_string'],
 			],
 			'color' => [
--- a/application/classes/Ushahidi/Validator/Form/Update.php
+++ b/application/classes/Ushahidi/Validator/Form/Update.php
@@ -38,9 +38,6 @@ class Ushahidi_Validator_Form_Update extends Validator
 			if ($fullData['name']) {
 				$data['name'] = $fullData['name'];
 			}
-			if ($fullData['description']) {
-				$data['description'] = $fullData['description'];
-			}
 			$this->validation_engine->setData($data);
 			// End
 
@@ -51,7 +48,6 @@ class Ushahidi_Validator_Form_Update extends Validator
 				['regex', [':value', Validator::REGEX_STANDARD_TEXT]], // alpha, number, punctuation, space
 			],
 			'description' => [
-				['not_empty'],
 				['is_string'],
 			],
 			'color' => [
--- a/application/classes/Ushahidi/Validator/Form/Update.php
+++ b/application/classes/Ushahidi/Validator/Form/Update.php
@@ -31,6 +31,19 @@ class Ushahidi_Validator_Form_Update extends Validator
 
 	protected function getRules()
 	{
+
+			// Always check validation for name and description
+			$fullData = $this->validation_engine->getFullData();
+			$data = $this->validation_engine->getData();
+			if ($fullData['name']) {
+				$data['name'] = $fullData['name'];
+			}
+			if ($fullData['description']) {
+				$data['description'] = $fullData['description'];
+			}
+			$this->validation_engine->setData($data);
+			// End
+
 		return [
 			'name' => [
 				['not_empty'],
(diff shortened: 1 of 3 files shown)


application/classes/Ushahidi/Validator/Form/Update.php
```diff
@@ -31,12 +31,27 @@ public function __construct(FormRepository $repo)
...
+			if ($fullData['description']) {
+				$data['description'] = $fullData['description'];
+			}
+			$this->validation_engine->setData($data);
+			// End
+
 		return [
 			'name' => [
+				['not_empty'],
 				['min_length', [':value', 2]],
```
rjmackay: I suspect the `min_length` requirement here would have always picked up errors with `name`

application/classes/Ushahidi/Validator/Form/Update.php
```diff
@@ -31,12 +31,27 @@ public function __construct(FormRepository $repo)
...
 				['regex', [':value', Validator::REGEX_STANDARD_TEXT]], // alpha, number, punctuation, space
 			],
 			'description' => [
+				['not_empty'],
 				['is_string'],
```
rjmackay: What happens if you simply add `['min_length', [':value', 1]],` for description too?

application/classes/Ushahidi/Validator/Form/Update.php
```diff
@@ -31,12 +31,27 @@ public function __construct(FormRepository $repo)
 
 	protected function getRules()
 	{
+
+			// Always check validation for name and description
+			$fullData = $this->validation_engine->getFullData();
```
rjmackay: This whole block shouldn't be required. By design we generally only validate changes to an entity.

ryanchristo:
@rjmackay I will take another look and see if I can find a better solution. I assumed description was intended to be required and can remove validation no problem. A migration would require setting the description to allow for null because a text column cannot have a default value but that should work.

Just to do a quick run through of my thinking as to why I took the route I did with this pull request, I wanted to add the `not_empty` validation to prevent a user from being able to update a survey and remove the name because the `min_length` validation was not throwing an error when the field was left blank. I added the block of code that plugs in the name and description when running validations because `not_empty` was throwing an error when the name and description were unchanged.

I'm sure there is a better way. I'll see what I can do.

application/classes/Ushahidi/Validator/Form/Update.php
ryanchristo: With `not_empty` added to `Form/Update.php`, it was throwing the error when attempting to update the survey without changing the name. Adding this block was the first solution I found after looking at the validation workaround in `Post/Create.php`.

application/classes/Ushahidi/Validator/Form/Update.php
ryanchristo: If the field is left empty, `min_length` does not throw the error, allowing for the name to be saved as an empty string, which is why I moved `not_empty` to `Form/Update.php`.

application/classes/Ushahidi/Validator/Form/Update.php
ryanchristo: Same as above, if the field is left empty, `min_length` does not throw the error, allowing for the description to be saved as an empty string, which is why I added `not_empty`. This can be removed if description should not be required.

rjmackay:
> Just to do a quick run through of my thinking as to why I took the route I did with this pull request, I wanted to add the not_empty validation to prevent a user from being able to update a survey and remove the name because the min_length validation was not throwing an error when the field was left blank. I added the block of code that plugs in the name and description when running validations because not_empty was throwing an error when the name and description were unchanged.

Thanks for explaining. That makes sense, my main concern is that it breaks the pattern used elsewhere: unchanged fields aren't validated on update. Its weird that min_length wasn't throwing an error when title/description was empty though. If they changed from having a value to being empty, then it should have been passing through. Maybe we just didn't have *any* not empty validation on description even during creation?

> A migration would require setting the description to allow for null because a text column cannot have a default value but that should work.

That sounds about right.

ryanchristo:
@rjmackay No luck finding an alternative solution on the server which would prevent the user from editing the survey, deleting the name and saving. If this is not a concern, I can rebase the branch and simplify the pull request with just the added migration. There might be a solution on the client doing something similar to categories, which disables the save button when the name is empty?

Issue #1821, opened 2017-06-26
Filter by incident status


jshorland:
Filter by custom statuses instead of the normal "draft, published, archived".

willdoran:
@Celestine22 What are the custom statuses to be?

rjmackay:
> pending, open, in-progress, finished

is mentioned in the google sheet. I'll assume thats correct unless otherwise specified

rjmackay:
Assuming this is a custom field, we can filter by passing a query param `values[key]=filtervalue`
Then we just need to add some custom UI specific to this client

rjmackay:
Just tried adding this as a custom field + task... but that makes it editable by the user when they create the post. We might need to hack around that :/

rjmackay:
Given the need to make status private, I'll look at making a change to make all tasks enables only by admins by default.

rjmackay:
@Celestine22 there's a little potential for confusion between our current post 'status' (Published, Under Review, Archived) and the client's status as a custom field. Could you confirm if there's any relationship between which posts are visible to the public, and the posts status?
If there isn't a relationship, do they have any other plan around what will/will not be public?

Basically I'm trying to figure out if we need to keep the filters for publish/under review/archived at all. If we do we might have to tweak terminology to make it less confusing.

Celestine22:
@rjmackay got it! I've asked the client about this! Awaiting feedback! My guess would visibility of posts to the public won't be affected by the posts status. Also we may have to remove some of our default filters to avoid that confusion/redundancy. 

@willdoran I've also asked for a confirmation on whether those are the custom statuses.

jshorland:
These will go in filter box. Remove icons from the status. How they are displayed is detailed in #1825

Celestine22:
@rjmackay visiblity of posts to the public will not be affected by the posts status! Custom status given by client today is (New, In Progress, Resolved) but he says he will confirm internally! Hope this helps!

rjmackay:
This is deployed

rjmackay:
- [ ] Check why filter box jumped
- [ ]  Test filtering actually works on demo site

dalezak:
@rjmackay how was this implemented in the backend? Does the API return a list of available statuses which the mobile app can use?

rjmackay:
See my comment here: https://github.com/ushahidi/platform/issues/1824#issuecomment-316852800

rjmackay:
@dalezak Did this get implemented on mobile?

[...]

PR #1849, opened 2017-07-06
Allow multiple value filter

rjmackay:
- Allow searching for multiple values within a post attributes
  Allow queries like `?values[key][]=value1&values[key][]=value2` rather than just `?values[key]=value1`

Test checklist:
- [ ] bin/behat

Refs ushahidi/platform#1821

Ping @ushahidi/platform


rjmackay:
Ugh. I messed this up slighty and rebased off develop first, which makes this fiddlier to re-integrate with the fix-our-environment deployment

Issue #1826, opened 2017-06-26
Editing of posts will be closed to users (you cannot go back and edit your post you submitted) and only available for admins


jshorland:
This should be built as an option for platform. weee! configurable setting to not allow people to edit their own posts after submitting.

rjmackay:
@justinscherer I'm thinking about *where* this setting should live. There are a couple of options:

1. In general settings as a simple toggle "Users can edit their own posts"
2. As a permission under Settings > Roles. We currently have various permissions for the default 'Member' role so we could add 'Edit own posts' as an option (enabled by default.
<img width="710" alt="screenshot 2017-07-10 13 21 01" src="https://user-images.githubusercontent.com/7965/27999572-c104d8f0-6572-11e7-9827-8034e44b60c0.png">

I think 2. is probably the most natural. Any contrary opinions?

PR #1855, opened 2017-07-10
Limit editing own posts 1826

rjmackay:
-  Rename `hasAccess()` to `canAccessDeployment()` in Authorizers
- Refactor PermissionAccess trait to allow checking for multiple permissions
- Add new 'EditOwnPosts' permission
- [x] Add tests for editing own posts w/ EditOwnPosts permission (should succeed)
- [x] Add tests for editing own posts w/o EditOwnPosts permission (should fail)
- [x] Add tests for deleting own posts w/ EditOwnPosts permission (should succeed)
- [x] Add tests for deleting own posts w/o EditOwnPosts permission (should fail)
- [x] Add tests for viewing own posts w/o EditOwnPosts permission (should succeed)
- [x] Add tests for when custom roles is disabled

Test checklist:
- [ ] bin/behat

Fixes ushahidi/platform#1826

Ping @ushahidi/platform


rjmackay:
I've realised one issue with this is that when roles are not enabled (ie. on a mapper plan), we don't run permission checks at all. Either you are an admin and you have all permissions, or a member has none. However with this new permission, members need to have 'EditOwnPosts' permissions.. so I might need a way to define the default permissions when 'Custom roles' is disabled.
Or maybe theres a better approach entirely? @willdoran thoughts?

migrations/20170710013345_add_edit_own_posts_permission.php
```diff
@@ -0,0 +1,24 @@
...
+class AddEditOwnPostsPermission extends AbstractMigration
+{
+    public function up()
+    {
+        $this->execute("INSERT INTO permissions (name, description)
+            VALUES ('Edit their own posts', 'Edit their own posts')
+            ");
+
+        $this->execute("INSERT INTO `roles_permissions` (`role`, `permission`)
+            VALUES ('user', 'Edit their own posts')
```
rjmackay: I probably need to do this for *all* roles in the DB

src/Core/Tool/Authorizer/PostAuthorizer.php
```diff
@@ -140,11 +135,21 @@ public function isAllowed(Entity $entity, $privilege)
             return true;
         }
 
-        // We check if the user is the owner of this post. If so, they are allowed
-        // to do almost anything, **except** change ownership and status of the post, which
-        // only admins can do.
-        if ($this->isUserOwner($entity, $user) && !$entity->hasChanged('user_id')
-            && $privilege !== 'change_status') {
+        // Only admins or users with 'Manage Posts' permission can change status
```
rjmackay: I think pulling these first 2 checks out by themselves is a bit clearer, hopefully

src/Core/Tool/Authorizer/PostAuthorizer.php
```diff
@@ -140,11 +135,21 @@ public function isAllowed(Entity $entity, $privilege)
...
+        // Only admins or users with 'Manage Posts' permission can the ownership of a post
+        if ($entity->hasChanged('user_id')) {
+            return false;
+        }
+
+        // We check if the user is the owner of this post & they have edit own posts permission
+        // they are allowed to edit or delete the post. They can't change the post status or
+        // ownership but those are already checked above
+        if ($this->isUserOwner($entity, $user)
+            && $this->hasPermission($user, Permission::EditOwnPosts)) {
```
rjmackay: So this fails, when custom roles is disabled (I think)... need a different solution *or* I have to refactor hasPermission a little

src/Core/Traits/PermissionAccess.php
```diff
@@ -43,14 +44,14 @@ protected function hasRolesEnabled()
 	 * Check if the user has permission
 	 * @return boolean
 	 */
-	protected function hasPermission($user)
+	protected function hasPermission(User $user, $permission)
 	{
 		// Don't check for permissions if we don't have the
 		// roles feature enabled
 		if (!$this->hasRolesEnabled()) {
```
rjmackay: I'm wondering if instead of this, I should swap out the Acl. ie. a DefaultAcl implementation, and a CustomRolesAcl implementation.

src/Core/Traits/PostValueRestrictions.php
```diff
@@ -15,12 +15,12 @@
 
 use Ushahidi\Core\Entity\User;
 use Ushahidi\Core\Entity\Post;
+use Ushahidi\Core\Entity\Permission;
 use Ushahidi\Core\Entity\FormRepository;
 
 trait PostValueRestrictions
```
rjmackay: I'm not at all sure this, and the classes which use it are set up right. Or the naming is just weird. I might need to pair up to review it.

application/classes/Ushahidi/Repository/Form/Attribute.php
```diff
@@ -26,8 +25,6 @@ class Ushahidi_Repository_Form_Attribute extends Ushahidi_Repository implements
 {
 	use UserContext;
 
-	use PostValueRestrictions;
```
rjmackay: I'm a bit concerned that this was here, but unused.

src/Core/Entity/Permission.php
```diff
@@ -19,6 +19,13 @@ class Permission extends StaticEntity
 	protected $name;
 	protected $description;
 
+	// Standard permissions names
```
rjmackay: This seems much better than a class per permission as previously defined

willdoran: Yeah this is way neater

src/Core/Tool/Authorizer/CSVAuthorizer.php
```diff
@@ -55,7 +50,7 @@ public function isAllowed(Entity $entity, $privilege)
 		$user = $this->getUser();
 
 		// Allow role with the right permissions
-		if ($this->hasPermission($user)) {
+		if ($this->hasPermission($user, Permission::DataImport)) {
```
willdoran: I find this more helpful for code readability

src/Core/Tool/Authorizer/ContactAuthorizer.php
```diff
@@ -42,7 +42,7 @@ public function isAllowed(Entity $entity, $privilege)
 		$user = $this->getUser();
 
 		// Only logged in users have access if the deployment is private
-		if (!$this->hasAccess()) {
+		if (!$this->canAccessDeployment($user)) {
```
willdoran: 👍

src/Core/Tool/Authorizer/PostAuthorizer.php
```diff
@@ -140,11 +135,28 @@ public function isAllowed(Entity $entity, $privilege)
...
-        // to do almost anything, **except** change ownership and status of the post, which
-        // only admins can do.
-        if ($this->isUserOwner($entity, $user) && !$entity->hasChanged('user_id')
-            && $privilege !== 'change_status') {
+        // Only admins or users with 'Manage Posts' permission can change status
+        if ($privilege === 'change_status') {
+            return false;
+        }
+
+        // Only admins or users with 'Manage Posts' permission can the ownership of a post
```
willdoran: missing the word changed should be
// Only admins or users with 'Manage Posts' permission can change the ownership of a post

src/Core/Traits/PermissionAccess.php
willdoran: Yeah that would actually be a nice approach it would be neat to separate roles into its own acl space

src/App/Acl.php
```diff
@@ -13,27 +13,73 @@
...
 	public function hasPermission(User $user, $permission)
 	{
+		// If the user has no role, they have no permissions
 		if (!$user->role) {
 			return false;
 		}
 
+		// Don't check for permissions if we don't have the
+		// roles feature enabled
+		if ($this->hasRolesEnabled()) {
```
rjmackay: So I didn't go as far as splitting the Acl in 2. But I've split the custom roles/default permissions into their method, and moved the handling of enabling/disabling into the Acl itself.

src/Core/Tool/Authorizer/CSVAuthorizer.php
```diff
@@ -55,7 +50,7 @@ public function isAllowed(Entity $entity, $privilege)
 		$user = $this->getUser();
 
 		// Allow role with the right permissions
-		if ($this->hasPermission($user)) {
+		if ($this->acl->hasPermission($user, Permission::DataImport)) {
```
rjmackay: I think this ends up clearer. We can see that the permission check is really happening in the Acl, not this class

src/Core/Tool/Permissions/AclTrait.php
```diff
@@ -1,22 +1,24 @@
...
- * @package    Ushahidi\Platform
+ * @package    Ushahidi\Application
  * @copyright  2014 Ushahidi
  * @license    https://www.gnu.org/licenses/agpl-3.0.html GNU Affero General Public License Version 3 (AGPL3)
  */
 
 namespace Ushahidi\Core\Tool\Permissions;
 
-interface Permissionable
+trait AclTrait
```
rjmackay: I think generally this style of Trait works well ie. a trait purely to handle repetitive DI

src/Core/Traits/PermissionAccess.php
```diff
@@ -1,56 +0,0 @@
...
-	protected function hasRolesEnabled()
-	{
-		return (bool) $this->roles_enabled;
-	}
-
-	/**
-	 * Check if the user has permission
-	 * @return boolean
-	 */
-	protected function hasPermission($user)
```
rjmackay: Rather than this kind of trait which adds magic functions to the class and leaves us unsure which class/trait is really handling permission checks

rjmackay:
@willdoran could you give this a final review?

application/classes/Ushahidi/Repository/Form/Attribute.php
willdoran: No, I think that may have become defunct.

migrations/20170710013345_add_edit_own_posts_permission.php
willdoran: It's nice that the permission is explicit

Issue #1772, opened 2017-06-01
Privacy Notice to Users


caharding:
We are updating our privacy policy to enable new features on our mobile app and RollCall. For every user who logs into Ushahidi after June 30th (subject to change), 

User Stories
- [x] User will be presented with a pop-up dialogue that disables all other features and says:"We have updated our [Terms and Conditions](https://www.ushahidi.com/terms-of-service) and [Privacy Policy](https://www.ushahidi.com/privacy-policy). Please review these documents carefully and accept the terms in order to continue to use Ushahidi's services." Check with Charlie for updates links. 
- [x] User is presented with a checkbox and "accept and continue" button
- [x] The checkbox must be checked to continue
- [x] User will be presented with this dialogue every time they try to login until they accept

Tech spec:
- [x] Extend user object to include Terms & Conditions boolean
- [x] Allow this boolean to be set via our ansible playbooks - for future events where we want to change T&C
- [x] Add Terms & Conditions module to Client with UX - this can extend the first time login system
- [x] Add Terms & Conditions endpoint to retrieve link to relevant terms and conditions
- [x] Ensure OSS Users are not effected by this
- [x] Users that do not agree to the T&C can not continue to their profile

willdoran:
@caharding @jshorland 
Questions:
1. This is per user right? Not per deployment

Tech spec:
1. Extend user object to include Terms & Conditions boolean
2. Allow this boolean to be set via our ansible playbooks - for future events where we want to change T&C
3. Add Terms & Conditions module to Client with UX - this can extend the first time login system
4. Add Terms & Conditions endpoint to retrieve link to relevant terms and conditions

caharding:
Per user. 

Charles Harding

Product Director

207-720-0713

On Fri, Jun 02, 2017 at 11:56 AM will

<
mailto:will <notifications@github.com>
> wrote:

a, pre, code, a:link, body { word-wrap: break-word !important; }

https://github.com/caharding https://github.com/jshorland
Questions:

This is per user right? Not per deployment

Tech spec:

Extend user object to include Terms & Conditions boolean

Allow this boolean to be set via our ansible playbooks - for future events where we want to change T&C

Add Terms & Conditions module to Client with UX - this can extend the first time login system

Add Terms & Conditions endpoint to retrieve link to relevant terms and conditions

—

You are receiving this because you were mentioned.

Reply to this email directly,
https://github.com/ushahidi/platform/issues/1772#issuecomment-305880903
, or

[...]

PR #1932, opened 2017-07-24
Terms of Service API

crcommons:

- [x] Add terms of service migration to add entries when user agrees to terms of service (ToS)
- [x] Add ability to create, read, and search for ToS entries 
- [x] Searching for ToS returns the most recent entry for a given user in the db
- [x] Add tests for creating, reading, and searching for ToS Entries

Test checklist:
- [x] Run Behat Tests

Refs ushahidi/platform#1772

Ping @ushahidi/platform


migrations/20170712152806_add_tos_table.php
```diff
@@ -0,0 +1,30 @@
...
+class AddTosTable extends AbstractMigration
+{
+    /**
+     * Migrate Up.
+     */
+    public function up()
+    {
+        $this->table('tos')
+          ->addColumn('user_id', 'integer', ['null' => false])
+          ->addColumn('agreement_date', 'integer', ['null' => false])
```
rjmackay: I'd usually suggest these should be mysql datetime column, not just integers. That said... if its working as-is we should really get this deployed. This part of the codebase is unlikely to change often so not a major

application/classes/Ushahidi/Repository/Tos.php
```diff
@@ -0,0 +1,76 @@
...
+    protected function getTable()
+    {
+        return 'tos';
+    }
+
+
+    // CreateRepository
+    public function create(Entity $entity)
+    {
+        //get the user ID
```
rjmackay: Minor style issue, our comments are generally of the form `// Get user ID` ie. space after //, and Upper case start, aim for full sentences where it make sense

application/classes/Ushahidi/Validator/Tos/Create.php
```diff
@@ -0,0 +1,52 @@
...
+        return [
+            'id' => [
+                ['numeric'],
+            ],
+            'user_id' => [
+                ['numeric'],
+                [[$this->user_repo, 'exists'], [':value']],
+            ],
+            'agreement_date' => [
+                [[$this, 'validDate'], [':value']],
```
rjmackay: You can just use `['date']` here, similar to numeric above

application/classes/Ushahidi/Formatter/Tos.php
```diff
@@ -0,0 +1,17 @@
...
+ * @package    Ushahidi\Application
+ * @copyright  2014 Ushahidi
+ * @license    https://www.gnu.org/licenses/agpl-3.0.html GNU Affero General Public License Version 3 (AGPL3)
+ */
+
+use Ushahidi\Core\Traits\FormatterAuthorizerMetadata;
+
+class Ushahidi_Formatter_Tos extends Ushahidi_Formatter_API
+{
+    use FormatterAuthorizerMetadata;
```
rjmackay: This should have formatters for the dates - have a look at `Ushahidi_Formatter_API` for what we use for created/updated dates

application/config/init.php
```diff
@@ -28,7 +28,7 @@
 	'base_url'    => '/',
 	'index_file'  => FALSE,
 	'charset'     => 'utf-8',
-	'errors'      => TRUE,
+	'errors'      => FALSE,
```
rjmackay: revert this before it lands?

src/Core/Tool/Authorizer/TosAuthorizer.php
```diff
@@ -0,0 +1,67 @@
...
+            return true;
+        }
+
+        if ($privilege === 'read') {
+            return true;
+        }
+
+
+        // If no other access checks succeed, we default to denying access
+        return true;
```
rjmackay: errr... this should default to `false` not `true` right?

src/Core/Tool/Authorizer/TosAuthorizer.php
```diff
@@ -0,0 +1,67 @@
...
+
+        if ($privilege === 'create') {
+            return true;
+        }
+
+        if ($privilege === 'search') {
+            return true;
+        }
+
+        if ($privilege === 'read') {
```
rjmackay: I think we should only allow reading when user id = current user id right?

src/Core/Usecase/CreateUsecase.php
```diff
@@ -64,19 +64,15 @@ public function interact()
 	{
 		// Fetch a default entity and apply the payload...
 		$entity = $this->getEntity();
```
rjmackay: Can you revert the whitespace changes?

src/Core/Usecase/Tos/CreateTos.php
```diff
@@ -0,0 +1,21 @@
...
+ * @license    https://www.gnu.org/licenses/agpl-3.0.html GNU Affero General Public License Version 3 (AGPL3)
+ */
+
+namespace Ushahidi\Core\Usecase\Tos;
+
+use Ushahidi\Core\Entity;
+use Ushahidi\Core\Entity\Tos;
+use Ushahidi\Core\Usecase\CreateUsecase;
+
+class CreateTos extends CreateUsecase
```
rjmackay: You don't need to extend this if nothing changes, you can just use CreateUsecase as-is

src/Init.php
```diff
@@ -308,6 +310,13 @@ function feature($name)
 	'create'  => $di->lazyNew('Ushahidi\Core\Usecase\Contact\CreateContact')
 ];
 
+// Add custom create usecase for terms of service
+$di->params['Ushahidi\Factory\UsecaseFactory']['map']['tos'] = [
+	'create'  => $di->lazyNew('Ushahidi\Core\Usecase\Tos\CreateTos'),
```
rjmackay: This isn't needed, since you're not changing any behaviour there anyway

tests/integration/tos.feature
```diff
@@ -0,0 +1,27 @@
+@resetFixture @tos 
+Feature: Testing the Tos API
+
+    Scenario: Create a ToS entry
+        Given that I want to make a new "tos"
+        And that the request "Authorization" header is "testbasicuser"
+        And that the request "data" is:
+            """
+            {
+                "tos_version_date":"1500059540"
```
rjmackay: We should accept and return dates in JSON date format.

application/classes/Ushahidi/Repository/Tos.php
crcommons: For some reason the linter was getting mad with spaces (only sometimes) after the // in comments. Will try to fix and see if it works this time.

application/classes/Ushahidi/Validator/Tos/Create.php
crcommons: To discuss. I changed and broke tests.

application/config/init.php
crcommons: done.

src/Core/Tool/Authorizer/TosAuthorizer.php
crcommons: Whoops! Done.

src/Core/Tool/Authorizer/TosAuthorizer.php
crcommons: done.

src/Core/Usecase/Tos/CreateTos.php
crcommons: Done.

src/Init.php
crcommons: Done.

src/Core/Entity/Tos.php
```diff
@@ -0,0 +1,46 @@
...
+            'user_id' => ['user', 'user.id']
+        ];
+    }
+
+    protected function getDefinition()
+    {
+        return [
+            'id'              => 'int',
+            'user_id'         => 'int',
+            'agreement_date'  => 'int',
```
rjmackay: This might need to be `'*date',`

src/Core/Usecase/CreateUsecase.php
crcommons: Done, I think. Ran the command you suggested.

src/Core/Tool/Authorizer/TosAuthorizer.php
```diff
@@ -39,9 +39,14 @@ class TosAuthorizer implements Authorizer
 
     /* Authorizer */
     public function isAllowed(Entity $entity, $privilege)
-    {
+    {   
+        // Get the current user ID
+        $current_user = service('session.user');
```
rjmackay: This is actually the same thing returned by `getUser` below. I think you need to compare with the $entity->user_id

PR #1938, opened 2017-07-28
Adding recursion to stateful data change tracking

willdoran:
- Adding recursive array diff function
- Updating StatefulData::setData to check nested arrays for change
- Updating StatefulData::hasChanged to support multi-level nested key checking

Test checklist:
- [ ] ./bin/behat

Fixes ushahidi/platform# No related issue at present.

Ping @ushahidi/platform

### DIFF
--- a/src/Core/Traits/StatefulData.php
+++ b/src/Core/Traits/StatefulData.php
@@ -17,6 +17,8 @@ trait StatefulData
 	// when comparing, for example, a string with a number (`'5' === 5`).
 	use DataTransformer;
 
+	use RecursiveArrayDiff;
+
 	/**
 	 * Tracks which properties have been changed, separately from internal object
 	 * properties, organized by the unique object id.
--- a/src/Core/Traits/StatefulData.php
+++ b/src/Core/Traits/StatefulData.php
@@ -125,6 +125,7 @@ trait StatefulData
 	 */
 	public function setState(array $data)
 	{
+
 		// Allow for data to be filled in by deriving from other values.
 		foreach ($this->getDerived() as $key => $possible) {
 			if (!array_key_exists($key, $data)) {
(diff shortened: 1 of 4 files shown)


willdoran:
@rjmackay I've pulled the Stateful Data changes from Kuwait into a PR here. I'd like to merge them and then use them as a way to merge the Comrades branch by fixing the tags divergence.

src/Core/Traits/StatefulData.php
```diff
@@ -183,14 +184,42 @@ public function setState(array $data)
 				continue;
 			}
 
-			if ($this->$key !== $value) {
+			if (is_array($value)){
+				$current_key = is_array($this->$key) ? $this->$key : [];
+
+				// Check for multi level recursion
+				$diff = array_merge(
```
rjmackay: I wonder if its worth hunting for a more efficient method of recursive diff'ing before this lands? ie. a diff thats already notes diffs from both sides, rather than running it twice

tests/unit/Core/Traits/MockPostData.php
```diff
@@ -0,0 +1,111 @@
...
+namespace Tests\Unit\Core\Traits;
+
+
+
+class MockPostData
+{
+	use \Ushahidi\Core\Traits\StatefulData;
+    
+    protected $id;
+	protected $parent_id;
```
rjmackay: indenting?

rjmackay:
@willdoran this should be ready to land I think

src/Core/Traits/StatefulData.php
willdoran: I'll do a quick bit of research today and see.

tests/unit/Core/Traits/MockPostData.php
willdoran: Yeah there is something wrong with my editor, I started using visual studio because everyone recommended it but I'm not sure I like it that much, I find the split panes awkwardly sized :(

PR #1982, opened 2017-08-07
cacheability: open access without authorization token to certain endp…

tuxpiper:
…oints

- Opens up a number of endpoint families to not require auth tokens for GET requests

Ping @ushahidi/platform


rjmackay:
We should probably cover the Media api too

tests/integration/acl.editposts.feature
```diff
@@ -67,7 +67,6 @@ Feature: Access Control when modifying posts
     @rolesEnabled
     Scenario: Anonymous users can not edit posts
         Given that I want to update a "Post"
```
rowasc: I think that since there is no "Authorization: Bearer xxxxxxx" header coming in, this test should expect a 401 error, not 403. Unless there's a different reason to expect 403 here ?

rjmackay: Yea thats probably right. This is a bit rushed for Uchaguzi but hopefully we can clean it up next week and bring it back to core. There will definitely be some confusion between 401 and 403 responses

application/classes/Ushahidi/Rest.php
```diff
@@ -266,7 +266,9 @@ protected function _check_access()
 		}
 		catch (OAuth2Exception $e)
 		{
-			if (!$this->_is_auth_required() AND $e instanceof MissingAccessTokenException)
+			if (!$this->_is_auth_required() AND (
+				$e instanceof MissingAccessTokenException OR
+				$e instanceof InvalidAccessTokenException))
```
rjmackay: I'm not sure about this change TBH. Its important we fail invalid tokens, so that the client knows the token isn't valid. Rather than just try to run them as unauth'd.

application/classes/Ushahidi/Rest.php
```diff
@@ -291,7 +293,9 @@ protected function _check_access()
 				}
 			}
 
-			$exception = HTTP_Exception::factory($status, $e->getMessage());
+			$message = "_is_auth_required(): " . $this->_is_auth_required() . ", " . $e->getMessage();
```
rjmackay: Need to revert this before it hits develop

application/classes/Ushahidi/Rest.php
rjmackay: @tuxpiper I've reverted this. But I'm not clear how important that change was for caching?

tuxpiper:
@willdoran Since it was originally put together pretty rushedly, I think it wouldn't hurt to give it some nice focused look. I vote to move back to dev backlog.

rjmackay:
@tuxpiper I'm going to disagree: this should land ASAP because we'll need it for 2chaguzi and it improves speed of other deployments too.

PR #2000, opened 2017-08-09
Uchaguzi csv export address

rjmackay:
- Not sure yet?

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/application/classes/Ushahidi/Repository/Post/Export.php
+++ b/application/classes/Ushahidi/Repository/Post/Export.php
@@ -11,6 +11,9 @@
 use Ushahidi\Core\Entity\Post;
 use Ushahidi\Core\Entity\PostRepository;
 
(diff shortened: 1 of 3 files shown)


rjmackay:
I think this will very quickly hit the rate limit for the geocoding API. We hit it with the geocoder in the client... this seems likely to be worse. If you think its worth trying anyway, all good *but* we should make sure its fault tolerant (ie. CSV still downloads even if it can't geocode.

rowasc:
I'm thinking maybe we can do the geocoding call when a post is created and/or when lat/lon changes, and not only here. When doing exports we could have a fallback to search and save the address if it doesn't exist(*), so that pre-existing posts can have their address exported too.

Issue #1987, opened 2017-08-08
Not all published posts are appearing on the map


dexwakhu:
Publishing Team has been publishing messages all morning but only 128 are appearing as published

jshorland:
From @dexwakhu : "Try it yourself - select the “Unpublished SMS saved search” select a few messages, Mark as published

jshorland:
@willdoran this is fixed now?

willdoran:
It should be

PR #2014, opened 2017-08-15
1987 fix post publishing

willdoran:
- Fixes validation issue where data not passed correctly.
- Ensure post->completed_stages is always an array of int's
- Ensure post fields are always validated (not just when 'post' stage is in completed_stages)

Test checklist:
- [ ]

Refs ushahidi/platform#1987

Ping @ushahidi/platform


application/classes/Ushahidi/Validator/Post/Create.php
```diff
@@ -227,6 +227,8 @@ public function checkTags(Validation $validation, $tags)
 
 	public function checkValues(Validation $validation, $attributes, $fullData)
 	{
+
+		$attributes = !empty($fullData['values']) ? $fullData['values'] : [];
```
rjmackay: I don't understand why this is changed??

willdoran: At one point during Uchaguzi, the $attributes value was not being passed, and validation was failing completely but I think this was actually a different issue. Related to publish from Post Detail and Timeline because the completed_stages in those circumstances was being passed as strings rather than an array of ints.

I'll change this around to reflect that instead.

willdoran: You can test it by creating a survey with a required task, create a post, mark the stage completed, save it, then from timeline or post detail try to publish it.

willdoran: I'm stepping through this now and it seems as though when the value is an array of strings it isn't being passed correctly to the function in the validator.

rjmackay:
Actually... can you just add tests to this that will fail without your changes and pass with them?

application/classes/Ushahidi/Validator/Post/Create.php
```diff
@@ -280,6 +282,7 @@ public function checkValues(Validation $validation, $attributes, $fullData)
 	 */
 	public function checkStageInForm(Validation $validation, $completed_stages, $fullData)
 	{
+		$completed_stages = !empty($fullData['completed_stages']) ? $fullData['completed_stages'] : [];
```
rjmackay: I don't think this is needed. We only need to validate the stages themselves when `completed_stages` change.

application/classes/Ushahidi/Validator/Post/Create.php
```diff
@@ -305,7 +308,7 @@ public function checkStageInForm(Validation $validation, $completed_stages, $ful
 	 */
 	public function checkRequiredStages(Validation $validation, $completed_stages, $fullData)
```
rjmackay: I need to verify if this runs every time, or only when completed_stages changes. The latter would be broken.

application/classes/Ushahidi/Validator/Post/Create.php
rjmackay: I can see why the other changes are needed but I'm unsure why we'd recheck `values` unless it changes.

rjmackay:
@willdoran I've pushed some changes that fix related issues with posts validation.
Can you double check they make sense?

application/classes/Ushahidi/Validator/Post/Create.php
rjmackay: Actually PostUpdate usecase always pushing the full values array through anyway - so this change should make any difference.

willdoran:
I'm not sure I undertstand what you mean by required Post fields? All attributes are children of a task so if you mean the "post task" i.e. the task that is not divided into its own section. I think it would be checked in the same way as the other tasks, or am I picking this up wrong?

application/classes/Ushahidi/Validator/Post/Create.php
```diff
@@ -331,7 +334,33 @@ public function checkRequiredStages(Validation $validation, $completed_stages, $
 	 * @param  Array      $attributes
 	 * @param  Array      $fullData
 	 */
-	public function checkRequiredAttributes(Validation $validation, $attributes, $fullData)
+	public function checkRequiredPostAttributes(Validation $validation, $attributes, $fullData)
+	{
```
willdoran: I think we're still missing attributes though, what about required attributes not on required stages?

rjmackay: Hrm... I don't even know what should happen with them. Technically required attributes aren't required until their stage is 'complete' ... but thats completely counter intuitive.

I'll defer solving that issue to later. This is still much better than before.

rjmackay:
> I'm not sure I undertstand what you mean by required Post fields? All attributes are children of a task so if you mean the "post task" i.e. the task that is not divided into its own section. I think it would be checked in the same way as the other tasks, or am I picking this up wrong?

The `post task` doesn't have a "Mark completed" toggle and will never get flagged as complete so will never be checked.

Issue #1975, opened 2017-08-07
CSV is broken


jshorland:
### Expected behaviour
When I navigate to uchaguzi.or.ke as a logged in admin and export all data with no filters selected, I should get ALL posts

### Actual behaviour
I get 20 rows of data -- I think it might be giving me only published?

### Steps to reproduce the behaviour/error
1. login to uchaguzi.or.ke as an admin
2. don't mess with filters AT ALL (you should be seeing all posts with locations on the map which is centered in KE)
3. go to share --> export
4. open the exported CSV file 
5. there are only 20 posts.

willdoran:
This is failing to include the filters from client, will fix now

willdoran:
@jshorland I believe this is fixed

jshorland:
PASSES QA.

Issue #1970, opened 2017-08-07
CSV export not working correctly


larsbromley:
The CSV export on https://uchaguzi.or.ke/views/map is not producing nicely structured CSV data. Instead, the data is very confused / garbled, with repetitive field headers and end-of-line errors. And, it seems the filters are not applied. To see this, click 'Share' and 'Export to CSV' and look at the results.

Celestine22:
#1968 linked to this

jshorland:
Duplicate. Closing in favor of #1968

PR #2015, opened 2017-08-15
1975 Returning Entity names as well as ids in CSV

willdoran:
- Fixes export to correctly replace tag ids with names
- Fixes export to correctly replace set ids with names
- Fixes export to correctly replace form ids with names
- Fixes export to correctly replace completed stages with names

Test checklist:
- [ ]

Fixes ushahidi/platform#1975
Fixes ushahidi/platform#1970

Ping @ushahidi/platform

### DIFF
--- a/application/classes/Ushahidi/Repository/Post/Export.php
+++ b/application/classes/Ushahidi/Repository/Post/Export.php
@@ -13,13 +13,71 @@ use Ushahidi\Core\Entity\PostRepository;
 
 class Ushahidi_Repository_Post_Export extends Ushahidi_Repository_Post
 {
-  public function getFormAttributes($values) {
+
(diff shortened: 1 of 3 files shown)


application/classes/Ushahidi/Repository/Post/Export.php
```diff
@@ -13,13 +13,74 @@
 
 class Ushahidi_Repository_Post_Export extends Ushahidi_Repository_Post
 {
-  public function getFormAttributes($values) {
+
+  public function transformData($data) {
```
rjmackay: Am I right that this is basically returning column names for export? Could you rename it to reflect that?

rjmackay:
Tested, works for me.

Issue #1956, opened 2017-08-06
Limit rights to editing saved searches/collections to roles


aoduor:
### Expected behaviour
I'd like to limit the ability to alter my existing saved searches to specific roles

### Actual behaviour
Right now, any logged in users who can view saved searches are able to edit it. (I had two of my saved searches that are crucial to the uchaguzi workflow broken by an unintended edit)

aoduor:
Related to https://github.com/ushahidi/platform/issues/1859

aoduor:
And to https://github.com/ushahidi/platform/issues/1711

willdoran:
This is done: to test it

- [ ] Create saved search
- [ ] set edit role to admin only
- [ ] set visibility to everyone
- [ ] save
- [ ] Log out 
- [ ] Log in as member
- [ ] go to savedsearch
- [ ] change the filters
- [ ] look for the edit button, you shouldn't see it or the options menu
- [ ] look for the reset/update savedsearch, you should see neither

jshorland:
@rjmackay I QA'd this twice. Good to go.

PR #2025, opened 2017-08-23
Adding edit role control to savedsearches

willdoran:
- Add a setting for who can edit a saved search

Test checklist:
- [ ] bin/behat

Fixes ushahidi/platform#1956

Ping @ushahidi/platform

### DIFF
--- a/src/Core/Entity/Set.php
+++ b/src/Core/Entity/Set.php
@@ -23,9 +23,11 @@ class Set extends StaticEntity
 	protected $view;
 	protected $view_options;
 	protected $role;
+	protected $edit_role;
 	protected $featured;
 	protected $created;
 	protected $updated;
+	
 
 	// DataTransformer
 	protected function getDefinition()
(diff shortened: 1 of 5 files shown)


rjmackay:
To do:
- Add feature limit

rjmackay:
@jshorland this PR still allows access to editing a saved search to anyone with the 'manage roles' permission. I'd suggest we should remove that. This would mean a saved search can only be edited by the author, an admin, or someone with a role (specified when creating the search).

I actually think the last condition isn't that useful, the core of this issue on Uchaguzi was that "Manage posts" grants access to too many things, and should have given access to other users saved searches or collections.

src/Core/Entity/Set.php
```diff
@@ -39,6 +41,7 @@ protected function getDefinition()
 			'view'         => 'string',
 			'view_options' => '*json',
 			'role'   => '*json',
```
rjmackay: we should rename `role` to `view_role` or `can_view`

rjmackay:
I think this is now replaced by #2075

Issue #947, opened 2016-02-18
CSV import fails on Mac OS9 line endings


rjmackay:

rjmackay:
![manage users qa deployment](https://cloud.githubusercontent.com/assets/7965/13144048/e45163ba-d6ac-11e5-84ea-13078c936603.png)

jshorland:
Also received this error from a .csv saved on microsoft for mac: The Array Must Contain Unique Values

willdoran:
That's interesting. I believe we need to set
auto_detect_line_endings=1

willdoran:
I think we can do that in
https://github.com/ushahidi/platform/blob/e834389e2443d0598e9be477540fbf22d814c09f/application/bootstrap.php

is that correct @zhalloran @rjmackay

rjmackay:
@willdoran makes sense. Can you test it out and PR?

rjmackay:
@rowasc this might be worth checking while you're hacking on other CSV issues?

rowasc:
@rjmackay will do.

rowasc:
This was done and tested. @jshorland @willdoran . Closing. (Feel free to reopen is this is wrong! )

PR #2057, opened 2017-09-13
947 - auto detect line endings added to CSV import

rowasc:
## This pull request makes the following changes:
- auto detect line endings added to CSV import

## Test checklist:
- [ ] Import a CSV File like 
[os9.csv.txt](https://github.com/ushahidi/platform/files/1299945/os9.csv.txt)
- [ ] In the step 2 of the import where you see the column mappings, make sure that you can select `only` headings, not column items from normal rows with values (ie: you should be able to map `Title` to `Name`, not `Devin Smith` )
- [ ] After submitting the import, make sure you see a successful import message for 18 rows. 

Fixes ushahidi/platform#947 .

Ping @ushahidi/platform


rowasc:
@rjmackay @willdoran no idea why but when I tried to write a test with a file using OS9 line endings, it never fails, even if I disable the new line endings code

tests/integration/csv.feature
```diff
@@ -11,6 +11,17 @@ Feature: Testing the CSV API
         And the "columns.0" property equals "title"
         Then the guzzle status code should be 200
 
+    Scenario: Uploading a CSV file with OS9 line endings
+        Given that I want to make a new "CSV"
+        And that the post file "file" is "tests/datasets/ushahidi/sample_os9.csv"
```
rowasc: this will always pass as long as the file has the content, it doesn't care that the line endings are wrong, for some reason (if I disable the feature for line endings, it still passes)

rowasc: ah.. ok I see the problem... will send an update in a few minutes

rowasc:
@willdoran  @rjmackay this is ready for review.

rowasc:
FYI I didn't add the code to the bootstrap.php directly only because there's a tiny difference in performance according to the docs, so it was worth it to enable this only for the actual import path.

application/classes/Controller/Api/CSV/Import.php
```diff
@@ -29,6 +29,12 @@ protected function _resource()
 
 	public function action_post_index_collection()
 	{
+		/**
+		 * Step two of import.
+		 * Support all line endings without manually specifying it (primarily added because of OS9 line endings which do not work by default )
+		 */
+		ini_set('auto_detect_line_endings', 1);
```
willdoran: Is there a downside to setting this option in the global ini? We can do this via ansible config. Or is there a reason we want this setting only during the execution cycle of the import?

rowasc: Yeah so mostly I went for the specific execution cycle because the docs mentioned a small performance downgrade when using this setting. maybe it wasn't worth it.We can change it back to init or to ansible if you think that's the best way to go. 
https://secure.php.net/manual/en/filesystem.configuration.php#ini.auto-detect-line-endings

rowasc: Now that I think about it, it's probably not going to affect `most` of the app tho, because it'll only really affects us in fgets/file related execution flows..
So, this might have been a silly thing to do.

rjmackay:
@rowasc @willdoran can this land? It looks 99% complete right?

rowasc:
@rjmackay it works fine. There's the pending question of whether we want to move this to config files outside of the repo instead. 
@willdoran after thinking more about this, and with your feedback,  I believe we either move this to the bootstrap script directly or leave it as is.  Sending it to another repo/server configuration isn't worth the problems/extra work it may cause especially to OSS users (to me). Let me know what you think, and we can move from there. 
Thanks
Romina

willdoran:
@Romina-Suarez I agree, let's leave it as is. I'll merge it now.

Issue #2028, opened 2017-08-25
CSV export should order columns by field order in survey


rjmackay:
### Expected behaviour
CSV export should order columns by field order in survey

### Actual behaviour
CSV column order is random

rjmackay:
@Angamanga or @rowasc I've assigned you both, basically whoever gets to this first should make a start. Comment when you get started so you're not duplicating work...

rowasc:
@Angamanga I plan to start this today.

Angamanga:
@rowasc 👍 Thank you! I am still working on the dataproviders :)

rowasc:
Assumption: When there are multiple surveys in the same export, we group the _survey fields_ together

rowasc:
This was failing in QA so I had to revert. 
Waiting for a dump/logs to see what's happening.

PR #2060, opened 2017-09-14
2028 - Custom Sort for CSV Heading 

rowasc:
####  Changes to the original 2028
- Simplified the labels logic to be all arrays and have a count property consistently - this helps fix a bug with arrays that can be null, which used to break the export. 
- added some label cleanup for situations where there's a multivalue field but only 1 value (don't show index) so that it's more readable whenever possible

#### Changes in 2028 that are here.

- Respects survey id, survey stage, priority number, alphabetical priority (in that order)
- Important: For fields that are part of the post itself, including title and description, we do not do stage and priority sorting. They are common fields and they are sorted alphabetically before any custom field is shown in the CSV
**The sorting procedure is:** 
- All native fields go first
- After native fields
--- group by survey ID and stage ID
--- sort groups by numerical key (ie: form id =2 + stage id = 2 go after form id 1, stage id 2)
--- sort by priority inside each group. if priority is the same, fall back to alphabetical order
--- flatten, attach to native fields
--- get all records exported with the correct field order by iterating the heading attributes for each record and printing if there is a value.  

## TODO 

- [x] Group and sort with survey (so that if you have 4 fields, two in each survey, they are next to each other and not just sorted by priority + stage which would usually match p1 - p1 , s1-s1, p2-p1,p2-p1, etc)

Test checklist:
- [ ] Make sure to test with a database that includes fields like completed stages being Null for some entries and array for others.  ie: QA Database
- [ ] Create one of each type of survey fields in a survey.
    - [ ] Sort them however you like. 
    - [ ] Add posts to the survey, filling the fields . Make sure you publish the posts. 
    - [ ] Export posts.
    - [ ] The CSV heading should match the sort order of your survey. 

- [ ] Create a second survey with some custom fields
    - [ ] Sort them however you like. 
    - [ ] Add posts to the survey, filling the fields . Make sure you publish the posts. 
    - [ ]  Make sure your other survey (both) has published posts.
    - [ ] Export posts.
    - [ ] The CSV heading should match the sort order of your survey, and each set of survey fields should be grouped by survey. 

- [ ] run ./bin/behat

Fixes ushahidi/platform#2028

Ping @ushahidi/platform

### DIFF
--- a/application/classes/Ushahidi/Formatter/Post/CSV.php
+++ b/application/classes/Ushahidi/Formatter/Post/CSV.php
@@ -107,69 +107,20 @@ class Ushahidi_Formatter_Post_CSV implements Formatter
 	 * @return array of sorted fields with a zero based index. Multivalue keys have the format keyxyz.index index being an arbitrary count of the amount of fields.
 	 */
 	private function createSortedHeading($fields){
-		$headingResult = [];
-		// fieldsWithPriorityValue: an associative array with the form ["uuid"=>[label: string, priority: number, stage: number],"uuid"=>[label: string, priority: number, stage: number]]
-		$fieldsWithPriorityValue = [];
-		/**
-		 * Assign $heading and $fieldsWithpriorityValue.
-		 */
-		$this->setPriorityAndNativeFieldArrays($headingResult, $fieldsWithPriorityValue, $fields);
-		/**
-		 * Sort the non custom priority fields alphabetically, ASC (default)
-		 */
-		uasort($headingResult, function($item1, $item2){
-			return strcmp($item1, $item2);
-		});
 		/**
 		 * sorting the multidimensional array of properties
 		 */
 		/**
 		 * First, group fields by stage and survey id
 		 */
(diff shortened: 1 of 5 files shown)


rowasc:
Merging so we can start testing this in QA env. 
@rjmackay let me know if you have any feedback 
Thanks

Angamanga:
@rowasc Results from QA: It works fine, the sorting is correct when exporting posts from different surveys. The only thing I wonder about is the tags-ids that is exported as a native field. Do we need those or is it enough with the category-names?

rowasc:
@Angamanga Thanks for the feedback.
@willdoran sending a PR with this adjustments here https://github.com/ushahidi/platform/pull/2060

Angamanga:
@rowasc The tags-ids in the are now gone and the rest of the headers are the same so it looks good!

rowasc:
Great . Thanks @Angamanga . I checked the `run /bin/behat` checkbox as well since that's running both on CI and locally.

PR #2061, opened 2017-09-16
Merge branch 'develop' into statefuldata-nested-change-tracking

rjmackay:
- Reapply #1938 
- Adding recursive array diff function
- Updating StatefulData::setData to check nested arrays for change
- Updating StatefulData::hasChanged to support multi-level nested key checking

Test checklist:
- [ ] bin/behat

Fixes ushahidi/platform# .

Ping @ushahidi/platform


rjmackay:
This has an issue where its comparing an array vs a string. It doesn't check the structure is the same first

rjmackay:
@willdoran this is just reapplying your stateful data change. I *think* it'd probably be quick to fix if you want to try and land it before Uchaguzi begins.

Issue #2078, opened 2017-09-20
Enforce saving the user_Id when creating a saved search  or Collection


rowasc:
### Expected behaviour
When I create a saved search or a collection as any user, my ID is linked to it
### Actual behaviour
When I create a saved search  or a collection  as any user, my ID is not presentt in the database. 

### Steps to reproduce the behaviour/error
- Create a saved search
- Verify that database table `sets` is not saving the user_id => This should be saved

### Steps to reproduce the behaviour/error (Collections)
- Create a collection
- Verify that database table `sets` is not saving the user_id => This should be saved

rowasc:
There are 2 parts. 
 - Client needs to send the data. https://github.com/ushahidi/platform-client/pull/757
- User id has to be required in the backend.  [pr pending]

rowasc:
>Client needs to send the data. ushahidi/platform-client#757
note
Important: because of the implementation we went for, this is not true. We don't `need` the user id to go in the request ,and you can't  really assign a different user_id because we get it from the current user making the request

PR #2079, opened 2017-09-20
2078 - Make user id a required field when creating a Collection or Saved search

rowasc:
- Make user id a required field when creating a Collection or Saved search

Test checklist:
- [ ] run ./bin/behat
- [ ] Create a new saved search as a normal user and edit as admin. The user id should remain the same as the original user
- [ ] Create a new saved search as a normal user and edit as the same user. The user id should remain the same as trhe original user
- [ ] Create a new saved search and force a different user_id in your request than the one you are logged in with. The backend should throw a validation error. 

Fixes ushahidi/platform#2078 .

Ping @ushahidi/platform


rowasc:
ok. I'll do that instead. Thanks.

rowasc:
@willdoran can you take a look ? thanks

Issue #1924, opened 2017-07-19
Bug reporting mechanism


caharding:
- People can report a bug web client and mobile
- Bug tool captures user log and time stamp
- Bug tool can submit a screenshot

rjmackay:
@jshorland over to you, this seems like a really minor priority to me.

PR #2103, opened 2017-10-03
Adding Post Locks to Post attribute

willdoran:
- Adds Post Locking for Posts
- Adds Post Locking Authorizer which uses Post Authorizer
- Adds lock privillege
- Adds basic integration tests

Test checklist:
- [ ] Ensure lock is created for Post
- [ ] Ensure lock is broken for Post
- [ ] Ensure only Users allowed to edit Posts can create/break locks

Fixes ushahidi/platform#1924

Ping @ushahidi/platform


application/classes/Controller/Api/Posts/Lock.php
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @author     Ushahidi Team <team@ushahidi.com>
+ * @package    Ushahidi\Application\Controllers
+ * @copyright  2017 Ushahidi
+ * @license    https://www.gnu.org/licenses/agpl-3.0.html GNU Affero General Public License Version 3 (AGPL3)
+ */
+class Controller_API_Posts_Lock extends Ushahidi_Rest {
+	protected $_action_map = array
+	(
+		Http_Request::PUT     => 'put',
```
rowasc: nice to see this being limited to only what we need.

application/classes/Ushahidi/Repository/Post/Lock.php
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	public function releaseLockByLockId($lock_id)
+	{
+		$query = DB::delete('post_locks')
+			->where('id', '=', $lock_id);
+		return $query->execute();
+	}
+
+	public function checkLock($entity_id)
```
rowasc: isActive or isExpired might make it easier to read. $this->isActive($is) vs !$thick->checkLock($id)

rowasc: At the same time, all the other methods in the class use doSomething`Lock`() instead of doSomething() so maybe it makes sense to name is isLockActive instead of isActive. Usually I prefer ->get vs getLock in this scenario, or ->create vs ->createLock unless we were adding that method on the post repository for some reason.

application/config/init.php
```diff
@@ -28,7 +28,7 @@
 	'base_url'    => '/',
 	'index_file'  => FALSE,
 	'charset'     => 'utf-8',
-	'errors'      => TRUE,
+	'errors'      => false,
```
rowasc: what does this do? (or, why was that True before)

src/Core/Entity/PostLock.php
```diff
@@ -0,0 +1,52 @@
...
+		];
+	}
+	// DataTransformer
+	protected function getDefinition()
+	{
+		return [
+			'id'            	=> 'int',
+			'user'          	=> false,
+			'user_id'       	=> 'int',
+            'post'          	=> false,
```
rowasc: FYI indentation seems weird here, I think because it's mixing spaces and tabs?

src/Core/Tool/Authorizer/PostAuthorizer.php
```diff
@@ -114,7 +114,7 @@ public function isAllowed(Entity $entity, $privilege)
         }
 
         // Non-admin users are not allowed to create posts for forms that have restricted access.
-        if (in_array($privilege, ['create', 'update'])
+        if (in_array($privilege, ['create', 'update', 'lock'])
```
rowasc: check: lock is a permission?

src/Core/Usecase/Post/DeletePostLock.php
```diff
@@ -0,0 +1,39 @@
...
+    // Usecase
+	public function interact()
+	{
+		// Fetch a default entity and apply the payload...
+		$post = $this->getPostEntity();
+
+		// ... verify that the entity can be created by the current user
+		$this->verifyLockAuth($post);
+
+		if ($this->getIdentifier('lock_id')) {
```
rowasc: weird indentation. Is it just github being weird?

application/classes/Ushahidi/Core.php
```diff
@@ -236,6 +236,10 @@ public static function init()
 			'update' => $di->lazyNew('Ushahidi_Validator_Post_Create'),
 			'import' => $di->lazyNew('Ushahidi_Validator_Post_Import'),
 		];
+		$di->params['Ushahidi\Factory\ValidatorFactory']['map']['posts_lock'] = [
```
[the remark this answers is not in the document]
willdoran: Fixed

application/classes/Ushahidi/Repository/Post/Lock.php
willdoran: Excellent notes, I'll review the naming.

application/config/init.php
willdoran: It shouldn't have changed! Good catch. So it's stupid, Kohana error logging is broken because it's basically decayed. If this is set to True Kohana intercept the error message but fails to actually except it properly. By disabling it, the errors pass straight through. I forgot to put it back to True for merge!

src/Core/Entity/PostLock.php
willdoran: Will fix

src/Core/Tool/Authorizer/PostAuthorizer.php
willdoran: In this context, it's a privilege which is more like an action. There is a divide between these and permissions. The permissions operate as separate controls, this is more like a descriptor of the action being performed. It is possible to restrict what actions are possible.

src/Core/Usecase/Post/DeletePostLock.php
willdoran: It might be my editor I'll check

application/classes/Ushahidi/Repository/Post/Lock.php
willdoran: for get, there is a default get method applied from a parent so I'm going to keep getLock to differentiate

application/config/init.php
rowasc: ahh good. Thanks for explaining 👍

application/classes/Ushahidi/Repository/Post/Lock.php
rowasc: Makes sense 👍

application/classes/Ushahidi/Validator/Post/Lock/Update.php
```diff
@@ -0,0 +1,34 @@
+<?php defined('SYSPATH') OR die('No direct access allowed.');
```
willdoran: Validation for lock existence should be added to the Post Update validator

willdoran:
TODO:
-  [ ] Refactor validation to use non-dependent field
-  [ ] Refactor Lock Repo to use more ORM structure
-  [ ] Add missing usecases

PR #2024, opened 2017-08-23
Task visibility

willdoran:
This pull request makes the following changes:
- Add `task_is_internal` flag to stages
- [ ] Add migration?

Test checklist:
- [ ] bin/behat

Fixes ushahidi/platform#1945

Ping @ushahidi/platform

PR #2113, opened 2017-10-05
Initial version of Task Visibility

willdoran:
- Adds new field internal only to form stage
- Adds migration to ensure new field is set for existing form stages
- Extends test to ensure that restricted form stages or their attributes are not returned unless the user has manage posts permission

Test checklist:
- [ ] ./bin/behat
- [ ] Use this PR in conjunction with ushahidi/platform-client#781 to complete test checklist

Fixes ushahidi/platform#2024

Ping @ushahidi/platform

### DIFF
--- a/migrations/20171005171849_set_default_task_only_for_existing_form_stage.php
+++ b/migrations/20171005171849_set_default_task_only_for_existing_form_stage.php
@@ -8,26 +8,21 @@ class SetDefaultTaskOnlyForExistingFormStage extends AbstractMigration
     public function up()
     {
         $pdo = $this->getAdapter()->getConnection();
-        $rows = $this->fetchAll("SELECT task_is_internal_only from form_stages");
         $insert = $pdo->prepare(
             "UPDATE form_stages
             SET task_is_internal_only = false"
         );
-        foreach ($rows as $row) {
-            $insert->execute();
-        }
+
+        $insert->execute();
     }
 
     public function down()
     {
         $pdo = $this->getAdapter()->getConnection();
-        $rows = $this->fetchAll("SELECT task_is_internal_only from form_stages");
         $insert = $pdo->prepare(
             "UPDATE form_stages
             SET task_is_internal_only = true"
         );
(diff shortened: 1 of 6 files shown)


migrations/20171005171849_set_default_task_only_for_existing_form_stage.php
```diff
@@ -0,0 +1,33 @@
...
+
+    public function up()
+    {
+        $pdo = $this->getAdapter()->getConnection();
+        $rows = $this->fetchAll("SELECT task_is_internal_only from form_stages");
+        $insert = $pdo->prepare(
+            "UPDATE form_stages
+            SET task_is_internal_only = false"
+        );
+        foreach ($rows as $row) {
```
rowasc: @willdoran isn't this just going to iterate each row and do the same thing to all of them? I don't see a reason we need to run a foreach here since the update will have done that for all rows in the first run.

migrations/20171005171849_set_default_task_only_for_existing_form_stage.php
```diff
@@ -0,0 +1,33 @@
...
+            $insert->execute();
+        }
+    }
+
+    public function down()
+    {
+        $pdo = $this->getAdapter()->getConnection();
+        $rows = $this->fetchAll("SELECT task_is_internal_only from form_stages");
```
rowasc: same as above comment

migrations/20171005171849_set_default_task_only_for_existing_form_stage.php
willdoran: hahahahaha...yes the foreach is irrelevant!

willdoran: Fixing

migrations/20171005171849_set_default_task_only_for_existing_form_stage.php
willdoran: Fixing

Issue #1864, opened 2017-07-12
Change log of posts


aoduor:
It's important to be able to keep track of logs of who has been in contact with a particular post for accountability purposes and backtracking too.

Aha! Link: https://ushahiditeam.aha.io/features/PROD-583

jshorland:
Simple log of who edited the post when is enough. Doesn't have to be pretty for now, but should be accessible by clicking a link on the post to view edit log.

jshorland:
I'm going to re-open this as a more refined version of this will be necessary for Epic: tasks

jshorland:
Permissions:
- [ ] Change log should only be visible to logged in users (not to members, or visitors)
- [ ] Manual entries can only be created by users with permission to edit posts

Any saved changes should appear in the change log automatically, indicating who changed what when (who = anonymous if author n/a)
- [ ] Post status change (created as draft, published, archived)
- [ ] Category added or removed
- [ ] Survey field edited (indicate which field was edited if possible)
- [ ] Task field edited (indicate which field was edited if possible)
- [ ] Task completed or changed back to incomplete 
- [ ] Added to or removed from collection 

Users with permission can create their own change log entries manually
- [ ] only users with the edit posts permission should see "add an entry" button
- [ ] clicking on add an entry opens long text box titled "new entry" with asterisk indicating text is required to save
- [ ] cancel option closes the box without saving
- [ ] Clicking save without any text shows error dialogue "You must enter something in the change log before you save it."
- [ ] saving with text updates the change log to reflect the new manual entry. Entry is visible to any logged in user (NOT TO MEMBERS OR VISITORS). It should show who created it and when.

jshorland:
Designs
![change. log.png](https://zube.io/files/ushahidi/06d9636eb4b272b589e1cd3ae62194a5.png)

willdoran:
Notes:

- On delete message warning that changelog will be deleted as well, suggest archive

kinstelli:
Spec/Security question @jshorland 

 Regarding _Change log should only be visible to logged in users (not to members, or visitors)_, I'm wondering if there might be a situation where items that could be visible in the changelog, such as Collection names, Tag names, etc, should only be seen by admins, rather than just any logged in user.

And if so, would we want to have different log visibility that only shows certain lines to certain users?

jshorland:
@kinstelli good nuance here. There are items that will be listed in the changelog that have permissions associated with them. They are the following:
- categories
[...]

PR #2127, opened 2017-10-11
1864 changelog backend [WIP] 

kinstelli:
- Adds a posts_changelog table
- Intercepts updates to Post fields, writes content to a table
- Provides API for manually writing comments to the changelog
- Provides route for viewing changelog of each Post
- Restricts changelog access via the API

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


application/classes/Ushahidi/Formatter/Post/ChangeLog.php
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Ushahidi API Formatter for PostsChangelog
+ *
+ * @author     Ushahidi Team <team@ushahidi.com>
+ * @package    Ushahidi\Application
+ * @copyright  2014 Ushahidi
+ * @license    https://www.gnu.org/licenses/agpl-3.0.html GNU Affero General Public License Version 3 (AGPL3)
+ */
+
+ use Ushahidi\Core\Exception\FormatterException;
```
willdoran: spaces are off I think

application/classes/Ushahidi/Listener/PostListener.php
```diff
@@ -33,15 +34,117 @@ public function setWebhookRepo(WebhookRepository $webhook_repo)
 		$this->webhook_repo = $webhook_repo;
 	}
 
-  public function handle(EventInterface $event, $post_id = null, $event_type = null)
+	public function setChangeLogRepo(PostsChangeLogRepository $changelog_repo)
+	{
+			$this->changelog_repo = $changelog_repo;
+	}
+
+//TODO: this is ugly, because we're only receiving the ID here.
```
willdoran: Indent

application/classes/Ushahidi/Listener/PostSetListener.php
```diff
@@ -15,24 +15,58 @@
...
+		];
+
+		//handle changes to collections
+		$changelog_entity = $this->changelog_repo->getEntity();
+		$changelog_entity->setState($changelog_state);
+		$this->changelog_repo->create($changelog_entity);
+	}catch (Exception $e)
+	{
+		Kohana::$log->add(Log::INFO, 'trying to handle a collection change event.' );
+	}
```
willdoran: Indenting I think I wrong here too

src/Core/Tool/Authorizer/PostsChangeLogAuthorizer.php
```diff
@@ -0,0 +1,93 @@
...
+	public function __construct(PostRepository $post_repo, PostAuthorizer $post_auth)
+	{
+		$this->post_repo = $post_repo;
+		$this->post_auth = $post_auth;
+	}
+
+	/* Authorizer */
+	public function isAllowed(Entity $entity, $privilege)
+	{
+		//$post = $this->getPost($entity);
```
willdoran: This can be removed.

src/Core/Tool/Authorizer/PostsChangeLogAuthorizer.php
```diff
@@ -0,0 +1,93 @@
...
+		$user = $this->getUser();
+
+		// First check whether there is a role with the right permissions
+		if ($this->acl->hasPermission($user, Permission::MANAGE_POSTS)) {
```
willdoran: Since changelog should derive from Post permissions we could inherit from the Post Authorizer and just call form_auth->isAllowed() but I think it makes sense for these sub entities to have their own permissions.

The alternative way to do this is here:

https://github.com/ushahidi/platform/blob/develop/src/Core/Tool/Authorizer/FormStageAuthorizer.php#L41

I'm not sure which is better to be honest. I think stick with this for now since I think it is more restricted than a normal Stage or Attribute might be.

willdoran:
@kinstelli I think this is still paused until develop is stable?

rowasc:
Closing per agreement that stale PRs won't remain in the list of open PRs. The branch will still be around and the ticket ushahidi/platform#1864 will remain open when we want to pick it up .

Issue #2152, opened 2017-10-16
Survey-fields does not appear when not logged in


Angamanga:
### Expected behaviour
When I go to a survey as not logged in, all fields I have have permissions to see should be visible

### Actual behaviour
If I create a survey 'normal' fields and a internal task, no fields except title and description are visible when trying to submit a post. This only happens when the survey is created with internal tasks. If creating a survey and adding internal tasks later on in edit-mode, it does not happen.

### Steps to reproduce the behaviour/error
- [ ] Create a survey with internal tasks
- [ ] Log out
- [ ] Add a post, only title and description is visible

##### Where
- [x ] [Local setup with Vagrant ](https://www.ushahidi.com/support/install-ushahidi#installing-for-development)
- [ ] [Local setup from platform-release ](https://www.ushahidi.com/support/install-ushahidi#installing-the-latest-release)
- [ ] Ushahidi.io / SaaS solution 
- [ x] Ushahidi's QA environment
- [ ] Other (explain): sandbox.uchaguzi.or.ke

jshorland:
When Logged out, I still only see Title and Description field for "anna testing" survey.
<img width="1275" alt="screen shot 2017-10-19 at 10 50 30 am" src="https://user-images.githubusercontent.com/3224856/31785871-658a82a2-b4bb-11e7-8c0a-e48d02ec0923.png">

PR #2173, opened 2017-10-17
2152 task visibility

willdoran:
- Setting Visibility logic as follows
  - "Show when published" takes precedence as follows
    - "Show when published" and "Internal only" set 
      - users with Manage Post or Admin can see the task when creating editing.
      - users/anon will only see the task when the post is published
    - "Show when published" set and "Internal only" *not* set
      - users with Manage Post or Admin can see the task when creating editing.
      - users/anon can see the task when creating editing.
      - users/anon will see the task when the post is published
    - "Show when published" not set
      - users with Manage Post or Admin can see the task when creating editing.
      - users with Manage Post or Admin can see the task when the post is published
      - users/anon can never see the task

Test checklist:
- Setting Visibility logic as follows
  - [ ] "Show when published" takes precedence as follows
    - [ ] "Show when published" and "Internal only" set 
      - [ ] users with Manage Post or Admin can see the task when creating editing.
      - [ ] users/anon will only see the task when the post is published
    - [ ] "Show when published" set and "Internal only" *not* set
      - [ ] users with Manage Post or Admin can see the task when creating editing.
      - [ ] users/anon can see the task when creating editing.
      - [ ] users/anon will see the task when the post is published
    - [ ] "Show when published" not set
      - [ ] users with Manage Post or Admin can see the task when creating editing.
      - [ ] users with Manage Post or Admin can see the task when the post is published
      - [ ] users/anon can never see the task

Fixes ushahidi/platform#2152

Ping @ushahidi/platform


Angamanga:
@will this works according to the test-check-list but it still messes up getting the attributes that are not attached to a task when trying to add the post as anon-user. When not logged in, the form-attribute-endpoint are not returning any attributes as all. And this only happens for scenario 1, when internal-only is set. (This is not returning any attributes when not logged in: https://github.com/ushahidi/platform-client/blob/develop/app/main/posts/modify/post-data-editor.directive.js#L151)

willdoran:
@Angamanga hmmm, and this fails when used with the client PR, right? 
https://github.com/ushahidi/platform-client/pull/839

willdoran:
@Angamanga It's working for me with the Client PR, could we pair on this when you're back on? I must be missing how to produce it!

PR #2180, opened 2017-10-18
Basic version of CLI CSV Exporter

willdoran:
- This adds a CLI which wraps around the Post Export Repo

Test checklist:
- [ ] It should run and get a number of posts based on the defined limit, currently set to 100

Next steps:
- [ ] Add ability to read JSON file of status filters
- [ ] Add ability to chunk CSV files automatically
- [ ] Add ability to use Export user
- [ ] Add ability to write to file system

Fixes ushahidi/platform# .

Ping @ushahidi/platform


application/classes/Ushahidi/Repository/Post.php
```diff
@@ -526,19 +526,21 @@ protected function setSearchConditions(SearchData $search)
...
-		if (!$user->id) {
-			$query->where("$table.status", '=', 'published');
-		} elseif (!$this->isUserAdmin($user) and
-				  !$this->acl->hasPermission($user, Permission::MANAGE_POSTS)) {
-			$query
-				->and_where_open()
-				->where("$table.status", '=', 'published')
-				->or_where("$table.user_id", '=', $user->id)
-				->and_where_close();
+		if (!$search->exporter) {
```
rowasc: just a question..  I think if we use the console authorizer we don’t need this, right? (like in the user export)

rowasc:
I will test it as soon as the branch is up to date/running tests, LMK!

application/classes/Ushahidi/Repository/Post.php
willdoran: That's a good point, I'll fix that.

rowasc:
@willdoran messages are not downloaded. Is that expected?

willdoran:
I believe so, though I can add them if desired. This is only exporting Posts.

rowasc:
ehhh, weird thing: chunks are not all the same size?? @willdoran

rowasc:
@willdoran is that because then we apply the filters and some get left out? I think?

willdoran:
@rowasc yeah it might be, actually

willdoran:
This version doesn't support filtering at all, it just gets everything.

Issue #2217, opened 2017-10-24
Large CSV download succeeds, but never prompts me for where to save the file??


rjmackay:
### Expected behaviour
CSV download works and asks me where to save the file

### Actual behaviour
CSV download runs (HTTP request succeeds!) but the file is never saved

### Steps to reproduce the behaviour/error
- Go to data view
- Select Share
- Pick export to CSV

##### Where
- [ ] Other (explain): Uchaguzi Sandbox

rowasc:
@rjmackay this is the browser pretty much crashing with those large files that are not being downloaded (we get it as normal content, not with the correct headers, then we manipulate it, etc) . When the number of posts is enough that you can still get them but it doesn't prompt for download, we have verified that you can get them with postman/curl/whatever because they don't crash like the browser on that response.  
TLDR:  the browser thinks that the download is still in progress

rjmackay:
Ugh. Everything is terrible!
Do we just need to fix the headers then? or something else?

rowasc:
Yes, that's an accurate description. :(
So, we need a few things. 
- we need to fix the headers to do a proper download, and related to that: 
- -  we need to STOP manipulating our CSV in the client (!)
- -  we need to have a way to access those CSVs ( a list) because otherwise the user will be super confused and they will lose their CSV processing request when they leave a page. 
- - we need some way (queues, we think) to scale how we do CSVs on the backend , and obviously to improve the sorting which is terrible 
-  in the backend area, we need to fix CSV performance in general. Will moved things to a console command, which helped somewhat because at least you can get them offline if you are in need (as a cli user.. ugh) but  it still ends up running out of memory with very large datasets

rowasc:
https://github.com/ushahidi/platform/issues/2121
https://github.com/ushahidi/platform/issues/2109
Related issues we have been creating...

jshorland:
@rowasc i'm thinking about increasing this to p0. it completely stops the user from being able to download to CSV, correct?

jshorland:
Yeah this is definitely a thing. Making this a p0. I just continue to get "download failed -
 network error" when trying to download ALL the data out of sandbox.uchaguzi.or.ke. Not sure if related, but there's no way for me to download large data sets.

rjmackay:
Seems to be working in sandbox

jshorland:
UPDATE: this is NOT working in sandbox anymore, nor in uchaguzi.

rowasc:
[...]

PR #2250, opened 2017-10-25
2217 CSV headers correctly set for download

rowasc:
- 2217 CSV headers correctly set for download, so we handle an arraybuffer in the FE and support larger files. 

Test checklist:
- [ ]  Test a download with all posts, while logged in (in sandbox this should be > 5k posts) . It should work as expected by creating a file and downloading it with the same file name format we had in the past ({{hostname}}-{{date}}.csv) 

Fixes ushahidi/platform#2217 .

Ping @ushahidi/platform


willdoran:
23k -- timeout
16k -- timeout
11k -- timeout
10k -- timeout
9k -- timeout
8k -- timeout
7k -- timeout
6118 -- works.

PR #2262, opened 2017-10-26
Adding initial version of Post Data Exports entity

willdoran:
- Post Data Export
- Post Data Export Job
- Repos, formatter, validators, migrations, tests

Needs better naming scheme, might be that it can live under Posts
Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


src/Core/Tool/Authorizer/PostDataExportAuthorizer.php
```diff
@@ -0,0 +1,69 @@
...
+	// Check if webhook feature is enabled
+	use WebhookAccess;
+
+
+	/* Authorizer */
+	public function isAllowed(Entity $entity, $privilege)
+	{
+
+		// Check if the webhooks feature enabled
+		if (!$this->isWebhookEnabled()) {
```
willdoran: fix this

migrations/20171026184202_create_post_data_export_table.php
```diff
@@ -0,0 +1,33 @@
...
+
+class CreatePostDataExportTable extends AbstractMigration
+{
+    /**
+     * Migrate Up.
+     */
+    public function up()
+    {
+        $this->table('postdataexports')
+          ->addColumn('user_id', 'integer', ['null' => false])
```
rowasc: filters

rowasc: we might not have a user, maybe contact?

rowasc: maybe we can have total rows

migrations/20171026184208_create_post_data_export_job_table.php
```diff
@@ -0,0 +1,30 @@
...
+
+class CreatePostDataExportJobTable extends AbstractMigration
+{
+
+    /**
+     * Migrate Up.
+     */
+    public function up()
+    {
+        $this->table('postdataexport_job')
```
rowasc: we might want to have the author info here actually

application/classes/Ushahidi/Core.php
```diff
@@ -162,8 +167,11 @@ public static function init()
 
 		// Post Exporter
 		$di->setter['Ushahidi\Console\Application']['injectCommands'][] = $di->lazyNew('Ushahidi_Console_PostExporter');
+		$di->setter['Ushahidi_Console_PostExporter']['setDatabase'] = $di->lazyGet('kohana.db');
 		$di->setter['Ushahidi_Console_PostExporter']['setPostExportRepo'] = $di->lazyGet('repository.posts_export');
 		$di->setter['Ushahidi_Console_PostExporter']['setDataFactory'] = $di->lazyGet('factory.data');
+		$di->setter['Ushahidi_Console_PostExporter']['setPostDataExportRepo'] = $di->lazyGet('repository.postdataexport');
```
rowasc: I think the current problem is that repository.posts_export is no longer an export, but a FilteredPostsCollection  (or, whatever,  I can think of a new and improved name in a bit if this makes some sense) 

The problem with what I'm thinking above is that we already have collections/sets so this makes less sense in that context, it creates more confusion. I'm starting to wonder if we can't instead of post data export use the saved searches system for this by having an 'exportable' flag, but that might be a bit of a stretch, and the requirements are not the same because you don't have to sort and get headers in sets, for instance. I think eventually, tho, this might be the direction to clean things up even more . 
`TLDR: This is an export, this makes sense as an export. But the posts export we HAD, that one is no longer an export. Does that make sense? ` 
I'm still thinking about this, just trying to capture my thoughts and posting here because you might have a better idea based on this rant

rowasc: sorry deleted my previous comment, I mean to clean up a bit more in my head . `MUST READ BEFORE HITTING BUTTON` x 100

rowasc: the gist of it was: 
`TLDR: This is an export, this makes sense as an export. But the posts export we HAD, that one is no longer an export. Does that make sense? `

application/classes/Ushahidi/Formatter/PostDataExport.php
```diff
@@ -0,0 +1,35 @@
...
+
+use Ushahidi\Core\Traits\FormatterAuthorizerMetadata;
+
+class Ushahidi_Formatter_PostDataExport extends Ushahidi_Formatter_API
+{
+	use FormatterAuthorizerMetadata;
+
+	protected function get_field_name($field)
+	{
+		$remap = [
```
rowasc: $remap -> not super sure, why the name?

application/classes/Ushahidi/Repository/PostDataExport.php
```diff
@@ -0,0 +1,92 @@
...
+use Ushahidi\Core\Traits\AdminAccess;
+
+class Ushahidi_Repository_PostDataExport extends Ushahidi_Repository implements PostDataExportRepository
+{
+	use UserContext;
+	use AdminAccess;
+
+	protected function getId(Entity $entity)
+	{
+		$result = $this->selectQuery()
```
rowasc: We should be able to have more than 1 per user.
Am I understanding this wrong?

migrations/20171026184208_create_post_data_export_job_table.php
rowasc: @willdoran after the discussion with Jess and Justin we only need user id, not author, since anon won't be able to export.

PR #2278, opened 2017-11-02
Defer loading current user to avoid issues bootstrapping

rjmackay:
- Migrate user loading to a Session class
- Update UserContext to use Session class
- Add LumenSession implementation

Test checklist:
- [ ] composer test

Ping @ushahidi/platform

### DIFF
--- a/app/Tools/LumenSession.php
+++ b/app/Tools/LumenSession.php
@@ -1,0 +1,36 @@
+<?php
+
+/**
+ * Ushahidi Lumen Session
+ *
+ * @author     Ushahidi Team <team@ushahidi.com>
+ * @package    Ushahidi\Application
+ * @copyright  2014 Ushahidi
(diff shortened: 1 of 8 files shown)


app/Tools/LumenSession.php
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Ushahidi Lumen Session
+ *
+ * @author     Ushahidi Team <team@ushahidi.com>
+ * @package    Ushahidi\Application
+ * @copyright  2014 Ushahidi
+ * @license    https://www.gnu.org/licenses/agpl-3.0.html GNU Affero General Public License Version 3 (AGPL3)
+ */
+
+namespace Ushahidi\App\Tools;
```
rjmackay: Namespacing is probably wrong here. I don't really know what to call the `Session` layer.

Issue #2220, opened 2017-10-24
Terms of service doesn't respond to clicking accept 


jshorland:
### Expected behavior
Once I click accept and continue on the updated terms of service notification, it should go away and I should never see it again

### Actual behavior
It doesn't respond to me clicking it. It stays in place. But if I reload the screen it seems to go away (I haven't been able to get it to pop back up so I'm assuming it worked?)

### Steps to reproduce the behaviour/error
Go to a deployment you haven't logged into since new terms of service pop up was pushed (oct 20). Login 

##### Where
- [ ] [Local setup with Vagrant ](https://www.ushahidi.com/support/install-ushahidi#installing-for-development)
- [ ] [Local setup from platform-release ](https://www.ushahidi.com/support/install-ushahidi#installing-the-latest-release)
- [ ] Ushahidi.io / SaaS solution 
- [ ] Ushahidi's QA environment
- [ ] Other (explain):

crcommons:
@jshorland Can you confirm that the problem is when you have to agree to the terms of service before you attempt to log in? But when you attempt to log in and then agree to the terms of service, it works fine.

crcommons:
I can't reproduce this locally :(

rjmackay:
Fixed and deployed

PR #2282, opened 2017-11-03
Ensure anonymous users can't post to TOS

crcommons:
- Added a check to TOS Authorizer a to see if the user is anonymous because a race condition may cause the TOS functions to run in the client before a user is authenticated. 
- Added associated test to ensure it returns 403. 

*Client changes are still necessary.*

Test checklist:
- [ ] run behat tests

Fixes ushahidi/platform#2220

Ping @ushahidi/platform

### DIFF
--- a/behat.yml.dist
+++ b/behat.yml.dist
@@ -33,4 +33,3 @@ ci:
     suites:
         default:
             filters: ~
-            
--- a/behat.yml.dist
+++ b/behat.yml.dist
@@ -1,4 +1,4 @@
-### Example behat.yml file for running Ushahidi tests
+## Example behat.yml file for running Ushahidi tests
 ## Replace base_url values with the url of your dev site
 ##
 default:
@@ -12,7 +12,7 @@ default:
                 - Tests\Integration\Bootstrap\KohanaContext
                 - Tests\Integration\Bootstrap\FeatureContext
                 - Tests\Integration\Bootstrap\RestContext:
-                    baseUrl: http://192.168.33.110
+                    baseUrl: http://localhost:8000
                     # proxyUrl: localhost:8888
                 - Tests\Integration\Bootstrap\PHPUnitFixtureContext
                 - Tests\Integration\Bootstrap\MinkExtendedContext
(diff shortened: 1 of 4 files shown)


behat.yml.dist
```diff
@@ -12,24 +12,24 @@ default:
                 - Tests\Integration\Bootstrap\KohanaContext
                 - Tests\Integration\Bootstrap\FeatureContext
                 - Tests\Integration\Bootstrap\RestContext:
-                    baseUrl: http://localhost:8000
+                    baseUrl: http://192.168.33.110
```
rjmackay: Can you revert this? You shouldn't need to edit this file. You can copy it to `behat.yml` and edit that instead

src/Core/Tool/Authorizer/TosAuthorizer.php
```diff
@@ -43,6 +43,13 @@ public function isAllowed(Entity $entity, $privilege)
         // These checks are run within the user context.
         $user = $this->getUser();
 
+        //if user is not actual user user_id === 0
+        if ($privilege === 'create' &&
+            !$this->isUserOwner($entity, $user)
```
rjmackay: Do you need the `isUserOwner` check?

crcommons: @rjmackay I very distinctly remember thinking through whether or not I needed it and came to the conclusion that yes, I did. Now that I look back at it, I can't see why I thought that.

rjmackay: Weird. Reading the code I think that if `isUserAndOwnerAnonymous` returns `true`, then `isUserOwner` will always return `true`.

tests/integration/tos.feature
```diff
@@ -3,7 +3,7 @@ Feature: Testing the Tos API
 
     Scenario: Create a ToS entry
         Given that I want to make a new "tos"
-        And that the request "Authorization" header is "testbasicuser"
+        And that the request "Authorization" header is "Bearer testbasicuser"
```
rjmackay: @crcommons weirdly these were wrong from the beginning :/ I'm not really sure how these tests passed with an invalid auth header.
Going to test a little more before merge

Issue #2026, opened 2017-08-24
Don't make initial admin user agree to Terms Of Service


rjmackay:
### Expected behaviour
When I create a deployment I should not be prompted to agree to terms of service, because I already agreed to them in ushahidi.io/create

### Actual behaviour
I am asked to agree to terms of service *again* on my first login

### Steps to reproduce the behaviour/error

Refs https://github.com/ushahidi/platform/issues/1772

jshorland:
@rjmackay clarifying question: do you mean that initial admin that have already created a deployment before this change goes into effect would not need to agree to the ToS?

rjmackay:
AFTER. Those who already created a deployment need to be prompted, since the terms have changed. But new deployment agree to terms during creation.

rjmackay:
Passes QA!

PR #2328, opened 2017-11-30
adding migration to add first admin to tos #2026

Angamanga:
- Adds the first admin-user to tos-table 

Test checklist:
- Create a new deployment
- Log in as admin
- [ ] Expected result: You SHOULD NOT be prompted for ToS
- Create a new user
- Log in as new user
- [ ] You SHOULD BE prompted for ToS

Fixes ushahidi/platform#2026 .

Ping @ushahidi/platform

### DIFF
--- a/migrations/20171130101357_add_first_admin_to_tos.php
+++ b/migrations/20171130101357_add_first_admin_to_tos.php
@@ -1,0 +1,43 @@
+<?php
+
+use Phinx\Migration\AbstractMigration;
+
+class AddFirstAdminToTos extends AbstractMigration
+{
+   /**
+     * Migrate Up.
+     */
+    public function up()
+    {
+        $pdo = $this->getAdapter()->getConnection();
+        $user = $pdo->query("SELECT * FROM users where id = 1")->fetch();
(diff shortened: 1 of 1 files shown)


migrations/20171130101357_add_first_admin_to_tos.php
```diff
@@ -0,0 +1,47 @@
...
+
+class AddFirstAdminToTos extends AbstractMigration
+{
+   /**
+     * Migrate Up.
+     */
+    public function up()
+    {
+        $pdo = $this->getAdapter()->getConnection();
+        $user = $pdo->query("SELECT * FROM users where id = 1")->fetch();
```
rjmackay: On ushahidi.io is the first admin reliably id 1? 

Another approach to this might be to add a `--tos` flag to the user create CLI tasks. That task is used by the cloud job queue to create the admin user on new deployments.

Angamanga: Good point, I was thinking it is the same as the admin created in the migrations but of course it is not... :/

PR #2359, opened 2017-12-12
Contact hydration

willdoran:
- Adds contacts to User object
- Hydrates User object with contacts
- Removes contacts based on permission in formatter
- Allows for searching of users by contact

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/application/classes/Ushahidi/Repository/User.php
+++ b/application/classes/Ushahidi/Repository/User.php
@@ -54,7 +54,29 @@ class Ushahidi_Repository_User extends Ushahidi_Repository implements
 	// Ushahidi_Repository
 	public function getEntity(Array $data = null)
 	{
+		if (!empty($data['id']))
+		{
+			$data += [
+				'contacts' => $this->getContacts($data['id']),
+			];
+		}
 		return new User($data);
+	}
+
+	protected function getContacts($entity_id)
+	{
+		// Unfortunately there is a circular reference created if the Contact repo is
+		// injected into the User repo to avoid this we access the table directly
+		// NOTE: This creates a hard coded dependency on the table naming for contacts
+		$query = DB::select('*')->from('contacts')
+					->where('user_id', '=', $entity_id);
+
+		$results = $query->execute($this->db);
+
+		Kohana::$log->add(Log::ERROR, print_r($entity_id, true));
+		Kohana::$log->add(Log::ERROR, print_r($results->as_array(), true));
+
+		return $results->as_array();
 	}
 
 	// CreateRepository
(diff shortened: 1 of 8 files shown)


willdoran:
If I have Post entity derived from a data source, if the Message the Post is derived from is associated with the Contact of an existing User. That User should be set as the user id of the Post.

For existing data this may require a migration akin to 

update messages set user_id = ( select user_id from contacts c where messages.contact_id = c.id) where user_id is null;
update posts set user_id = ( select user_id from messages m where m.post_id = posts.id) where user_id is null;

application/classes/Ushahidi/Repository/User.php
```diff
@@ -54,9 +54,28 @@ protected function getTable()
...
+		if (!empty($data['id']))
+		{
+			$data += [
+				'contacts' => $this->getContacts($data['id']),
+			];
+		}
 		return new User($data);
 	}
 
+	protected function getContacts($entity_id)
```
rjmackay: Hopefully we can eventually just combine contacts + users so this gets less circular

src/Core/Entity/User.php
```diff
@@ -27,6 +27,7 @@ class User extends StaticEntity
 	protected $updated;
 	protected $role;
 	protected $language;
+	protected $contacts;
```
rjmackay: Could you check what happens if someone sends an update with contacts in the payload?
We've seen previous issues where the repo's try to automatically set a user.contacts field in the db and everything breaks.

willdoran: I tested it via the client and it worked but I can add a test for it

willdoran: Ok, I updated one of the tests to ensure that including contacts doesn't cause a problem and changed the user update method to unset contacts explicitly.

application/classes/Ushahidi/Repository/User.php
```diff
@@ -114,13 +134,18 @@ public function getSearchFields()
...
 		if ($search->q)
 		{
 			$query->and_where_open();
 			$query->where('email', 'LIKE', "%" . $search->q . "%");
 			$query->or_where('realname', 'LIKE', "%" . $search->q . "%");
 			$query->and_where_close();
+
+			// Adding search contacts
+			$query->join('contacts')->on("$table.id", '=', 'contacts.user_id')
+			->or_where('contacts.contact', 'like', '%' . $search->q . '%');
```
rjmackay: @willdoran it occurs to me now that we should probably limit searching by email or contact to users with permissions, potentially you could ID a user via search despite the fact the API doesn't return their email or contact info.

willdoran: True we should add a check, I'll patch it later today.

PR #2454, opened 2018-02-02
Move config to env

rjmackay:
- Allow configuring ratelimiter and cdn through envvars

Test checklist:
- [x] Configure CDN and ratelimiter. Check they work (Going to try it on staging and see)

Ping @ushahidi/platform


rjmackay:
@tuxpiper FYI you might need to merge these changes into custom deployments to work with the updated ansible role here: https://github.com/ushahidi/ansible-platform-api-code/pull/2

tuxpiper:
Thanks for the headsup @rjmackay . I'll try to keep this for reference, for when we upgrade any of the current custom deployments in the future.

PR #2472, opened 2018-02-12
WIP - Adding Export Job to API

willdoran:
-

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


migrations/20180212204016_create_export_job_table.php
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Migrate Up.
+     */
+    public function up()
+    {
+        $this->table('export_job')
+            ->addColumn('user_id', 'integer', ['null' => false])
+            ->addColumn('entity_type', 'string', ['null' => false])
+            ->addColumn('fields', 'string', ['null' => true])
+            ->addColumn('filters', 'string', ['null' => true])
```
kinstelli: Would filters include the survey(s) being requested? I believe that having a different job request for each survey might make sense, since the field list would potentially be vastly different for each CSV.

application/classes/Ushahidi/Formatter/Post/CSV.php
```diff
@@ -71,12 +76,45 @@ protected function generateCSVRecords($records)
...
+		$extension = pathinfo($filepath, PATHINFO_EXTENSION);
+		
+		$mimeType = MimeType::detectByFileExtension($extension) ?: 'text/plain';
+		
+		$config = ['mimetype' => $mimeType];
+		
+		$this->fs->putStream($filepath, $stream, $config);
+
+		if (is_resource($stream)) {
+			fclose($stream);
```
rowasc: I think if you do this here it'll just be destroyed in each row because it uses tmpfile. Is that the idea? (sorry I'm still trying to understand everything here, this might be a dumb question)

migrations/20180212204016_create_export_job_table.php
rowasc: about "different job for each survey" @kinstelli @willdoran => that's not part of the scope and needs to be user tested before/if we want to do it. Just an FYI because it's going to be more work for something we don't know for sure we want or how we want it

src/Core/Tool/Authorizer/ExportJobAuthorizer.php
```diff
@@ -0,0 +1,71 @@
...
+use Ushahidi\Core\Traits\OwnerAccess;
+use Ushahidi\Core\Traits\UserContext;
+use Ushahidi\Core\Entity\Permission;
+use Ushahidi\Core\Traits\PrivAccess;
+use Ushahidi\Core\Traits\PrivateDeployment;
+use Ushahidi\Core\Tool\Permissions\AclTrait;
+
+class ExportJobAuthorizer implements Authorizer
+{
+	// The access checks are run under the context of a specific user
```
rowasc: Love the comments about why we have each auth trait here , even more after the user issue

src/Core/Usecase/Export/Job/CreateJob.php
```diff
@@ -0,0 +1,30 @@
...
+
+class CreateJob extends CreateUsecase
+{
+    protected function getEntity()
+	{
+		$entity = parent::getEntity();
+
+		// Add user id if this is not provided
+		// TODO: throw this away
+		if (empty($entity->user_id) && $this->auth->getUserId()) {
```
rowasc: ok made me curious: why this is here/ why it'd be thrown out? thanks

application/classes/Ushahidi/Formatter/Post/CSV.php
```diff
@@ -43,23 +54,17 @@ protected function generateCSVRecords($records)
 		 */
 		$headingColumns = $this->getCSVHeading($records);
```
kinstelli: Are we still thinking of making CSV heading an optional feature, if we're going to do partial CSV exports and later concat them together? Apologies if this exists somewhere else and I don't see it.

migrations/20180212204016_create_export_job_table.php
willdoran: @rowasc I think separating by surveys is in, no? Because otherwise the surveys will still look terrible, It's relatively straightforward to do. The filters would contain the surveys, I can just pull them out and split them up, though obviously it would spawn a large number of files if there was a large number of surveys....hmmmm....actually may be for this moment we should just output a single file then split it BUT we will need to split them up.

I'll see if there is an easy way to do this tomorrow.

rowasc: No, it wasn't in/scoped for this cycle because we need user testing. I also can't find an issue saying it should be chopped by survey. @willdoran 
It was considered something to do in a next step, because we want to know for sure how to do it/what the user expects.

application/classes/Ushahidi/Formatter/Post/CSV.php
```diff
@@ -10,19 +10,30 @@
...
+	protected $tmpfname;
 
 	// Formatter
 	public function __invoke($records)
 	{
-		$this->generateCSVRecords($records);
+		return $this->generateCSVRecords($records);
+	}
+
+	public function setFS($fs)
```
kinstelli: Total nitpick question: can we also call this setFilesystem() to parallel the other methods? And/or name it something to clarify that we're also setting a tempfname in here, if we don't want to move that into writeStreamToFS()?

application/classes/Controller/Api/Exports/External/Cli.php
```diff
@@ -0,0 +1,100 @@
...
+			$shared_secret = getenv('PLATFORM_SHARED_SECRET');
+			$signer = new Signer($shared_secret);
+			$fullURL = URL::site(Request::detect_uri(), TRUE) . URL::query();
+
+			return $signer->validate($signature, $fullURL, $data);
+		}
+		return false;
+	}
+
+	// public function before()
```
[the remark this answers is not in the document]
willdoran: sorry this shouldn't be commented out at all my bad!

application/classes/Controller/Api/Exports/External/Cli.php
```diff
@@ -0,0 +1,100 @@
...
+	// 		throw HTTP_Exception::factory(403, 'Forbidden');
+	// 	}
+	// }
+
+	protected function _is_auth_required()
+	{
+		return false;
+	}
+
+	public function action_get_index($command = 'export')
```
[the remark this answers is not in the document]
willdoran: fixed

application/classes/Controller/Api/Exports/External/Cli.php
```diff
@@ -0,0 +1,100 @@
...
+		$output = new BufferedOutput();
+
+		
+		// Run the command
+		$command->run($input, $output);
+
+		//Retrieve results
+		$export_results = json_decode($output->fetch());
+
+		$this->_response_payload = [
```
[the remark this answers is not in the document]
willdoran: Sorry, this is a slight misnomer I'll rename, what is actually returned is the url of the file as a json formatted string, this is because the CLI tool prints to the command line a string of serialised json which we then read and then re-encode as json.

Fixed by changing name and improving with a comment.

application/classes/Controller/Api/Exports/External/Count.php
```diff
@@ -0,0 +1,70 @@
...
+	{
+		return 'export_jobs';
+	}
+
+	protected function _is_auth_required()
+	{
+		return false;
+	}
+
+	public function checkApiKey($data)
```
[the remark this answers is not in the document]
willdoran: Yep it's in a few places, I'll move it out to a trait thanks!

application/classes/Ushahidi/Console/PostExporter.php
```diff
@@ -62,33 +84,48 @@ protected function executeList(InputInterface $input, OutputInterface $output)
...
+			// Load the export job
+			$job = $this->exportJobRepository->get($job_id);
+			
+			// Merge the export job filters with the base filters
+			if ($job->filters) {
+				$filters = array_merge($filters, $job->filters);
+			}
+			
+			// Set the fields that should be included if set
+			if ($job->fields) {
```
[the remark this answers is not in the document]
willdoran: done

application/classes/Ushahidi/Formatter/Post/CSV.php
```diff
@@ -10,19 +10,30 @@
...
+	{
+		$this->tmpfname = "/tmp/" . strtolower(uniqid() . '-' . strftime('%G-%m-%d') . '.csv');
```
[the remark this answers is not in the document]
willdoran: will fix!

application/classes/Ushahidi/Formatter/Post/CSV.php
willdoran: No it's not dumb, it's annoying how this works let me try to explain.

We write to a temp file managed by php, once we have completed writing to that file, we take it and write it to the CDN(rackspace, aws). The reason we have to do this is because you can not append to a file via the flyssytem lib on either rackspace or aws. Aws actually supports that but the lib doesn't.

src/Core/Usecase/Export/Job/CreateJob.php
willdoran: Not sure will recheck

application/classes/Ushahidi/Formatter/Post/CSV.php
willdoran: Yes, I can add a means to do that, will do

application/classes/Ushahidi/Formatter/Post/CSV.php
willdoran: yes, my bad!
fixed

application/classes/Ushahidi/Formatter/Post/CSV.php
```diff
@@ -71,12 +76,45 @@ protected function generateCSVRecords($records)
...
+		}
+
+		return $this->writeStreamToFS($stream);
+	}
+
+
+	private function writeStreamToFS($stream)
+	{
+
+		$filepath = implode('/', [
```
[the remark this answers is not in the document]
willdoran: done

application/classes/Ushahidi/Formatter/Post/CSV.php
willdoran: done

application/classes/Controller/Api/Exports/External/Cli.php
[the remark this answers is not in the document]
rowasc: thanks ! that makes sense

PR #2476, opened 2018-02-14
Fixing missing user check

willdoran:
-

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/src/Core/Tool/Authorizer/PostAuthorizer.php
+++ b/src/Core/Tool/Authorizer/PostAuthorizer.php
@@ -125,7 +125,7 @@ class PostAuthorizer implements Authorizer
             return true;
         }
 
-        if ($user->getId() and in_array($privilege, ['search'])) {
+        if (in_array($privilege, ['search'])) {
             return true;
         }
 
--- a/src/Core/Tool/Authorizer/PostAuthorizer.php
+++ b/src/Core/Tool/Authorizer/PostAuthorizer.php
@@ -121,7 +121,11 @@ class PostAuthorizer implements Authorizer
         }
 
         // All users are allowed to create and search posts.
-        if (in_array($privilege, ['create', 'search'])) {
+        if (in_array($privilege, ['create'])) {
+            return true;
+        }
+
+        if ($user->getId() and in_array($privilege, ['search'])) {
             return true;
         }
 
--- a/src/Core/Tool/Authorizer/PostAuthorizer.php
+++ b/src/Core/Tool/Authorizer/PostAuthorizer.php
@@ -121,11 +121,7 @@ class PostAuthorizer implements Authorizer
         }
 
         // All users are allowed to create and search posts.
-        if (in_array($privilege, ['create'])) {
-            return true;
-        }
-
-        if (in_array($privilege, ['search'])) {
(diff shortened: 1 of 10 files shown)


src/Core/Tool/Authorizer/PostAuthorizer.php
```diff
@@ -121,7 +121,11 @@ public function isAllowed(Entity $entity, $privilege)
         }
 
         // All users are allowed to create and search posts.
-        if (in_array($privilege, ['create', 'search'])) {
+        if (in_array($privilege, ['create'])) {
+            return true;
+        }
+
+        if (in_array($privilege, ['search'])) {
```
rowasc: wait. Why is this divided but doing the same thing?

src/Core/Tool/Authorizer/TagAuthorizer.php
```diff
@@ -78,7 +78,7 @@ public function isAllowed(Entity $entity, $privilege)
 			return true;
 		}
 
-		if ($privilege === 'search') {
+		if ($user->getId() and $privilege === 'search') {
```
rowasc: for tag I'm realizing we need the search. We just need to avoid leaking private tags

rowasc: (we need to see tags as annon)

src/Core/Tool/Authorizer/UserAuthorizer.php
```diff
@@ -86,7 +86,7 @@ public function isAllowed(Entity $entity, $privilege)
 		}
 
 		// Regular user can always read
-		if (in_array($privilege, ['read', 'search'])) {
+		if ($user->getId() and in_array($privilege, ['read', 'search'])) {
```
rowasc: Shouldn't this only use the 		if ($this->acl->hasPermission($user, Permission::MANAGE_USERS)) for search ?

rowasc: and isUserOwner for read

src/Core/Tool/Authorizer/SetAuthorizer.php
```diff
@@ -92,7 +92,7 @@ public function isAllowed(Entity $entity, $privilege)
 		}
 
 		// Finally, all users can search sets
-		if ($privilege === 'search') {
+		if ($user->getId() and $privilege === 'search') {
```
rowasc: not logged in users should be able to see lists of collections (sets), but only those that are public

rowasc: the same applies to saved searches endpoint. That seems to be working OK in develop (filtered at another stage? )

src/Core/Tool/Authorizer/UserAuthorizer.php
```diff
@@ -85,11 +85,6 @@ public function isAllowed(Entity $entity, $privilege)
 			return true;
 		}
 
-		// Regular user can always read
```
rowasc: in line 84, where they are checking 		if ($this->isUserSelf($entity) && in_array($privilege, ['update', 'read_full',])) {
we should add 'read' to the list. Users should see their own data

PR #2481, opened 2018-02-16
API should not support adding child categories with different permission than parent

rowasc:
-

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


application/classes/Ushahidi/Repository/Tag.php
```diff
@@ -162,8 +160,30 @@ public function delete(Entity $entity)
 	public function deleteTag($id)
 	{
 		// Remove tag from attribute options
-		$this->removeTagFromAttributeOptions($entity->id);
-
+		$this->removeTagFromAttributeOptions($id);
```
rowasc: I'm really not sure why this was using $entity->id since there was not entity to use. I could revert it .. it's probably not being used . :/

rowasc:
@willdoran can you take a look at this PR? I'm not sure TBH, I think isRoleValid could move to a trait and have a better name, (besides the implementation itself) ??

application/classes/Ushahidi/Repository/Tag.php
willdoran: Hmmm yeah this must not have been used anymore

PR #2486, opened 2018-02-20
fix(category permissions) role must match for child and parent

rowasc:
- Fixes permission check on tag create/update

Test checklist:
- [ ] Follow checklist in platform/#2408

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/application/classes/Ushahidi/Repository/Tag.php
+++ b/application/classes/Ushahidi/Repository/Tag.php
@@ -177,10 +177,9 @@ class Ushahidi_Repository_Tag extends Ushahidi_Repository implements
 		$isChild = $fullData['parent_id'];
 		$hasRole = !!$fullData['role'];
 		$parent = $fullData['parent_id'] && $isChild && $this->getEntity(['id' => $fullData['parent_id']]);
-
 		if ($hasRole && $isChild && $parent) {
 			$parent = $this->selectOne(['id' => $fullData['parent_id']]);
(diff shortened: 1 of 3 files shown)


application/classes/Ushahidi/Repository/Tag.php
```diff
@@ -174,12 +174,13 @@ public function deleteTag($id)
 	public function isRoleValid(Validation $validation, $fullData)
 	{
 		$valid = true;
-		$isChild = $fullData['parent_id'];
```
rowasc: @willdoran had to refactor this. 
1. I had a dumb bug checking $valid incorrectly, I suspect I  had tested and then changed something. Aaaanyway. FIxed now
2. I realized I needed to use entities otherwise we need to wrangle the fulldata role formats and we really shouldn't need to do that! 
3. Fixed tests to actually, you know.. test the correct thing.

rowasc:
@willdoran yep, that's correct . 

About the warning: IF that happens (when editing children we don't show a way to edit the role, it is always the parent's role) we would send back a message with a 422 that says "Role must match the parent category"  (https://github.com/ushahidi/platform/commit/70e91c9803b310273a70f3581b32cda969039c17#diff-05eed86ed0c3b7db17b92d1d4fbf735a) 

This was mostly for preventing any type of client bugs with role settings from messing up the database and for preventing postman/ cli based post creation from messing up the category permissions.

PR #2487, opened 2018-02-20
Post csv export formatter

rowasc:
- Adds media formatter (makes images be URLs instead of IDs) 
- Adds fix for categories (should be in 1 column, comma separated)
- Adds fix for Date (should be Date, not DateTime)

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


application/classes/Ushahidi/Formatter/Post/CSV.php
```diff
@@ -81,15 +83,37 @@ protected function generateCSVRecords($records)
 
 	private function getValueFromRecord($record, $keyParam){
 		$return = '';
-		$keySet = explode('.', $keyParam); //contains key + index of the key, if any
```
rowasc: ugh . this is awful. Looking to figure out something less terrible if I can.

application/classes/Ushahidi/Formatter/Post/CSV.php
```diff
@@ -81,15 +83,37 @@ protected function generateCSVRecords($records)
...
+
 		$headingKey = $keySet[0];
 		$key = isset($keySet[1]) ? $keySet[1] : null;
-		$recordValue = isset ($record['attributes']) && isset($record['attributes'][$headingKey])? $record['values']: $record;
-		if($key === 'lat' || $key === 'lon'){
+		$recordAttributes = isset($record['attributes'][$headingKey]) ? $record['attributes'][$headingKey] : null;
+		$recordValue = isset ($record['attributes']) && $recordAttributes? $record['values']: $record;
+		$isDateField = $recordAttributes['input'] === 'date' && $recordAttributes['type'] === 'datetime';
+
+		if ($isDateField) {
```
rowasc: not sure if the right place..  but I think there's not much to this. Basically, it just needs to be switched to y-m-d

application/classes/Ushahidi/Repository/Post.php
```diff
@@ -135,6 +135,7 @@ public function getEntity(Array $data = null)
 			$data += [
 				'values' => $this->getPostValues($data['id']),
 				// Continued for legacy
+				// 'media' => $this->getMediaForPost($data['id']),
```
rowasc: need to remove this reference

application/classes/Ushahidi/Repository/Post/Media.php
```diff
@@ -14,9 +14,56 @@
 
 class Ushahidi_Repository_Post_Media extends Ushahidi_Repository_Post_Value
 {
+
+	// Ushahidi_Repository
+	public function getEntity(Array $data = null)
+	{
+		$data['value'] = $data['o_filename'];
```
rowasc: forcing 'value' to be filename here could have consequences in client. Need to check

rowasc: might be best to let it be an id and change the getAllForPost

rowasc: the problem with this is that we use the id for getting media from the server so we can't do this except for csv

application/classes/Ushahidi/Repository/Post/Export.php
```diff
@@ -30,7 +30,7 @@ public function retrieveColumnNameData($data) {
 	foreach ($data['values'] as $key => $val)
     {
         $attribute = $this->form_attribute_repo->getByKey($key);
-        $attributes[$key] = ['label' => $attribute->label, 'priority'=> $attribute->priority, 'stage' => $attribute->form_stage_id, 'type'=> $attribute->type, 'form_id'=> $data['form_id']];
+        $attributes[$key] = ['label' => $attribute->label, 'input' => $attribute->input, 'priority'=> $attribute->priority, 'stage' => $attribute->form_stage_id, 'type'=> $attribute->type, 'form_id'=> $data['form_id']];
```
rowasc: this is a bit weird.. but I couldn't find another way to know if it's actually a date or datetime other than using 'input' . The type only says datetime

application/classes/Ushahidi/Repository.php
```diff
@@ -115,7 +115,7 @@ public function setSearchParams(SearchData $search)
 	}
 
 	// SearchRepository
-	public function getSearchResults()
+	public function getSearchResults($csv = false)
```
rowasc: $csv flag should be gone

src/Core/Usecase/Post/Export.php
```diff
@@ -49,7 +49,7 @@ public function interact()
 		$this->repo->setSearchParams($search);
 
 		// ... get the results of the search
-		$results = $this->repo->getSearchResults();
+		$results = $this->repo->getSearchResults(true);
```
rowasc: true not needed here.

application/classes/Ushahidi/Formatter/Post/CSV.php
rowasc: I think we should have a CSV format field that drives how they are shown in a csv.  Right now it's ridiculous.

rowasc: ok my comment is outdated. This is still awful BUT now I removed the extra awfulness that I introduced with the single column arrays (categories in a single column instead of many, for instance ) and made it depend on an array of pre-defined formats by key/value pair  that depend on field type. It's still far from nice, but it's a step in the right direction I believe

rowasc:
@willdoran can you take a look at this? this is where I landed after our call yesterday. I tried not to touch the previous code that much since it's super messy, but some changes were obviously necessary.

rowasc:
@kinstelli LMK if you can take a look too. Would really appreciate a fresh look at it .

rowasc:
[PR-csvwithallfields.csv.txt](https://github.com/ushahidi/platform/files/1749631/PR-csvwithallfields.csv.txt)
@willdoran @kinstelli uploading a csv with the . current status

rowasc:
[5a8f3bb241db4-2018-02-22.csv.txt](https://github.com/ushahidi/platform/files/1749699/5a8f3bb241db4-2018-02-22.csv.txt)
Updated to include (UTC) as agreed in the format ticket

rowasc:
@willdoran yea, I had to remove it because of that. Thanks for the heads up.

rowasc:
>Looks good, the media thing is demonstrating that this whole post values get all approach is limited. We should be using proper orm.

+100 to this. Yes. The post values approach is NOT working.

application/classes/Ushahidi/Formatter/Post/CSV.php
rowasc: removing outdated comment

PR #2497, opened 2018-02-24
Override auth exportcli

willdoran:
-

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


src/App/Init.php
```diff
@@ -114,3 +114,17 @@
 // Rate limiter violation handler
 $di->setter['BehEh\Flaps\Flap']['setViolationHandler'] =
 	$di->lazyNew('Ushahidi\App\ThrottlingViolationHandler');
+
+/**
+ * @TODO
+ * This should only be for our csv exporter
+ */
+if (php_sapi_name() !== "cli") {
```
willdoran: @rowasc Does this behave as expected when the cli is called from the API cli route? I'm not sure whether or not that enters a totally separate process.

rowasc:
@willdoran right now the only test failing is the one where it complains about /media . Do you know how to make it not complain?

Issue #2593, opened 2018-03-13
Importing CSV doesn't allow assigning multiple categories to a post


tuxpiper:
### Expected behaviour
If a post has to be related to multiple categories, it should be possible to reflect this in a CSV file for importing.

### Actual behaviour
There's currently no CSV column formatting or mapping that platform can correctly parse to understand that the post needs to be assigned to multiple categories.

Only one category can be assigned to each imported post.

#### Is is ready for development when
- [ ] Has a description that includes purpose of this particular feature/improvement (how it will help the user and for which use cases)
- [ ] Can be delivered within a sprint (make sure issues are small pieces of work, not huge epics)
- [ ] Has all designs and UI assets available
- [ ] Has all (probable) dependencies identified
- [ ] Is estimated by the team
- [ ] Has acceptance criteria defined (do we want to do this before we send to dev? might help us identify dependencies/holes earlier?)

##### Where
- [x] [Local setup with Vagrant ](https://www.ushahidi.com/support/install-ushahidi#installing-for-development)
- [x] [Local setup from platform-release ](https://www.ushahidi.com/support/install-ushahidi#installing-the-latest-release)
- [x] Ushahidi.io / SaaS solution 
- [x] Ushahidi's QA environment
- [x] Other (explain):

Aha! Link: https://ushahiditeam.aha.io/features/PROD-411

tuxpiper:
I believe this will affect importing any data types that can take multiple values: multiple select lists and checkbox lists .

The essence of the problem is this:

While the CSV import code in the API assumes that any multiple values will be spread across several columns in the CSV ... the UI doesn't have any way for the user to specify a mapping of multiple CSV columns to a single form value.

tuxpiper:
Closing without merging due to the whole hackiness of it

tuxpiper:
Convo from slack:

@rjmackay dixit:
> options that [would accept comma separated values in a single column, i.e. ("1,2,3")] we can do without too much rewriting, we could:
> 1. explode category values in the post repo. That'd allow us to do this for all category fields.
> 2. allow specifying that a value should be exploded as part of the transformer `map` ie. a map value of `fieldkey[]` instead of just `fieldkey`. We'd need to add a corresponding toggle in the UI ie. `[ ] cell's contain multiple values` when setting up mappings.

@tuxpiper dixit:
> I think for a quick fix, option 1 is good. Maybe just needs to document it to make it more transparent to the end-user, but it would make enough sense.
[...]

PR #2594, opened 2018-03-13
ushahidi/platform#2593 hack to parse CSV columns as an array

tuxpiper:
- introduces a mechanism for performing transformation to the values contained in a CSV file, on the fly, while doing the import
- the transformation is triggered for CSV columns that are named using the following syntax "[name]->{[transformation]}"
- implements the "explodeCommas" transformation for converting values such as "1,2,3" to array("1","2","3")

Test checklist:
- [x] in a deployment with the default form
- [x] create categories C1,C2,...,C6
- [x] create CSV file with columns "title", "description" and "categories->{explodeCommas}"
- [x] fill rows in the csv, in the third column fill values like "1,2" , "2,4,5" etc
- [x] import the CSV into the deployment by mapping categories to the last column

Helps with ushahidi/platform#2593 .

Ping @ushahidi/platform

### DIFF
--- a/application/classes/Controller/Api/CSV/Import.php
+++ b/application/classes/Controller/Api/CSV/Import.php
@@ -52,6 +52,7 @@ class Controller_API_CSV_Import extends Ushahidi_Rest {
 		$records = $reader->process($file);
 
 		// Set map and fixed values for transformer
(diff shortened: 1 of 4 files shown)


tuxpiper:
Closing without merging due to the whole hackiness of it

PR #2596, opened 2018-03-13
Master update to develop@20180212

tuxpiper:
This is a huge chunk of updates that apparently have been due for a while.

We are somewhat forced to apply them now, since the corresponding changes to platform-client have somehow made it master .. and compatibility has been broken between master branches


rjmackay:
@tuxpiper have you check if they deploys to production OK? At least part of this was changes to how we handled CDN/Media/Ratelimiter config. It might require some tweaks to the ansible vars.

tuxpiper:
@rjmackay I haven't done that yet, I'll get to it before pushing. Would that be the only risk of pushing this to production? Are we confident everything else is fine?

rjmackay:
Honestly, no not super confident. The COMRADES changes were fairly wide reaching *but* they had to merge eventually. However the migrations mostly look fairly safe so I suspect we can rollback if anything goes badly wrong.
I'd be happy to move forward and get the master branches in sync.

PR #2605, opened 2018-03-16
feat(DataSources): Reimplement inbound field mappings

rjmackay:
- Reintroduce ability to map data source to form

Test checklist:
- [ ] TBD

Ping @ushahidi/platform

### DIFF
--- a/src/App/DataSource/Email/Email.php
+++ b/src/App/DataSource/Email/Email.php
@@ -15,6 +15,7 @@ use Ushahidi\App\DataSource\IncomingAPIDataSource;
 use Ushahidi\App\DataSource\OutgoingAPIDataSource;
 use Ushahidi\App\DataSource\Message\Type as MessageType;
 use Ushahidi\App\DataSource\Message\Status as MessageStatus;
+use Ushahidi\App\DataSource\Concerns\MapsInboundFields;
 use Ushahidi\Core\Entity\MessageRepository;
 use Illuminate\Contracts\Mail\Mailer;
 use Ushahidi\Core\Entity\Contact;
@@ -22,6 +23,7 @@ use Log;
 
 class Email implements IncomingAPIDataSource, OutgoingAPIDataSource
 {
+	use MapsInboundFields;
 
 	protected $config;
 	protected $mailer;
@@ -109,6 +111,17 @@ class Email implements IncomingAPIDataSource, OutgoingAPIDataSource
 				'rules' => array('required')
 			)
 		);
+	}
+
+	public function getInboundFields()
+	{
+		return [
+			'From' => 'text',
+			'Subject' => 'text',
+			'To' => 'text',
+			'Date' => 'datetime',
+			'Message' => 'text'
+		];
 	}
 
 	/**
--- a/src/Core/Usecase/Message/ReceiveMessage.php
+++ b/src/Core/Usecase/Message/ReceiveMessage.php
@@ -173,10 +173,13 @@ class ReceiveMessage extends CreateUsecase
 		$content = $message->message;
 
 		if ($message->additional_data) {
-			if (isset($message->additional_data['form_id'])) {
-				$form_id = $message->additional_data['form_id'];
+			// @todo inject
+			$source = app('datasources')->getSource($message->data_source);
+			$formId = $source->getInboundFormId();
+
+			if ($formId) {
 				// Check provider fields for form attribute mapping
-				$inbound_fields = $message->additional_data['inbound_fields'];
+				$inbound_fields = $source->getInboundFieldMappings();
 
 				if (isset($this->payload['title']) && isset($inbound_fields['Title'])) {
 						$values[$inbound_fields['Title']] = array($this->payload['title']);
@@ -213,6 +216,7 @@ class ReceiveMessage extends CreateUsecase
 					}
 				}
 			}
+
 			// Pull locations from extra metadata
 			$values['message_location'] = [];
 			if (isset($message->additional_data['location'])) {
--- a/src/App/DataSource/Concerns/MapsInboundFields.php
+++ b/src/App/DataSource/Concerns/MapsInboundFields.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace Ushahidi\App\DataSource\Concerns;
+
+trait MapsInboundFields
+{
+	public function getInboundFormId()
+	{
+		return isset($this->config['form_id']) ? $this->config['form_id'] : false;
+	}
+
+	public function getInboundFieldMappings()
+	{
+		return isset($this->config['inbound_fields']) ? $this->config['inbound_fields'] : false;
+	}
+}
--- a/src/App/DataSource/DataSource.php
+++ b/src/App/DataSource/DataSource.php
@@ -23,4 +23,7 @@ interface DataSource
 	public function getId();
 	public function getServices();
 	public function getOptions();
+	public function getInboundFields();
+	public function getInboundFormId();
+	public function getInboundFieldMappings();
 }
--- a/src/App/DataSource/FrontlineSMS/FrontlineSMS.php
+++ b/src/App/DataSource/FrontlineSMS/FrontlineSMS.php
@@ -15,11 +15,13 @@ use Ushahidi\App\DataSource\CallbackDataSource;
 use Ushahidi\App\DataSource\OutgoingAPIDataSource;
 use Ushahidi\App\DataSource\Message\Type as MessageType;
 use Ushahidi\App\DataSource\Message\Status as MessageStatus;
+use Ushahidi\App\DataSource\Concerns\MapsInboundFields;
 use Ushahidi\Core\Entity\Contact;
 use Log;
 
 class FrontlineSMS implements CallbackDataSource, OutgoingAPIDataSource
 {
+	use MapsInboundFields;
 
 	protected $config;
 
@@ -64,6 +66,15 @@ class FrontlineSMS implements CallbackDataSource, OutgoingAPIDataSource
 				'rules' => array('required')
 			)
 		);
+	}
+
+	public function getInboundFields()
+	{
+		return [
+			'From' => 'text',
+			'To' => 'text',
+			'Message' => 'text'
+		];
 	}
 
 	/**
--- a/src/App/DataSource/Nexmo/Nexmo.php
+++ b/src/App/DataSource/Nexmo/Nexmo.php
@@ -15,11 +15,13 @@ use Ushahidi\App\DataSource\CallbackDataSource;
 use Ushahidi\App\DataSource\OutgoingAPIDataSource;
 use Ushahidi\App\DataSource\Message\Type as MessageType;
 use Ushahidi\App\DataSource\Message\Status as MessageStatus;
+use Ushahidi\App\DataSource\Concerns\MapsInboundFields;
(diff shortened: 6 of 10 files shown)


src/Core/Usecase/Message/ReceiveMessage.php
```diff
@@ -173,10 +173,13 @@ protected function createPost(Entity $message)
 		$content = $message->message;
 
 		if ($message->additional_data) {
-			if (isset($message->additional_data['form_id'])) {
-				$form_id = $message->additional_data['form_id'];
+			// @todo inject
+			$source = app('datasources')->getSource($message->data_source);
+			$formId = $source->getInboundFormId();
```
rjmackay: @willdoran I've moved this out of the message data since its really tied to the data source. 
I think this makes sense?

src/App/DataSource/Email/Email.php
```diff
@@ -111,6 +113,17 @@ public function getOptions()
 		);
 	}
 
+	public function getInboundFields()
+	{
+		return [
+			'From' => 'text',
```
rjmackay: I'm not 100% clear why these keys are capitalized rather than say matching the key on the message payload.

I wonder if these were more like the `options` definition they might be more extendable ie
```
'somefield' => [
  'type' => 'text',
  'description' => 'Some field',
]
```

willdoran: No idea, they were originally capitalised. Perhaps it was related to earlier incarnations of the message system when it was first built.

src/Core/Usecase/Message/ReceiveMessage.php
willdoran: If it still works, sure.

rjmackay:
@rowasc @willdoran so I refactored this to make it easier to test, however its now a much bigger PR.
It takes us a step towards using Lumen's events instead of our own. I can walk through it if thats helpful.

app/Providers/AppServiceProvider.php
```diff
@@ -49,6 +49,26 @@ public function registerServicesFromAura()
             return service('repository.message');
         });
 
+        $this->app->singleton(\Ushahidi\Core\Entity\ContactRepository::class, function ($app) {
```
rjmackay: Not sure what the sustainable path is for this going forward, but for now this works

src/App/DataSource/Email/Email.php
```diff
@@ -111,6 +113,15 @@ public function getOptions()
 		);
 	}
 
+	public function getInboundFields()
```
rjmackay: This implementation of inbound field mappings drops the ability to map From and To values to post fields.
It was going to add a bit more mess to the code, but also I'm not convinced its a great idea to be able to map PII into the post fields.

@willdoran will losing this break anything that was using in COMRADES?

src/App/Listener/CreatePostFromMessage.php
```diff
@@ -0,0 +1,123 @@
+<?php
+namespace Ushahidi\App\Listener;
+
+use Ushahidi\Core\Entity\PostRepository;
+use Ushahidi\Core\Entity\Message;
+use Ushahidi\Core\Entity\MessageRepository;
+use Ushahidi\Core\Entity\TargetedSurveyStateRepository;
+
+class CreatePostFromMessage
```
rjmackay: This is where creating new posts from messages and any field mapping happens now

src/App/Listener/HandleTargetedSurveyResponse.php
```diff
@@ -0,0 +1,143 @@
+<?php
+namespace Ushahidi\App\Listener;
+
+use Ushahidi\Core\Entity\Message;
+use Ushahidi\Core\Entity\MessageRepository;
+use Ushahidi\Core\Entity\TargetedSurveyState;
+use Ushahidi\Core\Entity\TargetedSurveyStateRepository;
+use Ushahidi\Core\Entity\FormAttributeRepository;
+
+class HandleTargetedSurveyResponse
```
rjmackay: And this is where all the targeted survey wrangling happens now

src/App/Listener/HandleTargetedSurveyResponse.php
```diff
@@ -0,0 +1,143 @@
...
+     *
+     * @return void
+     */
+    public function handle($incomingMessageId, $incomingMessage, $inbound_form_id, $inbound_fields)
+    {
+        // \Log::info('Running HandleTargetedSurveyResponse', func_get_args());
+
+        // check if contact is part of an open targeted_survey
+        if (!$this->targetedSurveyStateRepo
+            ->isContactInActiveTargetedSurveyAndReceivedMessage($incomingMessage->contact_id)
```
rjmackay: This was previously done in a 3 stage check where we:
1. checked if the contact was in a survey
2. Loaded the survey state
3. checked if there was an outbound message.

I moved that all into a single SQL query in the repo.

src/App/Listener/HandleTargetedSurveyResponse.php
```diff
@@ -0,0 +1,143 @@
...
+            );
+
+            // If this for some unknown reason fails, log it
+            if (!$outgoingMessageId) {
+                \Log::error(
+                    'Could not create new outgoing message',
+                    compact('contact_id')
+                );
+
+                // Return the incoming message as per usual
```
rjmackay: oops this comment doesn't make sense now

src/App/Repository/ContactRepository.php
```diff
@@ -130,32 +130,6 @@ public function getByContact($contact, $type)
 		return $this->getEntity($this->selectOne(compact('contact', 'type')));
 	}
 
-	/**
-	 * @param string $contact_id
-	 * @return bool
-	 */
-	public function isInActiveTargetedSurvey($contact_id)
```
rjmackay: Move to the targeted survey state repo

src/App/Repository/TargetedSurveyStateRepository.php
```diff
@@ -70,4 +72,31 @@ public function setStatusByFormId($form_id, $status)
...
+			->where('targeted_survey_state.contact_id', '=', $contact_id)
+			->and_where(
+                'survey_status',
+                'IN',
+				[
+					Entity\TargetedSurveyState::PENDING_RESPONSE,
+					Entity\TargetedSurveyState::RECEIVED_RESPONSE
+				]
+			)
+			->and_where(
```
rjmackay: Added this check to avoid multiple queries to figure this out.

src/Core/Entity/ContactRepository.php
```diff
@@ -13,10 +13,13 @@
...
 use Ushahidi\Core\Entity\Repository\EntityGet;
 use Ushahidi\Core\Entity\Repository\EntityExists;
+use Ushahidi\Core\Usecase\CreateRepository;
+use Ushahidi\Core\Usecase\UpdateRepository;
 
 interface ContactRepository extends
 	EntityGet,
-	EntityExists
+	EntityExists,
+	CreateRepository
```
rjmackay: I've added this to a few repos. It much easier to mock for tests is everything goes into a single interface. Hopefully it'll make DI better in future too

src/Core/Entity/Message.php
```diff
@@ -61,7 +61,7 @@ protected function getDefinition()
 			'user_id'    => 'int',
 			'title'      => 'string',
 			'message'    => 'string',
-			'datetime'   => '*timestamp',
+			'datetime'   => '*date',
```
rjmackay: I haven't completely verified this but pretty sure it should work. It matches the date handling on `Post` for the `post_date` field

src/Core/Entity/TargetedSurveyStateRepository.php
```diff
@@ -13,15 +13,23 @@
...
 	 * @param string  $contact
 	 * @return boolean
 	 */
 	public function getActiveByContactId($contact_id);
+
+    /**
+	 * @param string  $contact
+	 * @return boolean
+	 */
+	public function isContactInActiveTargetedSurveyAndReceivedMessage($contact);
```
rjmackay: This method name is getting a bit ridiculous

src/Core/Traits/Events/DispatchesEvents.php
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @author     Ushahidi Team <team@ushahidi.com>
+ * @package    Ushahidi\Application
+ * @copyright  2014 Ushahidi
+ * @license    https://www.gnu.org/licenses/agpl-3.0.html GNU Affero General Public License Version 3 (AGPL3)
+ */
+
+namespace Ushahidi\Core\Traits\Events;
+
+use Illuminate\Contracts\Events\Dispatcher;
```
rjmackay: So this creates an explicit dependency in our Core code on an external lib. It probably breaks some fancy SOLID rules but I couldn't bring myself to add more weird  misdirection to our code.

src/Core/Usecase/Message/ReceiveMessage.php
```diff
@@ -23,36 +23,18 @@
 use Ushahidi\Core\Tool\Validator;
 use Ushahidi\Core\Usecase\CreateUsecase;
 use Ushahidi\Core\Usecase\CreateRepository;
+use Ushahidi\Core\Traits\Events\DispatchesEvents;
 
 use Ushahidi\Core\Exception\ValidatorException;
 
 class ReceiveMessage extends CreateUsecase
```
rjmackay: So this is a pretty straightforward usecase now. It looks for a contact, creates one if it didn't exists, then saves the message.

tests/unit/App/Listener/CreatePostFromMessageTest.php
```diff
@@ -0,0 +1,162 @@
...
+use Ushahidi\Core\Entity\PostRepository;
+use Ushahidi\Core\Entity\TargetedSurveyStateRepository;
+use Tests\TestCase;
+use Mockery as M;
+
+/**
+ * @backupGlobals disabled
+ * @preserveGlobalState disabled
+ */
+class CreatePostFromMessageTest extends TestCase
```
rjmackay: So this actually gets tests now

tests/unit/App/Listener/HandleTargetedSurveyResponseTest.php
```diff
@@ -0,0 +1,308 @@
...
+use Ushahidi\Core\Entity\FormAttributeRepository;
+
+use Tests\TestCase;
+use Mockery as M;
+
+/**
+ * @backupGlobals disabled
+ * @preserveGlobalState disabled
+ */
+class HandleTargetedSurveyResponseTest extends TestCase
```
rjmackay: And this gets tests too

tests/unit/Core/Usecase/ReceiveMessageTest.php
```diff
@@ -0,0 +1,127 @@
...
+use Ushahidi\Core\Entity\FormAttributeRepository;
+
+use Tests\TestCase;
+use Mockery as M;
+
+/**
+ * @backupGlobals disabled
+ * @preserveGlobalState disabled
+ */
+class ReceiveMessageTest extends TestCase
```
rjmackay: This test is a bit weird but I really wanted something that tested the ReceiveMessage usecase *with* the dispatcher and the listeners to verify I actually wired it up right. 
It turned out I didn't so I fixed that too :)

Issue #2599, opened 2018-03-13
Add API Endpoint for country codes


crcommons:

crcommons:
[...]

PR #2614, opened 2018-03-22
Endpoint for country-codes

Angamanga:
- adds migration for adding a country-code table + adding the actual country-codes
- adds migration for adding country-codes to scope
- adds an endpoint to get the countrycodes

Test checklist:
- [ ] make a get-request to /country-codes as an admin or manage-settings-permissions
- [ ] You should get back a list with country-codes
- [ ] make a get-request to /country-codes as a not logged in user or user with no admin or manage-settings-permissions
- [ ] you should get a 401-reply

Fixes ushahidi/platform#2599 .

Ping @ushahidi/platform


application/classes/Ushahidi/Core.php
```diff
@@ -395,6 +397,7 @@ public static function init()
 
 		$di->setter['Ushahidi_Formatter_Set']['setAuth'] = $di->lazyGet("authorizer.set");
 		$di->setter['Ushahidi_Formatter_CSV']['setAuth'] = $di->lazyGet("authorizer.csv");
+        $di->setter['Ushahidi_Formatter_CountryCode']['setAuth'] = $di->lazyGet("authorizer.country_code");
```
rowasc: indent

migrations/20180316133403_add_country_code_table.php
```diff
@@ -0,0 +1,57 @@
...
+            ])
+            ->addColumn('dial_code', 'string', [
+                'null' => false,
+                'default' => false
+            ])
+            ->addColumn('country_code', 'string', [
+                'null' => false,
+                'default' => false
+            ])
+          ->create();
```
rowasc: country code could be an index, I think. Leaving it up to Will, tho

src/Core/Tool/Authorizer/CountryCodeAuthorizer.php
```diff
@@ -0,0 +1,58 @@
...
+	use AclTrait;
+
+	/* Authorizer */
+	public function isAllowed(Entity $entity, $privilege)
+	{
+		// These checks are run within the `User` context.
+		$user = $this->getUser();
+
+		// Allow role with the right permissions to do everything else
+		if ($this->acl->hasPermission($user, Permission::MANAGE_SETTINGS)) {
```
rowasc: I might be wrong, but shouldn't everyone be able to see a list of countries or a specific country?

rowasc: as in , I think isAllowed would probably always need to return true

src/Init.php
```diff
@@ -379,6 +381,8 @@ function feature($name)
 $di->setter['Ushahidi\Core\Usecase\Set\SetRepositoryTrait']['setSetRepository'] = $di->lazyGet('repository.set');
 $di->setter['Ushahidi\Core\Usecase\Set\AuthorizeSet']['setSetAuthorizer'] = $di->lazyGet('authorizer.set');
 
+// $di->setter['Ushahidi\Core\Usecase\CountryCode']['setAuthorizer'] = $di->lazyGet('authorizer.set');
```
rowasc: remove?

src/Core/Tool/Authorizer/CountryCodeAuthorizer.php
rowasc: (for read/search , other methods should just be a 405)

Angamanga: I don't know. It really doesn't matter if its available for everyone. I limited it since I wanted to keep it consistent with the permissions that the targeted surveys have.

src/Init.php
Angamanga: heh, yes 🙈

application/classes/Ushahidi/Core.php
Angamanga: 🙈

Angamanga: Hm, it looks like its only githubs ui acting up, its not indented wrongly in the file itself

migrations/20180316133403_add_country_code_table.php
willdoran: Yeah, we under index now, I'd rather turn off indexes to tune down. It's worth having one here for sure.

crcommons:
Tested with the client code and it works perfectly!

Angamanga:
@willdoran Hmm, it should really be 401 :/ Looking into it

Issue #2523, opened 2018-02-27
Add SMS Beta Feature Flag to API


willdoran:
### Overview
The SMS feature will not be available to the entirety of .io, instead it will be feature flag limited to only the DREAMS partner deployments. There are around 15 of these, the full list should be retrieved from the DREAMS PEOs.

Unlike other feature flags which are simple enable/disable toggles this flag will need to understand which deployment it is currently on and whether that deployment is allowed to use the Beta Feature.

We should define this system so that it is resusable and augmentable, it should be possible to edit deployments participation in a Beta feature. We should also define how a feature leaves Beta and what happens(feature flag checks are removed or converted into simple enable/disable toggles).

### Requirements
- [ ] A Beta feature can be defined for all .io
- [ ] A list of deployments that have access to the Beta Feature can be defined and edited
- [ ] The Platform Client and API will take account of whether the feature is enabled or disabled and behave accordingly
- [ ] A beta feature flag can be converted to a standard feature flag
- [ ] The beta feature flag creation/edition can be controlled via a Cloud API endpoint and/or an ansible playbook

### Simple Approach
- [ ] Define new playbook in [Platform Cloud Ansible](https://github.com/ushahidi/platform-cloud-ansible) called Platform-Feature-Manager or extend existing Playbook
- [ ] Define Task to call Platform API feature CLI
- [ ] Define variables describing feature-name and deployments to set flag on
- [ ] Add feature flipper CLI to Platform API
- [ ] Add feature check inside Platform API to check if current Deployment Subdomain is in deployments list. If it is set Targeted Survey feature flag to true else ignore

### Test script

rowasc:
https://docs.google.com/spreadsheets/d/1cxzOJdR58x3EI6hDSCGnay7lPB0S62-MfwNokt3-jRU/edit#gid=1663680881
This list has all the deployments that need to be included, right? 
@dexwakhu 

@willdoran thinking we should add a separate ticket to actually enable the flag so it doesn't get lost?

rowasc:
to discuss: 
> The beta feature flag creation/edition can be controlled via a Cloud API endpoint and/or an ansible playbook
I feel a playbook may be technically better/easier to manage (fewer security concerns, for instance) at the dev level for this kind of thing, but I think also makes it harder if we ever want to provide a way for non-devs to enable beta flags in some deployments.  It's a tradeoff. 
[...]

PR #2634, opened 2018-03-28
Adding feature flag

willdoran:
- Adding feature flag exists
- Is false by default

Test checklist:
- [ ] When combined with related Client PR the Targeted Survey feature should be hidden by default
- [ ] When combined with related Client PR the Targeted Survey feature should be available when feature flag is set to True

Fixes ushahidi/platform#2523

Depends on [1055](https://github.com/ushahidi/platform-client/pull/1055)

Ping @ushahidi/platform

### DIFF
--- a/application/config/features.php
+++ b/application/config/features.php
@@ -62,8 +62,8 @@ return [
 	],
 
 	// Post locking
-	'post-locking' => [
-		'enabled' => TRUE,
+	'targeted-surveys' => [
+		'enabled' => FALSE,
 	],
 
 	// Post locking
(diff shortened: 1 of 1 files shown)


application/config/features.php
```diff
@@ -62,8 +62,8 @@
 	],
 
 	// Post locking
-	'post-locking' => [
```
rowasc: @willdoran sorry, why is the post-locking feature removed and targeted survey replacing it?

willdoran: ugh, stupidity

willdoran: fixed

PR #2677, opened 2018-03-29
2635 testservice provider

willdoran:
- Adding test sms service data provider 
- Feature flag will hide this from users by default

Test checklist:
- [ ] Enable feature on API
- [ ] You should know see a new data provider in the data source section of client
- [ ] You should be able to save params for this datasource
- [ ] When you leave and re-enter the data source page the params persist
- [ ] Disable feature on API
- [ ] You should not see the test data provider

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/application/classes/Ushahidi/Core.php
+++ b/application/classes/Ushahidi/Core.php
@@ -514,7 +514,9 @@ abstract class Ushahidi_Core {
 
 		// Form Stage repository parameters
 		$di->params['Ushahidi_Repository_Form_Contact'] = [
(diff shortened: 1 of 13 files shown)


composer.json
```diff
@@ -51,7 +51,7 @@
     },
     "require-dev": {
         "fzaninotto/faker": "~1.4",
-        "guzzle/guzzle": "~3.9.1",
+        "guzzlehttp/guzzle": "~6.3.2",
```
kinstelli: 🤘

Issue #2678, opened 2018-03-29
Ensure that contacts are only part of one active survey at a time


kinstella:
###  Background

Because an inbound message can't be automatically associated with a particular survey, if a contact is part of multiple active surveys, the answers they send to various SMS questions can easily be confused with the wrong questions.

Therefore, our original spec requires that contacts can only be part of one active survey at a time. 

The current plan is to assume that all newly imported contacts will just be part of the latest survey, and they will simply be marked inactive in the old survey.

### Requirements 
- [ ] It's assumed that the contact will just be associated with the newest survey.
- [ ] During the contact import process, check the table to see if the contact already exists as part of another `active` survey.
- [ ] We should establish exactly how to represent how each contact is considered part of an 'active' survey. (**Do we need to add another flag to this table, just to be sure?**)
- [ ] If the contact is already part of another survey, the API endpoint should inform platform-client (e.g., via a JSON array with this list of preexisting contacts) that a contact already exists as part of another survey.
- [ ] Contacts will be automatically included in the **new** survey, but will be assumed or somehow designated as 'inactive' in the **old** survey.

PR #2694, opened 2018-04-04
2561 SMS Incoming

rowasc:
This pull request makes the following changes:
-

Test checklist:
- [ ]

- [ ] I certify that I ran my checklist

Fixes ushahidi/platform# .

Ping @ushahidi/platform

PR #2697, opened 2018-04-05
2678 one targeted survey at a time

rowasc:
- #2678 Ensure contacts are active in only 1 targeted survey at a time

Test checklist:
- [ ] Try to add a contact that already exists (by number) in a targeted survey
- You should receive: a list of contacts that are valid in this survey but NOT in others due to having been invalidated 

{
    "form_id": 6,
    "count": 1,
    "invalidated_contacts": [
        {
            "contact": "99999999", 
            "contact_id": "8",
            "form_id": "7"
        }
    ]
}

Fixes ushahidi/platform# .

Ping @ushahidi/platform


rowasc:
@willdoran this is what we checked today, but depends/builds on top of  https://github.com/ushahidi/platform/pull/2694 so we should close that one out first

application/classes/Ushahidi/Repository/Contact.php
```diff
@@ -42,6 +42,13 @@ protected function getTable()
 		return 'contacts';
 	}
 
+
+    public function isWaitingForQuestion($id, $form_id)
+    {
+        //@TODO: implement this
+    }
```
kinstelli: This is probably unnecessary now, right?

application/classes/Ushahidi/Repository/Form/Contact.php
```diff
@@ -81,7 +83,17 @@ public function updateCollection(Array $entities, $form_id = null)
...
+			$contactOnActiveSurvey = $this->existsInActiveTargetedSurveyByContactNumber($entity->contact);
+			if ($contactOnActiveSurvey) {
+				$this->setInactiveTargetedSurvey($contactOnActiveSurvey['targeted_survey_state_id']);
+				$invalidatedContacts[] = [
+					'contact' => $contactOnActiveSurvey['contact'],
+					'contact_id' => $contactOnActiveSurvey['contact_id'],
+					'form_id' => $contactOnActiveSurvey['form_id']
+				];
+			}
 			//@fixme how to avoid this ugly line?
```
kinstelli: If we're going to live with this ugly line for now, can we at least explain why it's here?

rowasc: good point. updating

application/classes/Ushahidi/Repository/Form/Contact.php
```diff
@@ -203,6 +217,45 @@ public function existsInFormContactByContactNumber($contact, $form_id)
...
+
+	/**
+	 * @param int $contact_id
+	 * @param int $form_id
+	 * @return bool
+	 */
+	public function setInactiveTargetedSurvey($tss_id)
+	{
+		$repo = $this->targeted_survey_state_repo->get($tss_id);
+		$entity = $repo->setState(array('survey_status' => 'INVALID'));
```
kinstelli: do we currently depend on this to say 'INVALID', or can it be switched to something like 'REMOVED_FROM_SURVEY' or 'MOVED_TO_SURVEY_XXXX' if we know what that new survey is?

rowasc: mmmm. I think that might be a good idea yea. Let me check with Will. I also will hold on this merge because I want to move statuses to consts. This hardcoded statuses without even a const are gross

rowasc:
this is missing some controls for outgoing/incoming messagges to be more reliable. I'll ping in ~40 for a new review @kinstelli

src/Core/Entity/TargetedSurveyState.php
```diff
@@ -23,6 +23,12 @@ class TargetedSurveyState extends StaticEntity
 	protected $survey_status;
 	protected $message_id;
 
+	const STATUS_INVALID = 'INVALID';
+	const PENDING_RESPONSE = 'PENDING RESPONSE';
+	const RECEIVED_RESPONSE = 'RECEIVED RESPONSE';
+	const SURVEY_FINISHED = 'SURVEY FINISHED';
+	const INVALID_CONTACT_MOVED = 'ACTIVE CONTACT IN SURVEY ###';
```
rowasc: @willdoran instead of INVALID , @kinstelli suggested we set something like this with the survey number. I was for it initially but not sure now. I guess it could be useful?what do you think?

rowasc:
@kinstelli flagging to re-review

src/Core/Entity/TargetedSurveyState.php
willdoran: If it's not hard to implement then it might be neat extra info to have that we could convert later but if it is hard then don't. Because we can add it as an upgrade later. If they move a contact that's their choice.

rowasc: ok awesome. I'll add it should be a minute

rowasc:
@kinstelli added the new status message. Thanks for the feedback!

application/classes/Ushahidi/Formatter/Form/Contact.php
```diff
@@ -14,12 +14,19 @@
...
-	
-	public function __invoke($form_id, $entities = [])
+
+	/**
+	 * @param $form_id
+	 * @param array $entities (the entities that were added)
+	 * @param array $invalidatedContacts ([{'contact': number, 'contact': id, 'form_id:' form_id}...]
+	 * @return array|mixed
+	 */
+	public function __invoke($form_id, $entities = [],$invalidatedContacts =[])
```
kinstelli: Nitpicking: space after equals? `$invalidatedContacts =[]`

application/classes/Ushahidi/Formatter/Form/Contact.php
```diff
@@ -14,12 +14,19 @@
...
 	{
 		$data = [
 			'form_id'  => $form_id,
 			'count' => count($entities),
+			'invalidated_contacts'=> $invalidatedContacts
```
kinstelli: More spaces nitpicking :)

application/classes/Ushahidi/Repository/Form/Contact.php
```diff
@@ -81,8 +86,32 @@ public function updateCollection(Array $entities, $form_id = null)
...
+			$contactOnActiveSurvey = $this->existsInActiveTargetedSurvey($entity->contact);
+			if ($contactOnActiveSurvey) {
+				$this->targeted_survey_state_repo->setStatusByFormId(
+					$contactOnActiveSurvey['form_id'],
+					str_replace('###', $form_id,Entity\TargetedSurveyState::INVALID_CONTACT_MOVED)
+				);
+				/** force the message in the survey state to be expired
+				** so we don't send outbound messages by mistake on an invalidated contact-survey
+				**/
+				$message = $this->message_repo->get($contactOnActiveSurvey['message_id']);
```
rowasc: oh wow this is a bug.

application/classes/Ushahidi/Repository/Form/Contact.php
```diff
@@ -260,6 +257,19 @@ public function existsInActiveTargetedSurvey($contact)
...
+
+	/**
+	 * @param int $contact_id
+	 * @param int $form_id
+	 * @return bool
+	 */
+	public function setInactiveTargetedSurvey($tss_id, $form_id)
+	{
+		$repo = $this->targeted_survey_state_repo->get($tss_id);
+		$entity = $repo->setState(array('survey_status' => str_replace('###', $form_id,Entity\TargetedSurveyState::INVALID_CONTACT_MOVED)));
```
kinstelli: Nice work with adding to the constant; I didn't realize how complicated that would be when I suggested it.

PR #2709, opened 2018-04-06
Unique contacts bugfix

rowasc:
- Looks up contact before creating a new one

Test checklist:
- [ ] Pre-requisite: have a contact in the database
- [ ] Send the a contact number  to /forms/:id/contacts to add to a targeted survey (same as a contact number you already have)
- [ ] The targeted_survey_state table should reference the contact
- [ ] No new contact should be created

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/application/classes/Ushahidi/Repository/Form/Contact.php
+++ b/application/classes/Ushahidi/Repository/Form/Contact.php
@@ -121,19 +121,25 @@ class Ushahidi_Repository_Form_Contact extends Ushahidi_Repository implements
 			 * in phone number validation but we don't want to save it
 			 */
 			unset($entity->country_code);
-			$query = DB::insert($this->getTable())
-				->columns(array_keys($entity->asArray()));
-			$query->values($entity->asArray());
-			$result = $query->execute($this->db);
-			if (!isset($result[0])) {
-				throw new HTTP_Exception_500(
-					sprintf(
-						'Could not create contacts. Result:  %s',
-						var_export($entity, true)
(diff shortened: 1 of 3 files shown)


application/classes/Ushahidi/Repository/Form/Contact.php
```diff
@@ -110,19 +121,25 @@ public function updateCollection(Array $entities, $form_id = null)
...
-			if (!isset($result[0])) {
-				throw new HTTP_Exception_500(
-					sprintf(
-						'Could not create contacts. Result:  %s',
-						var_export($entity, true)
-					)
-				);
+			if (!$entity->id) {
+
+				$query = DB::insert($this->getTable())
```
kinstelli: Not a blocker, but curious if we could refactor the insert into its own method, (if something similar doesn't already exist), since `updateCollection()` is looking a little large.

application/classes/Ushahidi/Repository/Form/Contact.php
```diff
@@ -55,6 +54,18 @@ public function getEntity(Array $data = null)
 		return new Entity\Contact($data);
 	}
 
+	// CreateRepository
+	// ReadRepository
```
kinstelli: Maybe just a comment to clarify why we'd do this?
Does CreateRepository and ReadRepository still apply to this method?

rowasc:
great feedback  , thanks Mark. Merging

PR #2711, opened 2018-04-09
chore(ci): output contents of the application/logs upon test failure

tuxpiper:
If there's a failure in the tests, this will send the contents of the log files under application/logs to the CI build output

### DIFF
--- a/codeship-steps.yml
+++ b/codeship-steps.yml
@@ -1,6 +1,6 @@
 - name: "Test"
   service: test
-  command: composer test
+  command: test_reporter
 
 - type: parallel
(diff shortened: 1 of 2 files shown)


codeship-steps.yml
```diff
@@ -1,6 +1,6 @@
 - name: "Test"
   service: test
-  command: composer test
+  command: test_reporter
```
rjmackay: Does this always run composer test? Can we run custom commands through it too?
I've found it helpful for local testing to be able to run just a specific test so if we can preserve that it'd be really helpful

PR #2713, opened 2018-04-09
Fix outgoing message creation

rowasc:
- makes sure outgoing messages go through the validation of message.create

Test checklist:
- [ ] When I sent a message from a contact in a targeted survey with pending questions (ie there's at least one more attribute that will be sent out), a new OUTGOING message is created after receiving my message. 

Fixes ushahidi/platform# .

Ping @ushahidi/platform


rowasc:
@willdoran adding you since we discussed it earlier

Issue #2693, opened 2018-04-04
Name of post from sms should not contain contact-info


Angamanga:
The posts that are generated from each contact for targeted surveys contains contact info. This should be removed. 

## Spec: 
- When I retrieve a post that belongs to a targeted survey, I should not see the contact number that it was created for
- When I retrieve a post that belongs to a targeted survey, I should see "Survey Name" and an autogenerated number. Example: "newsurvey - 107812" 
- 

## Test script:
- Create a targeted survey with a few contacts
- Retrieve a list of posts (ie in data mode) that belong to the targeted survey
- Check that there is NO contact information in any of the posts

-------------

Context/options provided:
- Survey-name + contact ID?
- Auto-generate a person e.g. “Person 1…Person 2…Person 52”
- Abstract the whole number but keep the end of it - so “Message Text - XXXXXXXXX257” - last 3 numbers of the mobile number are still kept to keep the entry meaningful.
- Use a Time stamp so - “Message Text - 12:33” Entirely possible some messages will come in at identical times so is this enough to differentiate.  
- Are there any other pieces of information we can display a tie to an individual that is sub sequentially abstract to keep anon but meaningful to the deployment workers?

Survey name would be most recognisable for deployment workers for context and any kind of unique ID number/string will help differentiate between multiple posts that may be similar

Perhaps "Survey name + Auto-generated person number"
![screen_shot_2018-04-03_at_16.25.19_960.png](https://images.zenhubusercontent.com/5a6f2bce4b5806bc2bc4b158/e9be1e4b-972a-409f-b13c-8ef88856a71b)

Erioldoesdesign:
@justinscherer If you haven't seen this because of the various slack and git 'white noise'

Very in favour of 'Survey name + Auto-generated person number' solution.

justinscherer:
> "Very in favour of 'Survey name + Auto-generated person number' solution."

seconded.

rowasc:
@Angamanga @willdoran 
I can start on this today if we want to. It should be simple . Two questions :
- How long should the number be?
- Do we want some kind of uuid or just the contact id instead ?

willdoran:
@rowasc Let's not use a real system id, let's just use a somewhat inoffensive uuid - a number is probably easier for a user to parse than a hex - it would be neat if we could do a human readable version but it might come out as "furious-green-ideas" which might be more confusing.

It would seem likely that users might expect the number to be in someway meaningful, so why don't we just use the Post Id?

[...]

PR #2724, opened 2018-04-10
2693 - do not use contact number in the post

rowasc:
- Contact's posts are saved with form name + random number as their title 

Test checklist:
- [x] Create a set of contacts for a form
- [x] Verify that the posts created for each contact match the format "{Form title} - {random number}"

Fixes ushahidi/platform#2693

Ping @ushahidi/platform


application/classes/Ushahidi/Listener/ContactListener.php
```diff
@@ -60,6 +60,19 @@ public function setTargetedSurveyStateRepo(TargetedSurveyStateRepository $repo)
 		$this->targeted_survey_state = $repo;
 	}
 
+	// CreateRepository
+	private function getContactPostTitle($form_name, $contact_id)
+	{
+		try {
+			$uuid = substr(Ramsey\Uuid\Uuid::uuid4()->toString(), 0 , 5);
```
rowasc: @rjmackay we need to generate a random number that's around 4-6 characters long to use in the post title for posts that belong to a contact (just so they don't all have the same title). Ideas on a nicer way to do that?

rjmackay: http://php.net/manual/en/function.random-int.php <- requires PHP 7 though.
But there's a userland implementation too https://github.com/paragonie/random_compat (we might already be using that I think)

rjmackay: PHP has finally passed the bar of simple and safe crypto methods... its like a real programming language 😆

rowasc: ha! thanks I didn't realize that was a thing

rowasc:
@willdoran can you take a look? I'm using the library Robbie suggested now

Issue #2443, opened 2018-02-01
Extend Post Exporter CLI Capabilities


willdoran:
### Overview
This is the tool which physically loads, formats and outputs the Posts. 

### Spec
API:
- Post Exporter 
  - [ ] should handle filters to limit export data
  - [ ] should be use correct formatter to ensure fields are correctly formatted
  - [ ] should ensure data is written to file to avoid memory exhaustion
  - [ ] should return address of file created for the chunk it is responsible for
  - [ ] should correctly close file upon completion

### Test
- [x] Post Exporter CLI should accept filters
- [x] Post Exporter should output file with fields correctly formatter
- [x] Post exporter should output file with correct header column names
- [x] Post exporter should output file for only the chunk it is assigned

kinstelli:
Should we include "should accept a set of column names in order to limit export data"?

willdoran:
I think that should be handled by filters.

kinstelli:
I'm not sure if this is more relevant to the CSV exporter tool, or if it's an issue with the creation of CSVs to begin with, but since this is a problem that exists at the 200-row line -- which is where each chunk is divided -- I think this might be related to the exporter. 

In short, the number of columns is different for the first export than it is for the subsequent chunks. See the ID column (column M) in this spreadsheet. The IDs exist in this column for the first 200 records, but as you scroll past line 200 and all subsequent records, this ID is moved to 
column J.

It looks like three of the eight columns showing M are getting truncated on subsequent exports, but unfortunately, that's not clear from our data set:
https://docs.google.com/spreadsheets/d/1e0svJkSYSpei6J-d4oTiOcQhSUn9gsfMhsdYAEQ70cI/edit?usp=sharing

willdoran:
We should hold off on any implementation of removing empty columns it would be a complex piece of work for us to do whereas removing empty columns from any data analysis application is trivial. For me, this crosses a line into data cleaning which is not what we do. We are responsible for ensuring the data is in an intelligible consistent format.

@romina We should ensure that each of the files generated is consistent in terms of number of columns. Is there an issue where we are dropping/including columns dependent on filtering when generating the individual files?

kinstelli:
Closing this as the original spec is complete. Issues related to appear to be defined in  #2730 and #2739

PR #2730, opened 2018-04-11
2443 - Save attribute headers in export_job and ensure that csv columns are aligned correctly in all chunks of data

willdoran:
- Saves the attribute headers to the export_job table to make sure that when we run it in chunks with different data we still have ALL the headers we need and they aren't scrambled 

Test checklist:
Note: to call the exporter, use: 
`./bin/ushahidi exporter export --limit 200 --offset 3000 --include_header true --job 1` 
Replacing with the correct parameters. 

When I request data with or without a header, it remains in the same order
- [ ] Request the same chunk of data from the exporter twice, one with headers and one without headers
- [ ] Both data exports should be equal, except for the header

When I request multiple chunks of data (limit/offfset set) my data is still correctly sorted. 

- [ ] Request range 0-100 of posts from exporter, with headers
- [ ] Request range 100-201 of posts from exporter, with headers. This should have some fields that the first request didn't have (ie if there are no categories in the first request, this posts should have categories)
- [ ] Join the two CSVs and verify that the data is correctly aligned/sorted according to the headers

Fixes ushahidi/platform#2443 .

Ping @ushahidi/platform


application/classes/Ushahidi/Repository/Post/Export.php
```diff
@@ -8,90 +8,101 @@
...
-  public function retrieveSetNames($set_ids) {
-    $set_repo = service('repository.set');
-    $names = [];
-    foreach($set_ids as $set_id) {
-      $set = $set_repo->get($set_id);
-      array_push($names, $set->name);
-    }
-    return $names;
-  }
+	public function retrieveSetNames($set_ids)
```
rowasc: how are we setting this up with the headers?

rowasc: ie if there's more than 1set name

application/classes/Ushahidi/Console/PostExporter.php
```diff
@@ -141,23 +141,47 @@ protected function executeExport(InputInterface $input, OutputInterface $output)
...
+		$this->postExportRepository->setSearchParams($data);
+		$posts = $this->postExportRepository->getSearchResults();
+		service("formatter.entity.post.$format")->setFileSystem($this->fs);
+		service("formatter.entity.post.$format")->setAddHeader($add_header);
 
-        $this->postExportRepository->setSearchParams($data);
-		
-        $posts = $this->postExportRepository->getSearchResults();
+		$form_ids = $this->postExportRepository->getFormIdsForHeaders();
+		// we still need thenative fields such as create date
```
rowasc: no we don't they are here

application/classes/Ushahidi/Console/PostExporter.php
```diff
@@ -107,10 +107,10 @@ protected function executeExport(InputInterface $input, OutputInterface $output)
         $data = $this->data->get('search');
 
 		// Get CLI params
-		$limit = $input->getOption('limit', 100);
-		$offset = $input->getOption('offset', 0);
-		$job_id = $input->getOption('job', null);
-		$add_header = $input->getOption('include_header', true);
+		$limit = $input->getOption('limit') ? $input->getOption('limit') : 100;
```
rowasc: Note on this change: This was not working. Basically the default do not exist.GetOption accepts just the option name.

application/classes/Ushahidi/Console/PostExporter.php
```diff
@@ -141,23 +141,47 @@ protected function executeExport(InputInterface $input, OutputInterface $output)
...
+		service("formatter.entity.post.$format")->setFileSystem($this->fs);
+		service("formatter.entity.post.$format")->setAddHeader($add_header);
 
-        $this->postExportRepository->setSearchParams($data);
-		
-        $posts = $this->postExportRepository->getSearchResults();
+		$form_ids = $this->postExportRepository->getFormIdsForHeaders();
+		// we still need thenative fields such as create date
+		$attributes = $this->postExportRepository->getAttributes($form_ids);
+		$attributes[] = ['label' => 'Post ID', 'key' => 'id', 'type' => 'integer', 'input' => 'number', 'form_id' => 0, 'form_stage_id' => 0, 'form_stage_priority' => 0, 'priority' => 1];
```
rowasc: move this to getattributes maybe so it's less in the middle of everything

application/classes/Ushahidi/Console/PostExporter.php
```diff
@@ -141,23 +141,47 @@ protected function executeExport(InputInterface $input, OutputInterface $output)
...
-			$posts[$idx] = $post;
+			$post = $this->postExportRepository->retrieveMetaData($post->asArray(), $keyAttributes);
+ 			$posts[$idx] = $post;
+ 		}
+		/**FIXME: how to make sure header_row is null/empty instead off an array with an empty item in it? */
+		if (empty($job->header_row) || $job->header_row[0] == '') {
+
+			$header_row = service("formatter.entity.post.$format")->createHeading($attributes, $posts);
+			$job->setState(['header_row' => $header_row]);
+			//FIXME update does not exist $job->update();
```
rowasc: oh @willdoran $job says update does not exist.

application/classes/Ushahidi/Formatter/Post/CSV.php
```diff
@@ -34,25 +35,44 @@ class Ushahidi_Formatter_Post_CSV extends Ushahidi_Formatter_API
...
+	}
+
 	public function setFilesystem($fs)
 	{
 		$this->tmpfname = "tmp" . DIRECTORY_SEPARATOR . strtolower(uniqid() . '-' . strftime('%G-%m-%d') . '.csv');
 		$this->fs = $fs;
 	}
 
+	public function createHeading($attributes, $records)
+	{
```
rowasc: maybe $attributes can be set as a property of the formatter to avoid passing it around?

rowasc: it's not the same as heading because it has all the data to decide what to do about each field

application/classes/Ushahidi/Console/PostExporter.php
```diff
@@ -141,23 +141,47 @@ protected function executeExport(InputInterface $input, OutputInterface $output)
...
+		$form_ids = $this->postExportRepository->getFormIdsForHeaders();
+		// we still need thenative fields such as create date
+		$attributes = $this->postExportRepository->getAttributes($form_ids);
+		$attributes[] = ['label' => 'Post ID', 'key' => 'id', 'type' => 'integer', 'input' => 'number', 'form_id' => 0, 'form_stage_id' => 0, 'form_stage_priority' => 0, 'priority' => 1];
+		$attributes[] = ['label' => 'Created (UTC)','key' => 'created', 'type' => 'datetime', 'input' => 'native', 'form_id' => 0, 'form_stage_id' => 0, 'form_stage_priority' => 0, 'priority' => 2];
+		$attributes[] = ['label' => 'Updated (UTC)', 'key' => 'updated', 'type' => 'datetime', 'input' => 'native', 'form_id' => 0, 'form_stage_id' => 0, 'form_stage_priority' => 0, 'priority' => 3];
+		$attributes[] = ['label' => 'Contact ID','key' => 'contact_id', 'type' => 'integer', 'input' => 'number', 'form_id' => 0, 'form_stage_id' => 0, 'form_stage_priority' => 0, 'priority' => 4];
+		$attributes[] = ['label' => 'Contact', 'key' => 'contact', 'type' => 'text', 'input' => 'text', 'form_id' => 0, 'form_stage_id' => 0, 'form_stage_priority' => 0, 'priority' => 5];
+		$keyAttributes = [];
```
willdoran: I tried but gave up.

application/classes/Ushahidi/Console/PostExporter.php
willdoran: That's weird it should because it inherits from Repository, I'll look

application/classes/Ushahidi/Console/PostExporter.php
```diff
@@ -131,34 +131,46 @@ protected function executeExport(InputInterface $input, OutputInterface $output)
...
 
-        foreach ($filters as $key => $filter) {
-            $data->$key = $filter;
+		foreach ($filters as $key => $filter) {
+			$data->$key = $filter;
+		}
+		$this->postExportRepository->setSearchParams($data);
+		$posts = $this->postExportRepository->getSearchResults();
+		service("formatter.entity.post.$format")->setFileSystem($this->fs);
+		service("formatter.entity.post.$format")->setAddHeader($add_header ===  'true');
```
rowasc: is there a way to make the input from the cli be an expected type? (to get a bool instead of a "true" string)

migrations/20180411200942_add_header_to_export_job.php
```diff
@@ -0,0 +1,30 @@
...
+{
+
+    /**
+     * Migrate Up.
+     */
+    public function up()
+    {
+        $this->table('export_job')
+            ->addColumn('header_row', 'text',
+				['null' => true, 'limit' => 16777215, 'default' => null]
```
rowasc: I have no idea how long this can be. It's a representation of all attributes that the user is exporting for their posts with priority, labels, type, etc

application/classes/Ushahidi/Repository/Post/Export.php
rowasc: TODO  I need to fix this PR to include set names

application/classes/Ushahidi/Repository/Form/Attribute.php
```diff
@@ -234,6 +234,19 @@ public function getByForm($form_id)
 		return $this->getCollection($results->as_array());
 	}
 
+	public function getByForms($form_ids) {
```
rowasc: TODO: I should move the attributes query in postexport repo to live here and call this repo

application/classes/Ushahidi/Repository/Post/Export.php
```diff
@@ -8,90 +8,163 @@
...
+		$result = $searchQuery->resetSelect()
+			->select([DB::expr('DISTINCT(posts.form_id)'), 'form_id'])->execute($this->db);
+		$result =  $result->as_array();
+		return array_column($result, 'form_id');
+	}
+
+	/**
+	 * fixme move to correct repo
+	 */
+	public function getAttributes($form_ids) {
```
rowasc: TODO goes to form attribute repo

application/classes/Ushahidi/Repository/Post/Export.php
```diff
@@ -8,90 +8,163 @@
...
+			"INNER JOIN form_stages ON form_attributes.form_stage_id = form_stages.form_id " .
+			"INNER JOIN forms ON form_stages.form_id = forms.id " .
+			"where forms.id IN :forms
+			ORDER BY form_stages.priority, form_attributes.priority";
+		$results = DB::query(Database::SELECT, $sql)
+			->bind(':forms', $form_ids)
+			->execute($this->db);
+		$attributes = $results->as_array();
+
+		$native = [[
```
rowasc: not sure what to do about this right now

application/classes/Ushahidi/Console/PostExporter.php
```diff
@@ -104,21 +104,21 @@ protected function executeExport(InputInterface $input, OutputInterface $output)
 	{
 		$userContextService = service('usercontext.service');
```
rjmackay: This is definitely future music... but in an ideal world all our commands would just execute a usecase with minimal setup/shutdown. Any idea how hard it would be to move most of this into a custom export usecase? how much is specific to running at the CLI? or outputting to the filesystem? Could we just set up the formatter and then handoff to a usecase?

In case its not clear: *Please don't even consider doing that work now*

application/classes/Ushahidi/Formatter/Post/CSV.php
```diff
@@ -34,25 +35,52 @@ class Ushahidi_Formatter_Post_CSV extends Ushahidi_Formatter_API
...
-	protected $add_header;
+	protected $add_header = true;
+	protected $heading;
 
 	// Formatter
-	public function __invoke($records)
+	public function __invoke($records, $attributes = [])
 	{
-		return $this->generateCSVRecords($records);
+		if ($this->heading) {
```
rjmackay: So headers are set for the entire CSV and then we call `invoke` for each line? Am I following this all correctly?

application/classes/Ushahidi/Repository/Form/Attribute.php
rowasc: done

application/classes/Ushahidi/Repository/Post/Export.php
rowasc: done

migrations/20180411200942_add_header_to_export_job.php
rowasc: right now it's set the length for mysql's "Medium text" ...

application/classes/Ushahidi/Formatter/Post/CSV.php
```diff
@@ -144,25 +165,26 @@ private function writeStreamToFS($stream)
 
 	}
 
-	private function getValueFromRecord($record, $keyParam){
+	private function getValueFromRecord($record, $keyParam, $attributes){
```
rjmackay: I'm having a lot of trouble following this method... can you write the story in comments? ie. literate style comments

application/classes/Ushahidi/Formatter/Post/CSV.php
```diff
@@ -144,25 +165,26 @@ private function writeStreamToFS($stream)
...
 		$isDateField = $recordAttributes['input'] === 'date' && $recordAttributes['type'] === 'datetime';
 
-		if ($isDateField) {
+		if ($isDateField && isset($recordValue[$headingKey])) {
 			$date = new DateTime($recordValue[$headingKey][$key]);
 			$recordValue[$headingKey][$key] = $date->format('Y-m-d');
 		}
-
- 		if($format === 'single_value_array'){
+		if($format === 'single_value_array'){
```
rjmackay: code style... missing spaces

application/classes/Ushahidi/Repository/Post/Export.php
```diff
@@ -8,90 +8,89 @@
...
-    $names = [];
-    foreach($tag_ids as $tag_id) {
-      $tag = $tag_repo->get($tag_id);
-      array_push($names, $tag->tag);
-    }
-    return $names;
-  }
+	public function retrieveTagNames($tag_ids)
+	{
+		$tag_repo = service('repository.tag');
```
rjmackay: You should inject this. Using `service` should be an exception and mostly only appear in controllers/commands, not in the core repos

application/classes/Ushahidi/Repository/Post/Export.php
```diff
@@ -8,90 +8,89 @@
...
-  public function retrieveTagNames($tag_ids) {
-    $tag_repo = service('repository.tag');
-    $names = [];
-    foreach($tag_ids as $tag_id) {
-      $tag = $tag_repo->get($tag_id);
-      array_push($names, $tag->tag);
-    }
-    return $names;
-  }
+	public function retrieveTagNames($tag_ids)
```
rjmackay: Maybe move this to the tag repo itself?

application/classes/Ushahidi/Repository/Post/Export.php
```diff
@@ -8,90 +8,89 @@
...
-    $names = [];
-    foreach($set_ids as $set_id) {
-      $set = $set_repo->get($set_id);
-      array_push($names, $set->name);
-    }
-    return $names;
-  }
+	public function retrieveSetNames($set_ids)
+	{
+		$set_repo = service('repository.set');
```
rjmackay: Same deal on using service... and maybe moving this to another repo?

application/classes/Ushahidi/Repository/Post/Export.php
```diff
@@ -8,90 +8,89 @@
...
-    }
-
-    $data += ['attributes' => $attributes];
-
-
-    // Set Set names
-    if (!empty($data['sets'])) {
-        $data['sets'] = $this->retrieveSetNames($data['sets']);
-    }
+	public function retrieveMetaData($data, $attributes)
```
rjmackay: What does this actually get? Whats in data?

application/classes/Ushahidi/Console/PostExporter.php
```diff
@@ -131,34 +136,47 @@ protected function executeExport(InputInterface $input, OutputInterface $output)
...
+			$keyAttributes[$item['key']] = $item;
+		}
 
 		// // ... remove any entities that cannot be seen
 		foreach ($posts as $idx => $post) {
-
 			// Retrieved Attribute Labels for Entity's values
-			$post = $this->postExportRepository->retrieveColumnNameData($post->asArray());
-
+			$post = $this->postExportRepository->retrieveMetaData($post->asArray(), $keyAttributes);
```
rjmackay: Isn't this going to be pretty horribly inefficient? ie. reload attributes etc for *every* post?
Are we caching somewhere? or can we do this in aggregate somehow?

application/classes/Ushahidi/Repository/Post/Export.php
rowasc: done need to push

application/classes/Ushahidi/Console/PostExporter.php
rowasc: oh. OK that sounds way more reasonable .

application/classes/Ushahidi/Formatter/Post/CSV.php
rowasc: > So headers are set for the entire CSV 
YES
> and then we call invoke for each line?
NOPE. This is called once `
		$file = service("formatter.entity.post.$format")->__invoke($posts, $keyAttributes);`

application/classes/Ushahidi/Repository/Set.php
```diff
@@ -227,4 +227,19 @@ public function addPostToSet($set_id, $post_id)
...
+	 * @param $tag_ids
+	 * @return array
+	 */
+	public function getNamesByIds($sets_ids)
+	{
+		$result = $this->selectQuery(array('id' => $sets_ids))
+			->resetSelect()
+			->select('name')
+			->execute($this->db);
+		$result = $result->as_array('name');
```
rowasc: I only need the name so I used 'name' as the key ... I'm not sure if there's a better way to easily get an array like ['Tagg1', 'Tag random', 'Explosion tag'] other than getting the keys. Thoughts?

rowasc: @rjmackay tagging

rjmackay: Try `as_array(null, 'name')` ? I'm not 100% sure it work but I think I've done it somewhere before

application/classes/Ushahidi/Console/PostExporter.php
rjmackay: Good point. Follow up PR? Or just backlog a tech debt task?

application/classes/Ushahidi/Formatter/Post/CSV.php
rjmackay: Thats weird then. Why don't we just pass headers as a param with all the posts?

application/classes/Ushahidi/Console/PostExporter.php
rowasc: follow up PR for sure. I'd rather solve it while it's fresh. But I'll open myself a ticket because otherwise It'll get lost in the morning

application/classes/Ushahidi/Formatter/Post/CSV.php
rowasc: oh. I should add a comment if I didn't . The headers var is good for output (has the correct columns) but it's useless for the processing tasks we need to do in the formatter (they don't have type and other attrs that $keyAttributes has)

application/classes/Ushahidi/Console/PostExporter.php
```diff
@@ -83,42 +82,32 @@ protected function configure()
...
-		];
+			->addOption('limit', ['l'], InputOption::VALUE_OPTIONAL, 'limit', 100)
+			->addOption('offset', ['o'], InputOption::VALUE_OPTIONAL, 'offset', 0)
+			->addOption('job', ['j'], InputOption::VALUE_OPTIONAL, 'job', null)
+			->addOption('include_header', ['ih'], InputOption::VALUE_OPTIONAL, 'include_header', 1)
+		;
 	}
 
-	protected function executeExport(InputInterface $input, OutputInterface $output)
+	protected function execute(InputInterface $input, OutputInterface $output)
```
rjmackay: Just check what the parent execute method does... you might need to do some output formatting yourself now you've dropped the actions.

application/classes/Ushahidi/Console/PostExporter.php
```diff
@@ -164,25 +148,18 @@ protected function executeExport(InputInterface $input, OutputInterface $output)
...
-			$_attributes[$item['key']] = $item;
+
+		if (empty($job->header_row)) {
+			$job->setState(['header_row' => $attributes]);
+            $this->exportJobRepository->update($job);
 		}
+		$header_row = $this->formatter->createHeading($job->header_row, $posts);
+		$this->formatter->setHeading($header_row);
+
+		$file = service("formatter.entity.post.$format")->__invoke($posts, $keyAttributes);
```
rjmackay: This can be `this->formatter` now right?

rjmackay: although I don't see the actual injection set up yet?

application/classes/Ushahidi/Console/PostExporter.php
```diff
@@ -131,34 +120,46 @@ protected function executeExport(InputInterface $input, OutputInterface $output)
...
-		$file = service("formatter.entity.post.$format")->__invoke($posts);
-		
+		if (empty($job->header_row)) {
+			$job->setState(['header_row' => $attributes]);
+            $this->exportJobRepository->update($job);
+		}
+		$header_row = $this->formatter->createHeading($job->header_row, $posts);
+		$this->formatter->setHeading($header_row);
+
+		$file = service("formatter.entity.post.$format")->__invoke($posts, $keyAttributes);
```
rjmackay: This should be a `this->formatter` too?

application/classes/Ushahidi/Formatter/Post/CSV.php
```diff
@@ -34,25 +35,52 @@ class Ushahidi_Formatter_Post_CSV extends Ushahidi_Formatter_API
...
 
 	// Formatter
-	public function __invoke($records)
+	public function __invoke($records, $attributes = [])
 	{
-		return $this->generateCSVRecords($records);
+		if ($this->heading) {
+			return $this->generateCSVRecords($records, $attributes);
+		} else {
+			//throw exception
```
rjmackay: Probably time to add this?

rowasc: oops. YEP on it

application/classes/Ushahidi/Console/PostExporter.php
rowasc: ./debug ./bin/ushahidi exporter --limit 200 --include_header true --job 1
[{"file":"csv\/tmp\/5acf8b46ba35e-2018-04-12.csv"}]
Apparently it's looking good 👍 and the handlereponse call is aligned with parent options

application/classes/Ushahidi/Console/PostExporter.php
rowasc: solved

rowasc:
https://github.com/ushahidi/platform/issues/2739
Reference to techdebt/follow up ticket

PR #2738, opened 2018-04-13
refactor: Clean up tag's isRoleValid check so its self contained

rjmackay:
- This was breaking because Validation wasn't namespaced
- Remove need for validation at all
- Clean up error message structures
- Add test for mismatched role
- Check error message in test for invalid role

Todo:
- [x] Confirm if dropped hasRole check made sense

Test checklist:
- [ ] Pre-requisite: Logged in, in creation page for a category. Example: http://csvexportsforever.v3-qa.ush.zone/settings/categories/create
- [ ] Create a category and set `Who can see this category?` to Admin
- [ ] Create a child category and check `Who can see this category?` shows permissions are managed by parent
- [ ] Both categories saved correctly
- [ ] Create a category and set `Who can see this category?` to everyone
- [ ] Create a child category and check `Who can see this category?` shows permissions are managed by parent
- [ ] Both categories saved correctly
- [ ] On post data view, in filters: restricted categories are visible when logged in as admin
- [ ] On post data view, in filters:  restricted categories are not visible when logged out
- [ ] Add categories to a surveys category fields
- [ ] Open the post create page for that survey: restricted categories are visible when logged in as admin
- [ ] Open the post create page for that survey:  restricted categories are not visible when logged out

Ping @ushahidi/platform

### DIFF
--- a/application/classes/Ushahidi/Repository/Tag.php
+++ b/application/classes/Ushahidi/Repository/Tag.php
@@ -182,7 +182,7 @@ class Ushahidi_Repository_Tag extends Ushahidi_Repository implements
 			$parent = $this->getEntity($parent);
 
 			// ... and check if the role matches its parent
-			if ($parent->role == $tag['role']) {
+			if ($parent->role != $tag['role']) {
 				// If it doesn't, set a validation error
 				// We have to do this here because an empty field gets ignored
 				// by KohanaValidation
--- a/application/classes/Ushahidi/Repository/Tag.php
+++ b/application/classes/Ushahidi/Repository/Tag.php
@@ -171,9 +171,8 @@ class Ushahidi_Repository_Tag extends Ushahidi_Repository implements
 	 * @param $fullData
 	 * @return bool
 	 */
-	public function isRoleValid($tag)
+	public function isRoleValid(\Ushahidi\Core\Tool\ValidationEngine $validation, $tag)
 	{
-		$valid = true;
 		$isChild = !!$tag['parent_id'];
 		$parent = $isChild ? $this->selectOne(['id' => $tag['parent_id']]) : null;
 
@@ -181,11 +180,19 @@ class Ushahidi_Repository_Tag extends Ushahidi_Repository implements
 		if ($isChild && $parent) {
 			// ... load the parent
 			$parent = $this->getEntity($parent);
+
 			// ... and check if the role matches its parent
-			return $parent->role == $tag['role'];
+			if ($parent->role == $tag['role']) {
+				// If it doesn't, set a validation error
+				// We have to do this here because an empty field gets ignored
+				// by KohanaValidation
+				$validation->error('role', 'isRoleValid');
+				// And return false
+				return false;
+			}
 		}
 
-		// Otherwise
+		// Otherwise role is fine
 		return true;
 	}
 }
--- a/application/classes/Ushahidi/Repository/Tag.php
+++ b/application/classes/Ushahidi/Repository/Tag.php
@@ -171,20 +171,21 @@ class Ushahidi_Repository_Tag extends Ushahidi_Repository implements
 	 * @param $fullData
 	 * @return bool
 	 */
-	public function isRoleValid(Validation $validation, $fullData)
+	public function isRoleValid($tag)
 	{
 		$valid = true;
-		$entityFullData = $this->getEntity($fullData);
-		$isChild = !!$entityFullData->parent_id;
-		$hasRole = !!$entityFullData->role;
-		$parent = $isChild ? $this->selectOne(['id' => $entityFullData->parent_id]) : null;
-		if ($hasRole && $isChild && $parent) {
+		$isChild = !!$tag['parent_id'];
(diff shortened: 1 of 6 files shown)


application/messages/tag.php
```diff
@@ -2,9 +2,11 @@
 
 return array(
 	'isSlugAvailable' => ':field :value is already in use',
-	'isRoleValid' => 'Role must match the parent category',
-	'tag.role.isRoleValid' => 'Role must match the parent category',
+	'role' => [
+		'isRoleValid' => 'Role must match the parent category',
+		'exists' => 'Role :value does not exist'
```
rjmackay: This text was outright missing

application/classes/Ushahidi/Repository/Tag.php
```diff
@@ -171,20 +171,21 @@ public function deleteTag($id)
...
-		$entityFullData = $this->getEntity($fullData);
-		$isChild = !!$entityFullData->parent_id;
-		$hasRole = !!$entityFullData->role;
-		$parent = $isChild ? $this->selectOne(['id' => $entityFullData->parent_id]) : null;
-		if ($hasRole && $isChild && $parent) {
+		$isChild = !!$tag['parent_id'];
+		$parent = $isChild ? $this->selectOne(['id' => $tag['parent_id']]) : null;
+
+		// If tag has a role and is a child category
+		if ($isChild && $parent) {
```
rjmackay: I've dropped `hasRole` because even if the tag *doesn't* have a role we should still double check that it matches the parent role (ie. no role on both!)

application/classes/Ushahidi/Repository/Tag.php
```diff
@@ -171,20 +171,21 @@ public function deleteTag($id)
 	 * @param $fullData
 	 * @return bool
 	 */
-	public function isRoleValid(Validation $validation, $fullData)
+	public function isRoleValid($tag)
 	{
 		$valid = true;
-		$entityFullData = $this->getEntity($fullData);
```
rjmackay: Dropped this. I don't think we need to build an object.

tests/integration/tags.feature
```diff
@@ -345,4 +345,27 @@ Feature: Testing the Tags API
         When I request "/tags"
         Then the response is JSON
         And the response has a "errors" property
+        And the "errors.1.message" property contains "Role nope does not exist"
```
rjmackay: Verifying error messages \o/

application/classes/Ushahidi/Repository/Tag.php
rowasc: I'll come back to this when my brain works properly ... but I think children tag usually didn't have a role setup from the client* and a null role in children usually represented just that it's the same as parent 
*the validation was for the weird cases where we somehow got one

rjmackay: Ok. I'll check the flow. We should probably have set that role before this point if thats the case.

rjmackay: It turns out this still fails, but also failed before. 
Basically: if the role is null the validator just ignore the result of this check entirely :( I'll figure out a workaround tomorrow

rjmackay: Ugh So we didn't need to `hasRole` check. But we did have to manually set the error via `$validation` because Kohana sucks so much

rowasc: as long as categories always have a role, this will work . But "everyone can see" matches null role instead of a real role. That might be something we want to solve in the client ASAP with a 'annonymous' role, just saying that the hasRole check is needed at the moment.
source: checked in production and we don't send or receive a role for "Everyone can see" permission

rjmackay: But thats sorta the point of my change. Previously: if we had a null role, the check didn't run so a parent could have role=admin, but the child would have role=null.
Don't we want to always force the role to match the parent? Or have a wildly misinterpreted?

TBH... the safer solution to all of this might be to forcibly overwrite the role in the UseCase

application/classes/Ushahidi/Repository/Tag.php
```diff
@@ -171,20 +171,28 @@ public function deleteTag($id)
...
+		if ($isChild && $parent) {
+			// ... load the parent
 			$parent = $this->getEntity($parent);
-			$valid = $parent->role == $entityFullData->role;
-		}
-		if (!$valid) {
-			$validation->error('role', 'tag.role');
+
+			// ... and check if the role matches its parent
+			if ($parent->role == $tag['role']) {
```
rjmackay: Ugh this should be !=. How did my tests even pass :/

application/classes/Ushahidi/Repository/Tag.php
rowasc: >TBH... the safer solution to all of this might be to forcibly overwrite the role in the UseCase

yup.

rjmackay:
@rowasc I've updated to include auth checks on the parent.
I haven't update the usecases, I'm going to leave those as-is for now. So this can get landed sooner rather than later

PR #2765, opened 2018-04-16
Bugfix/2753

rowasc:
-

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/application/classes/Ushahidi/Repository/Contact.php
+++ b/application/classes/Ushahidi/Repository/Contact.php
@@ -132,7 +132,8 @@ class Ushahidi_Repository_Contact extends Ushahidi_Repository implements
     {
         $query = DB::select('targeted_survey_state.contact_id', 'targeted_survey_state.form_id')
             ->from('targeted_survey_state')
-            ->where('contact_id', '=', $contact_id);
+            ->where('contact_id', '=', $contact_id)
+			->and_where('survey_status', '!=', Entity\TargetedSurveyState::SURVEY_FINISHED);
 
         if($query->execute($this->db)->count() > 0)
         {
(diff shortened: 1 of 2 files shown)


application/classes/Ushahidi/Repository/Contact.php
```diff
@@ -143,6 +144,29 @@ public function isInTargetedSurvey($contact_id)
...
+			->where("messages.post_id", 'NOT IN',
+				DB::query
+				(
+					Database::SELECT, 'select targeted_survey_state.post_id from targeted_survey_state where contact_id = :contact'
+				)
+				->bind(':contact', $contact_id))
+			->where(DB::expr('messages.contact_id'), '=', $contact_id);
+
+		$post_ids_in_messages = $query_posts->execute($this->db)->as_array(null, 'post_id');
+		if (count($post_ids_in_messages) >= 1) {
```
rowasc: @kinstelli problem: what if the contact already has multiple posts? obviously this will happen because this limitation is new . I'm just getting one post id but that's not ideal
@rjmackay

application/classes/Ushahidi/Repository/Contact.php
```diff
@@ -132,7 +132,8 @@ public function isInTargetedSurvey($contact_id)
     {
         $query = DB::select('targeted_survey_state.contact_id', 'targeted_survey_state.form_id')
             ->from('targeted_survey_state')
-            ->where('contact_id', '=', $contact_id);
+            ->where('contact_id', '=', $contact_id)
+			->and_where('survey_status', '!=', Entity\TargetedSurveyState::SURVEY_FINISHED);
```
kinstelli: good catch. should this also include INACTIVE? (or whatever we ended up calling it?)

rowasc: ugh  I might be wrong . .. but I don't think so? , because then the user is in a targeted survey, just another one. Here we just want to filter out the finished ones so users can communicate /send inbounds after they fill  a survey.

rowasc: solved thank you !

application/classes/Ushahidi/Repository/Contact.php
kinstelli: I agree it's an issue, but I can't think of the best way to solve it yet.

application/classes/Ushahidi/Repository/Contact.php
```diff
@@ -132,8 +132,10 @@ public function isInTargetedSurvey($contact_id)
     {
```
kinstelli: For clarity, could we call this isInActiveTargetedSurvey, since I think that's really what we're trying to differentiate?

application/classes/Ushahidi/Repository/Contact.php
kinstelli: I keep thinking that for each incoming message, the flow should be like this:
- if this contact in an active targeted survey, then attach that message to the post for that active survey
- if they're not in an active survey, then we should just find the _latest_ post

On the client, if the user is looking at a targeted survey post, then we could filter out any messages that aren't related to that particular post_id.
If the user is _not_ looking at a targeted survey, then maybe we can include every message for that user _except_ for posts in any targeted surveys.

I might be forgetting something, but this seems like it might account for the situations we've discussed.

rowasc:
Closing, we are not doing this. Updating ticket (we are going to filter out in the threads but not address the new post concern)

PR #2776, opened 2018-04-17
Bugfix/csv order by filter error

rowasc:
- We do not need the orderBy that comes from filters to select post ids. This caused an exception in some filtering queries that came in

Test checklist:
- [x] start an Export  job with filters selected from data mode
- [x] A CSV should be generated according to your filters 
- [x] start an Export job from settings/export without filters.
- [x] A CSV should be generated without filters

Fixes ushahidi/platform# .

Ping @ushahidi/platform


rowasc:
@kinstelli need a review for this hotfix .

PR #2782, opened 2018-04-17
WIP ! Allow /external/:action/:param endpoints to use the UserContextService feature instead of session.user

rowasc:
Prototypey I should clean  it up after discussing
/External/xx endpoints need access to the user context that the exporter cli usess . This WIP allows that to happen by enabling some paths to not have [setUse]=session.user assigned

Context information for discussion:
- When you are in a cli task, there’s no OAuth. 
- For CSV, we needed a way to actually load the user because we need permission control to work. 
- This feature lets you set the user directly from a cli or some other place where you need to not depend on the normal session.user. 
- This originally was designed for CLI only. But … the /external  usecases need it too. So that came back as a bug where if you only had draft posts you would not get anything back since could would be zero for anon users. To solve that, I had to not only do the setContext thing for cli commands but for normal API things that do not have our session at all


rjmackay:
Replaced by #2785

PR #2785, opened 2018-04-17
Add user session based on export's user id - /external/count endpoint

rowasc:
- Adds user session based on export's user_id

Test checklist:
- [x]  Pre requisite: a database with published and draft posts
- [x] Send a count request for an export job with an admin user. You should get a count of all posts if you didn't assign filters.
- [x] Send a count request for an export job with a user with less permissions (ie role=user). You should get a count that includes only published posts

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/application/classes/Ushahidi/Console/PostExporter.php
+++ b/application/classes/Ushahidi/Console/PostExporter.php
@@ -14,7 +14,6 @@ use Ushahidi\Core\Entity\PostExportRepository;
 use Ushahidi\Core\Entity\ExportJobRepository;
 use \Ushahidi\Core\Entity\FormAttributeRepository;
 use Ushahidi\Factory\DataFactory;
-use Ushahidi\Core\UserContextService;
 use Ushahidi\Core\Tool\FormatterTrait;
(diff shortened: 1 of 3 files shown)


application/classes/Ushahidi/Console/PostExporter.php
```diff
@@ -14,7 +14,6 @@
 use Ushahidi\Core\Entity\ExportJobRepository;
 use \Ushahidi\Core\Entity\FormAttributeRepository;
 use Ushahidi\Factory\DataFactory;
-use Ushahidi\Core\UserContextService;
```
rowasc: @rjmackay I think this was missing in the initial PR ? adding

rowasc:
@kinstelli can you confirm: is there another external endpoint where the context matters?  (I think not since the others are cli and jobs?)

kinstelli:
@rowasc The only ones I'm aware of are for CSV:
`/api/v3/exports/external/count`
`/api/v3/exports/external/cli`
`/api/v3/exports/external/jobs`

PR #2792, opened 2018-04-19
Lumen refactor receive message

rjmackay:
- Refactor ReceiveMessage usecase
- Ensure message is always saved
- Remove HTTP errors. They should never be in a usecase
- Remove unneeded validation
- Tidy code
- Add way more comments as I figure out whats going on

Todo:
- [ ] Add more tests
- [x] Ensure messages are always associated to a post

Test checklist:
- [ ] composer test

Ping @ushahidi/platform

### DIFF
--- a/src/Core/Usecase/Message/ReceiveMessage.php
+++ b/src/Core/Usecase/Message/ReceiveMessage.php
@@ -174,7 +174,9 @@ class ReceiveMessage extends CreateUsecase
 				]
 			);
 
-			// @todo Should we actually create a new post if this happens?!
+			// Create a new post
+			$post_id = $this->createPost($incoming_message);
+			$incoming_message->setState(compact('post_id'));
 
 			// But always save the message anyway - otherwise its lost forever
 			return $this->repo->create($incoming_message);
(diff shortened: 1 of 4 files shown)


rjmackay:
@rowasc @kinstelli can you check this makes sense?

The biggest behaviour change is that receiving a message pretty much always returns 200 and tries to save the message. That because the SMS providers may not send the message again if we fail to save it, or some of them will send it over and over again until we return 200. Basically: we should always save an incoming message, even if the targeted survey is in a weird state.

src/Core/Usecase/Message/ReceiveMessage.php
```diff
@@ -107,120 +107,139 @@ public function setTargetedSurveyStateRepo(Entity\TargetedSurveyStateRepository
...
-					'contact_id' => $contact_id,
+					'contact_id' => $incoming_message->contact_id,
 					'form_id' => $surveyStateEntity->form_id
 				]
 			);
-			throw new HTTP_Exception_400(
-				'Outgoing question not found for contact ' . $contact_id . ' and form ' . $surveyStateEntity->form_id
-			);
+
+			// @todo Should we actually create a new post if this happens?!
```
rowasc: YES. If this happens and we don't reject, at least we should make sure it doesn't end up in a targeted survey's post.

rjmackay: So that already happens. It won't end up in the targetted survey *but* it ends up nowhere. It doesn't get its own post, and it doesn't go on the targetted survey.

I should probably refactor a bit more to make match the non-targeted survey flow

rjmackay:
Closing. Replaced by #2832

PR #2802, opened 2018-04-20
Lumen: Test and fix external export routes

rjmackay:
- Add tests for all exports/jobs/external routes
- Fix all the bugs I found (lots)
- [x] Add tests for the signature auth layer

Test checklist:
- [ ] composer test

Ping @ushahidi/platform

### DIFF
--- a/app/Http/Middleware/SignatureAuth.php
+++ b/app/Http/Middleware/SignatureAuth.php
@@ -2,11 +2,19 @@
 
 namespace Ushahidi\App\Http\Middleware;
 
-use Closure;
 use Ushahidi\Core\Tool\Verifier;
+use Closure;
 
 class SignatureAuth
 {
+
+    protected $verifier;
+
+    public function __construct(Verifier $verifier)
+    {
+        $this->verifier = $verifier;
+    }
+
     /**
      * Handle an incoming request.
      *
@@ -22,9 +30,13 @@ class SignatureAuth
         $shared_secret = getenv('PLATFORM_SHARED_SECRET');
         $fullURL = $request->fullUrl();
 
-        $verifier = new Verifier($signature, $api_key, $shared_secret, $fullURL, $request->getInputSource()->all());
-
-        if (!$verifier->verified()) {
+        if (!$this->verifier->verified(
+            $signature,
+            $api_key,
+            $shared_secret,
+            $fullURL,
+            $request->getContent()
+        )) {
             abort(403, "Forbidden.");
         }
 
--- a/src/App/Repository/Form/AttributeRepository.php
+++ b/src/App/Repository/Form/AttributeRepository.php
@@ -246,20 +246,23 @@ class AttributeRepository extends OhanzeeRepository implements
 	 */
 	public function getByForms($form_ids)
     {
-		$sql = "SELECT DISTINCT
-				form_attributes.*,
-				form_stages.priority as form_stage_priority,
-				form_stages.form_id as form_id
-			FROM form_attributes
-			INNER JOIN form_stages ON form_attributes.form_stage_id = form_stages.form_id
-			INNER JOIN forms ON form_stages.form_id = forms.id
-			where forms.id IN :forms
-			ORDER BY form_stages.priority, form_attributes.priority
-		";
-		$results = DB::query(Database::SELECT, $sql)
-			->bind(':forms', $form_ids)
-			->execute($this->db);
-		$attributes = $results->as_array();
+        $attributes = [];
+        if (count($form_ids) > 0) { // @FIXME: how would empty form_id even happen?
+			$sql = "SELECT DISTINCT
+					form_attributes.*,
+					form_stages.priority as form_stage_priority,
+					form_stages.form_id as form_id
+				FROM form_attributes
+				INNER JOIN form_stages ON form_attributes.form_stage_id = form_stages.form_id
+				INNER JOIN forms ON form_stages.form_id = forms.id
+				where forms.id IN :forms
+				ORDER BY form_stages.priority, form_attributes.priority
+			";
+			$results = DB::query(Database::SELECT, $sql)
+				->bind(':forms', $form_ids)
+				->execute($this->db);
+			$attributes = $results->as_array();
+		}
 
 		$native = [
 			[
--- a/src/App/Repository/Form/AttributeRepository.php
+++ b/src/App/Repository/Form/AttributeRepository.php
@@ -11,6 +11,7 @@
 
 namespace Ushahidi\App\Repository\Form;
 
+use Ohanzee\DB;
 use Ohanzee\Database;
 use Ushahidi\Core\Entity;
 use Ushahidi\Core\SearchData;
--- a/src/Console/Command/PostExporter.php
+++ b/src/Console/Command/PostExporter.php
@@ -32,7 +32,6 @@ use Symfony\Component\Console\Output\OutputInterface;
 class PostExporter extends Command
 {
 
-	use UserContext;
 	use FormatterTrait;
 
 	private $data;
@@ -52,7 +51,7 @@ class PostExporter extends Command
      *
      * @var string
      */
-    protected $signature = 'export {--limit=100} {--offset=0}';
+    protected $signature = 'export {--limit=100} {--offset=0} {--job} {--include-header=1}';
 
     /**
      * The console command description.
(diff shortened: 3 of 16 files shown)


src/App/Repository/Form/AttributeRepository.php
```diff
@@ -245,20 +246,23 @@ public function getByForm($form_id)
...
-			INNER JOIN forms ON form_stages.form_id = forms.id
-			where forms.id IN :forms
-			ORDER BY form_stages.priority, form_attributes.priority
-		";
-		$results = DB::query(Database::SELECT, $sql)
-			->bind(':forms', $form_ids)
-			->execute($this->db);
-		$attributes = $results->as_array();
+        $attributes = [];
+        if (count($form_ids) > 0) { // @FIXME: how would empty form_id even happen?
```
rowasc: ohh. I guess this could happen if the is no data?

src/Console/Command/PostExporter.php
```diff
@@ -68,15 +67,17 @@ public function handle()
...
 
         // Construct a Search Data object to hold the search info
         $data = $this->data->get('search');
 
         // Get CLI params
 		$limit = $this->option('limit');
         $offset = $this->option('offset');
-        $job_id = $input->getOption('job');
-        $add_header = $input->getOption('include_header');
+        $job_id = $this->option('job');
```
rowasc: I think this has spaces instead of tabs

src/Console/Command/PostExporter.php
```diff
@@ -32,53 +32,54 @@
...
+	 * @var string
+	 */
+	protected $name = 'export';
+
+	/**
+	 * The console command signature.
+	 *
+	 * @var string
+	 */
+	protected $signature = 'export {--limit=100} {--offset=0} {--job} {--include-header=1}';
```
kinstelli: Are these meant to be defaults?

src/Init.php
```diff
@@ -386,7 +386,9 @@ function feature($name)
 
 // Tools
 $di->set('tool.signer', $di->lazyNew('Ushahidi\Core\Tool\Signer'));
-$di->set('tool.verifier', $di->lazyNew('Ushahidi\Core\Tool\Verifier'));
+$di->set('tool.verifier', $di->lazyNew('Ushahidi\Core\Tool\Verifier', [
+	'apiKeyRepo' => $di->lazyGet('repository.apikey')
```
kinstelli: Nice.

tests/unit/Core/Tool/VerifierTest.php
```diff
@@ -0,0 +1,119 @@
+<?php
+
+/**
+ * Unit tests for Lumen implementation of Ushahidi\Core\Tool\Mailer
```
kinstelli: `Verifier`, I think :)

tests/unit/App/Http/Middleware/SignatureAuthTest.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+/**
+ * Unit tests for Lumen implementation of Ushahidi\Core\Tool\Mailer
```
kinstelli: Also maybe outdated comment

tests/unit/Core/Tool/VerifierTest.php
```diff
@@ -0,0 +1,119 @@
...
+		$repo = M::mock(ApiKeyRepository::class);
+		$verifier = new Verifier($repo);
+
+		$repo->shouldReceive('apiKeyExists')->with($this->apiKey)->andReturn(true);
+
+		$return = $verifier->checkApiKey($this->apiKey);
+
+		$this->assertTrue($return);
+	}
+}
```
kinstelli: It might be helpful to add test case(s) where the payload is an empty array, such as in a GET with long querystring, which is apparently also a valid situation, but caused me some grief once.

src/Console/Command/PostExporter.php
rjmackay: Yup

rjmackay: Lumen command definitions are much easier

tests/unit/Core/Tool/VerifierTest.php
rjmackay: oops

src/Core/Tool/Signer.php
```diff
@@ -27,42 +27,19 @@ public function sign($fullUrl, $json)
...
         // Or http://pear.php.net/package/Crypt_HMAC/
         return base64_encode(hash_hmac("sha256", $url, $this->authToken, true));
     }
 
-    public function validate($expectedSignature, $url, $data = array())
+    public function validate($expectedSignature, $url, $data = "")
     {
-        return self::compare(
-            $this->computeSignature($url, json_encode($data)),
+        return hash_equals(
```
rjmackay: @willdoran I'd really like you to check this. I've dropped the DIY comparison function in favour of PHP's built int.
The docs mention:

> Both arguments must be of the same length to be compared successfully. When arguments of differing length are supplied, FALSE is returned immediately and the length of the known string may be leaked in case of a timing attack.

However from my reading of `compare()` below, it looks like it exited out early if strlen was different so would has a similar timing vulnerability.

tests/unit/Core/Tool/VerifierTest.php
rjmackay: Ok. Can do

tests/unit/Core/Tool/VerifierTest.php
```diff
@@ -0,0 +1,134 @@
...
+		$signature = $this->makeSig($this->sharedSecret, $url, json_encode($payload));
+
+		$repo->shouldReceive('apiKeyExists')->with($this->apiKey)->andReturn(true);
+
+		$return = $verifier->verified($signature, $this->apiKey, $this->sharedSecret, $url, json_encode($payload));
+
+		$this->assertTrue($return);
+	}
+
+	public function testValidSignatureWithEmptyPayload()
```
rjmackay: @kinstelli does this work?

tests/unit/Core/Tool/VerifierTest.php
```diff
@@ -0,0 +1,134 @@
...
+	{
+		$repo = M::mock(ApiKeyRepository::class);
+		$verifier = new Verifier($repo);
+		$url = "http://localhost:8000/api/v3/exports/external/count/1?api_key={$this->apiKey}";
+		$payload = "";
+		$signature = $this->makeSig($this->sharedSecret, $url, "");
```
kinstelli: My brain is dead now, but I think this actually fails if it's not a json_encoded empty array. 🤷‍♂️

rjmackay: So I changed the Signer to stop json_encoding everything... because thats weird. We now validate the raw contents of the request body as a string

src/Core/Tool/Signer.php
```diff
@@ -27,42 +27,19 @@ public function sign($fullUrl, $json)
...
         // Note: hash_hmac requires PHP 5 >= 5.1.2 or PECL hash:1.1-1.5
         // Or http://pear.php.net/package/Crypt_HMAC/
         return base64_encode(hash_hmac("sha256", $url, $this->authToken, true));
     }
 
-    public function validate($expectedSignature, $url, $data = array())
+    public function validate($expectedSignature, $url, $data = "")
     {
-        return self::compare(
-            $this->computeSignature($url, json_encode($data)),
```
rjmackay: @kinstelli note I dropped the `json_encode` here... so we expect request body as a string

app/Http/Middleware/SignatureAuth.php
```diff
@@ -22,9 +30,13 @@ public function handle($request, Closure $next)
...
 
-        $verifier = new Verifier($signature, $api_key, $shared_secret, $fullURL, $request->getInputSource()->all());
-
-        if (!$verifier->verified()) {
+        if (!$this->verifier->verified(
+            $signature,
+            $api_key,
+            $shared_secret,
+            $fullURL,
+            $request->getContent()
```
rjmackay: This is now the string request contents, not an array decoded from json

kinstelli: Oh, Missed that. That's good.

tests/unit/Core/Tool/VerifierTest.php
rowasc: tiny comment in case it helps: $data was by default an array (when doing the compute signature) so that's why we got the weird empty array appended to our urls even in empty get requests.

src/Core/Tool/Signer.php
willdoran: I think you're correct, the one below does have the same issue, I'm not sure why the note says that it doesn't. We can either go with the hash_equals and note the issue or we can implement one of the time insensitive versions.

Issue #2559, opened 2018-03-02
Setting the "Post date" on edit has no effect


tuxpiper:
### Expected behaviour
Editing the post date should persist

### Actual behaviour
Editing the post date doesn't persist

![2018-03-02 13 30 36](https://user-images.githubusercontent.com/467393/36899177-3a39a9ee-1e1e-11e8-8f0a-85dd0d7cbbcc.gif)

#### Is is ready for development when
- [ ] Has a description that includes purpose of this particular feature/improvement (how it will help the user and for which use cases)
- [ ] Can be delivered within a sprint (make sure issues are small pieces of work, not huge epics)
- [ ] Has all designs and UI assets available
- [ ] Has all (probable) dependencies identified
- [ ] Is estimated by the team
- [ ] Has acceptance criteria defined (do we want to do this before we send to dev? might help us identify dependencies/holes earlier?)

#### Test checklist
- Go to any post and click on 'edit'
- select a new post-date in the date-box in edit-mode
- save
- [ ] New date should be reflected in the post-metadata-header above the title, both on the post, in the post-list and in the post-popup in the map

##### Where
- [ ] [Local setup with Vagrant ](https://www.ushahidi.com/support/install-ushahidi#installing-for-development)
- [ ] [Local setup from platform-release ](https://www.ushahidi.com/support/install-ushahidi#installing-the-latest-release)
- [x] Ushahidi.io / SaaS solution 
- [ ] Ushahidi's QA environment
- [ ] Other (explain):

tuxpiper:
@caharding could we add this to the next cycle?

Celestine22:
@caharding @tuxpiper when is the expected release date for this? our customer would need this bug fixed and I need a timeline to feedback with.

caharding:
Devs are tackling this bug next, hopefully early this week

Celestine22:
Thanks @caharding

Celestine22:
@Angamanga @tuxpiper when should I test for this? I will update our customer after that. Please let me know, thanks.

tuxpiper:
I haven't seen any pull request relative to this, so I don't think anything got deployed yet @Celestine22

Angamanga:
@tuxpiper @Celestine22 Its in review, not merged or deployed yet.

Celestine22:
@kinstelli this is closed but has not been deployed. I would need to test this before we close it, so I'll re-open it for easy tracking.

Angamanga:
[...]

PR #2803, opened 2018-04-20
fix: replace interval format comparison with timestamp comparison

kinstelli:
- Updated datetime objects were not being compared correctly in StatefulData, which potentially affects a lot of parts of platform.
- The `if ($interval->format('F') > 0)` comparison wasn't using a format character of `%`, as in `format('%F')` 
- The PHP docs are also misleading, since AFAIK `%F` _only_ compares the microsecond portion of the DateTime objects, it doesn't actually reduce the date values to a microsecond format.  (See https://secure.php.net/manual/en/dateinterval.format.php)
- Instead of comparing with formats, it seems cleaner to just compare timestamps, so I updated it to use that. 

-*Note: We might want to tune this comparison to only compare above a certain threshold (i.e., ignore date comparisons unless minutes have changed -- `$timestamp_diff > (60 * 1000) )`.*

Test checklist:
- [ ] Run all existing automated tests, because this tiny fix potentially affects a lot of existing code
- [ ] Check that post dates are actually updated as desired.

Fixes ushahidi/platform#2559.

Ping @ushahidi/platform

### DIFF
--- a/src/Core/Traits/StatefulData.php
+++ b/src/Core/Traits/StatefulData.php
@@ -207,9 +207,13 @@ trait StatefulData
 			// Compare DateTime Objects
 			} elseif ($value instanceof \DateTimeInterface && $this->$key instanceof \DateTimeInterface) {
 				$current_key = $this->$key;
-				$interval = $value->diff($current_key);
 
-				if ($interval->format('F') > 0) {
+                $stored_date_ts = date_timestamp_get($current_key);
+                $received_date_ts = date_timestamp_get($value);
+                $timestamp_diff = abs($stored_date_ts - $received_date_ts);
+
+                // TODO: should we set a tolerance for how much variation is allowed in milliseconds?
+				if ($timestamp_diff > 0) {
 					// Update the value...
 					$this->setStateValue($key, $value);
 					// ... and track the change.
--- a/src/Core/Traits/StatefulData.php
+++ b/src/Core/Traits/StatefulData.php
@@ -208,8 +208,8 @@ trait StatefulData
 			} elseif ($value instanceof \DateTimeInterface && $this->$key instanceof \DateTimeInterface) {
 				$current_key = $this->$key;
 
-                $stored_date_ts = date_timestamp_get($current_key);
-                $received_date_ts = date_timestamp_get($value);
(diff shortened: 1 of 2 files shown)


Angamanga:
@kinstelli I tested this and it works when I update the date in the client. I tested other places where dates are used too. 💃 @rjmackay Although it looks and works good in the client, could you give this pr a review, I don't trust myself enough on the api to predict potential other effects this change might have. 

I found another thing related to this, the time is not set and displayed correctly. But this looks like its a formatting/timezone-error since its actually sending the values its displaying to the api + its also happening in the develop-branch. Looking into it on the client-side.

src/Core/Traits/StatefulData.php
```diff
@@ -207,9 +207,13 @@ public function setState(array $data)
 			// Compare DateTime Objects
 			} elseif ($value instanceof \DateTimeInterface && $this->$key instanceof \DateTimeInterface) {
 				$current_key = $this->$key;
-				$interval = $value->diff($current_key);
 
-				if ($interval->format('F') > 0) {
+                $stored_date_ts = date_timestamp_get($current_key);
```
rjmackay: Super minor but I'd prefer we use `$current_key->getTimestamp()` since we already know we have an object.

src/Core/Traits/StatefulData.php
```diff
@@ -207,9 +207,13 @@ public function setState(array $data)
...
+                $received_date_ts = date_timestamp_get($value);
+                $timestamp_diff = abs($stored_date_ts - $received_date_ts);
```
rjmackay: I was going to say you could use `$current_key->diff($value, true);` ... but I have no idea how you figure out from a DateInterval if there's actually a difference :(

Or according to the docs, you can just use comparison operators on dates
`$current_key == $value`. This makes me wonder why we were doing weird date formatting tricks to start with.

kinstelli: Yeah, I was wondering this too. I think this might be a habit from Ye Olden Times when comparing date objects with `==` was discouraged.

I can't find where I saw the code in platform, but I think there was a place where date changes are ignored unless they're at least a day difference. Or I imagined it.

PR #2819, opened 2018-04-21
WIP - Process pending csv jobs via console / cron

kinstelli:
- adds a console tool called `processexports` that can be run manually or configured with cron to process pending export jobs

For instance, this will run every minute via cron:
`* * * * *  cd  /var/www/platform ; ./bin/ushahidi processexports pending >/dev/null 2>&1`

TODO:
- Add client side code to:
    - remove expire time from export history, if there is none
    - ensure links work with stored paths
- Move tmp files to a proper path
  - And then, complete export paths using server side config
- test with large post counts, since this does not currently chunk exports, just runs in a separate process

Test checklist:
- [ ] Create a CSV export locally
- [ ] Run `./bin/ushahidi processexports pending`

### DIFF
--- a/application/classes/Ushahidi/Console/ProcessExportJobs.php
+++ b/application/classes/Ushahidi/Console/ProcessExportJobs.php
@@ -10,57 +10,26 @@
  */
 
 use Ushahidi\Console\Command;
-use Ushahidi\Core\Entity\DataProviderRepository;
-use Ushahidi\Core\Entity\PostExportRepository;
-use \Ushahidi\Core\Entity\FormAttributeRepository;
-
-use GuzzleHttp\Promise\Promise;
-
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Input\ArrayInput;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Output\BufferedOutput;
-
-use Ushahidi\Factory\DataFactory;
-use Ushahidi\Core\Traits\UserContext;
-use Ushahidi\Core\Tool\FormatterTrait;
-
 use Ushahidi\Core\Entity\ExportJobRepository;
 
 /* Simple console command that processes pending jobs in the DB */
 
 class Ushahidi_Console_ProcessExportJobs extends Command {
 
-    use UserContext;
-	use FormatterTrait;
-
     private $data;
-    private $postExportRepository;
     private $exportJobRepository;
-    private $formAttributeRepository;
 
 
     public function setExportJobRepo(ExportJobRepository $repo)
     {
         $this->exportJobRepository = $repo;
     }
-
-    public function setDataFactory(DataFactory $data)
-	{
-		$this->data = $data;
-	}
-
-	public function setPostExportRepo(PostExportRepository $repo)
-	{
-		$this->postExportRepository = $repo;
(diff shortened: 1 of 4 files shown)


application/classes/Ushahidi/Repository/Form/Attribute.php
```diff
@@ -242,16 +242,21 @@ public function getByForm($form_id)
...
-			"INNER JOIN form_stages ON form_attributes.form_stage_id = form_stages.form_id " .
-			"INNER JOIN forms ON form_stages.form_id = forms.id " .
-			"where forms.id IN :forms
-			ORDER BY form_stages.priority, form_attributes.priority";
-		$results = DB::query(Database::SELECT, $sql)
-			->bind(':forms', $form_ids)
-			->execute($this->db);
-		$attributes = $results->as_array();
+        $attributes = [];
+        if (count($form_ids) > 0)  // @FIXME: how would empty form_id even happen?
```
rjmackay: I don't know but I just hit the same bug in the lumen branch.

application/classes/Ushahidi/Console/ProcessExportJobs.php
```diff
@@ -0,0 +1,131 @@
...
+    {
+        $this->exportJobRepository = $repo;
+    }
+
+	protected function configure()
+	{
+		$this
+			->setName('processexports')
+			->setDescription('Processes pending export jobs.')
+			->addArgument('action', InputArgument::OPTIONAL, 'list, pending', 'pending')
```
rjmackay: Can you drop the argument and just use `execute` directly? That maps much better to the lumen approach

application/classes/Ushahidi/Console/ProcessExportJobs.php
```diff
@@ -0,0 +1,131 @@
...
+			;
+	}
+
+   //display all jobs, including pending, failed, and successful
+   protected function executeList(InputInterface $input, OutputInterface $output)
```
rjmackay: If we really need a list command - add it as a separate class and command called `export:list` (make this one `exports:run` or something)

application/classes/Ushahidi/Console/ProcessExportJobs.php
```diff
@@ -0,0 +1,131 @@
+<?php defined('SYSPATH') or die('No direct script access');
```
rjmackay: Could you create this under src/Console and set up the DI in src/Console/Init.php too?
If you do it here its a pain for lumen conflicts

src/Console/Command/ProcessExportJobs.php
```diff
@@ -0,0 +1,120 @@
...
+            if ($responseInfo['success'] == true) {
+                $resultStatus = 'SUCCESS';
+            } else {
+                $resultStatus = 'FAILED';
+            }
+        }
+
+        $jobEntity = $this->exportJobRepository->get($jobId);
+        // @TODO: get accessible path for this URL from config!
+         $accessiblePath = 'media/uploads/';
```
kinstelli: @rjmackay - Is there a recommended path for storing completed files, such that the client can easily access them. If we're not using Kohana, and this isn't running from a webserver, is there a best way to build this path?

rjmackay: In Lumen we use the laravel filesystem https://laravel.com/docs/5.6/filesystem or Flysystem directly. Does `Ushahidi\Core\Tool\Uploader::upload()` do what you need here?
If not you might have to use a temporary workaround, then refactor for lumen

PR #2820, opened 2018-04-23
refactor: Replace Ushahidi\Tool\Filesystem with Flysystem from Lumen

rjmackay:
- Remove Ushahidi\Tool\Filesystem intermediary
- Fetch filesystem instance direct from Lumen
- Use Uploader for deleting files too
- Update CreateMedia and DeleteCSV usecases to remove Filesystem dependency
- Add filesystem config
- Add symlink from httpdocs/storage to storage/app/public
- Remove CDN config

Test checklist:
- [x] bin/phpspec run
- [x] bin/behat tests/integration/csv.feature
- [x] bin/behat tests/integration/media.feature

Ping @ushahidi/platform


src/Core/Tool/Uploader.php
```diff
@@ -92,6 +92,17 @@ public function upload(UploadData $file, $filename = null)
 			]);
 	}
 
+	/**
+	 * Clean up file if upload failed
+	 * @param  $file  filepath
+	 */
+	public function delete($file)
```
rjmackay: Its plausible that now I've added this method, I should really rename this to `FileActions` or something. But I don't really care that much :)

app/Providers/AppServiceProvider.php
```diff
@@ -101,6 +102,14 @@ protected function configureAuraServices(\Aura\Di\ContainerInterface $di)
             'clientUrl' => $di->lazyGet('clienturl')
         ]));
 
+        // Configure filesystem
+        // The Ushahidi filesystem adapter returns a flysystem adapter for a given
+        // cdn type based on the provided configuration
+        $di->set('tool.filesystem', function () {
+            // Get the underlying League\Flysystem\Filesystem instance
+            return $this->app->make('filesystem')->disk()->getDriver();
```
rjmackay: Future music: We should define both public and private filesystem. CSV uploads could be private, but image uploads should be public.

config/filesystems.php
```diff
@@ -0,0 +1,69 @@
...
+    |
+    | Here you may configure as many filesystem "disks" as you wish, and you
+    | may even configure multiple disks of the same driver. Defaults have
+    | been setup for each driver as an example of the required options.
+    |
+    | Supported Drivers: "local", "ftp", "sftp", "s3", "rackspace"
+    |
+    */
+
+    'disks' => [
```
rowasc: is rackspace not a thing anymore?

rowasc:
@rjmackay the tests fail for me. But most tests also fail in the lumen branch. So it might be something in my setup. Not really sure.

config/filesystems.php
rjmackay: Um it is but this is just the default config file from laravel. I'll expand that with rackspace stuff. Oops

rjmackay:
> the tests fail for me. But most tests also fail in the lumen branch. So it might be something in my setup. Not really sure.

The full suite will fail. The specific tests called out above should pass I think

Issue #2485, opened 2018-02-20
Error on attempts to duplicate survey


aoduor:
### Expected behaviour
I should be able to successfully save a duplicated survey.

### Actual behaviour
When i click in duplicate, and try to save my new duplicate survey, I get this error message.

<img width="823" alt="screen shot 2018-02-20 at 11 00 54" src="https://user-images.githubusercontent.com/721918/36413091-f66376b6-162d-11e8-94d4-d21ab68f41a1.png">

Funny thing though, as much as this is what gets returned to me, the survey gets saved. 

### Steps to reproduce the behaviour/error
- Log in
- Click on settings -> surveys
- Duplicate a survey
- Attempt to save your duplicate survey

##### Where
- [ ] [Local setup with Vagrant ](https://www.ushahidi.com/support/install-ushahidi#installing-for-development)
- [ ] [Local setup from platform-release ](https://www.ushahidi.com/support/install-ushahidi#installing-the-latest-release)
- [x] Ushahidi.io / SaaS solution 
- [ ] Ushahidi's QA environment
- [ ] Other (explain): 

##### Test-checklist
This pull request makes the following changes:

In the default survey that is added to a deployment, the input is set to 'varchar'. This is not accepted by the validator, and all other title-fields are set to 'text'.
This pr adds a migration to update the set input = "text" for all form_attributes of type = "title".

Test checklist:

- Run migrations

- Go to db

- [ ]  input should be "text" for all form_attributes of type="title

- Go to client -> settings->surveys, click on "Duplicate survey" for the basic survey

- Add a title and save

- [ ]  You should be able to save the new, duplicated survey and not get an error-message

- [ ] repeat on another survey as well, you should be able to save the new duplicated survey and not get an error-message

  I certify that I ran my checklist

jshorland:
P0. broken.

Angamanga:
This can now be tested on http://targetedsurveytesting.v3-qa.ush.zone

crcommons:
@Angamanga could you add a test script, please! thanks!

Angamanga:
@crcommons Yep, checklist added! this is also deployed to production

Eve-wanderi:
@Angamanga                        Got a few issues on this.

1. Duplicate surveys are not visible on the Data mode: Once i duplicate a survey i cannot view this survey on data mode.
[...]

PR #2822, opened 2018-04-23
Adding migration to update input-type for titles

Angamanga:
- In the default survey that is added to a deployment, the input is set to 'varchar'. This is not accepted by the validator, and all other title-fields are set to 'text'. 

This pr adds a migration to update the set input = "text" for all form_attributes of type = "title". 

Test checklist:
- Run migrations
- Go to db
-  [ ] input should be "text" for all form_attributes of type="title
- Go to client -> settings->surveys, click on "Duplicate survey" for the basic survey
- Add a title and save
- [ ] You should be able to save the new, duplicated survey and not get an error-message
- repeat on another survey as well

Fixes ushahidi/platform#2485 .

Ping @ushahidi/platform

### DIFF
--- a/migrations/20180423131108_change_title_type_to_text.php
+++ b/migrations/20180423131108_change_title_type_to_text.php
@@ -1,0 +1,23 @@
+<?php
+
+use Phinx\Migration\AbstractMigration;
+
+class ChangeTitleTypeToText extends AbstractMigration
+{
+    public function up()
+    {
+        $pdo = $this->getAdapter()->getConnection();
+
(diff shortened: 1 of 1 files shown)


rjmackay:
> In the default survey that is added to a deployment, the input is set to 'text'. This is not accepted by the validator, and all other title-fields are set to 'text'.

Huh? I'm not sure what the other fields were set to? text or varchar?

Angamanga:
Updated description, I meant "input is set to varchar"

Issue #2797, opened 2018-04-19
If targeted survey has no more outgoing questions, server responds with Error 400 to last incoming message


kinstelli:
### Expected behaviour
- HTTP should respond with 200, then just not send another message to the recipient

### Actual behaviour
- HTTP responds with 400
```
{
    "errors": [
        {
            "status": 400,
            "title": "Outgoing question not found for contact 2 and form 3",
            "message": "Outgoing question not found for contact 2 and form 3"
        }
    ]
}
```

### Steps to reproduce the behaviour/error
- Create a targeted survey
- Send messages to the server via Postman to an SMS endpoint
- When no more questions remain, observe the HTTP response from the server

#### Where does this issue happen?
- [x] [Local setup with Vagrant ](https://www.ushahidi.com/support/install-ushahidi#installing-for-development)
- [ ] [Local setup from platform-release ](https://www.ushahidi.com/support/install-ushahidi#installing-the-latest-release)
- [ ] Ushahidi.io / SaaS solution 
- [x] Ushahidi's QA environment
- [ ] Other (explain):

kinstelli:
@rowasc -- Would you have a chance to confirm this issue?
Looking at the code, I think it's throwing an exception when there might not be any more questions remaining to be sent, but I'm not sure.

rowasc:
@kinstelli did this have the latest code merged in? With the fixes for the message-receive to allow messages even if there's no survey, this wouldn't happen

rowasc:
@kinstelli yea just checked. They are in `develop` :/  I was hoping there was a weird merge issue

rowasc:
ugh trying to repro locally does not result in this bug :( checking qa now..

rowasc:
@kinstelli ok so this happens when your contact WAS in a different survey before, as Mark suspected. 
The issue is in 
>		$surveyStateEntity = $this->targeted_survey_state_repo->getByContactId($contact_id);

where we just select one and not the one that's active

rowasc:
@kinstelli @StaicyG @Eve-wanderi 
 This specific issue is solved now. 
@kinstelli because of https://app.zenhub.com/workspace/o/ushahidi/platform/issues/2795 if you use the test service you might run into issues. For instance , if you add numbers with + before country code (+59891222333) and send a message with test service, all is lost because it seems test service is stripping the + from country code at some point.  If when you add your number, you use the number without the country code, it finds the contact correctly because the test sms service is not doing anything weird to it. 
[...]

PR #2832, opened 2018-04-24
refactor: Cleaning up ReceiveMessage usecase

rjmackay:
- Ensure message is always saved
- Remove HTTP errors. They should never be in a usecase
- Remove unneeded validation
- Tidy code
- Add way more comments as I figure out whats going on

### Test script

With postman (added collection below) or with platform-client: 

- Send a message to the platform with a contact you haven't used in a targeted survey. it should be in messages but not be associated with a targeted survey. 
- Create first targeted survey (TS1) with two or more attributes, and at least 2 contacts
- After the first set of questions is sent:
    - Send a message from one of the contacts (ie using testservice) to answer the first question
    - Repeat with the same contact until its targeted_survey_state record is in survey_status "Survey finished" in the database 
    - Send another message now that it finished the survey. You should see that message in `messages` and it should not be assigned to the same post_id as the messages related to the survey
    - Send a message from the second contact to answer their first question. Do no finish the survey for this contact
- Create second targeted survey (TS2) with two or more attributes, and at least 2 contacts that are the same you used before (in TS1)
- Verify that in targeted_survey_state, the record for the contact that didn't complete the survey in TS1 reads "ACTIVE CONTACT IN SURVEY <surveyId>" where surveyId is the survey id of the first targeted survey you created
- After the first set of questions is sent, with your first contact: 
    - Send a message from one of the contacts (ie using testservice) to answer the first question
    - Verify that targeted survey has the correct attribute set affter you answer (next question's id)
    - Complete the survey with this contact. Verify that it says "Survey finished" in targeted_survey_state `survey_status` field. 
    - Send another message now that it finished the survey. You should see that message in `messages` and it should not be assigned to the same post_id as the messages related to the survey
- Send a message to the platform with a contact you haven't used in a targeted survey. it should be in messages but not be associated with a targeted survey. 

[Test targeted surveys.postman_collection.txt](https://github.com/ushahidi/platform/files/1949224/Test.targeted.surveys.postman_collection.txt)
 

Ported from #2792 
Fixes ushahidi/platform#2797

Ping @rowasc @kinstelli

### DIFF
--- a/src/Core/Usecase/Message/ReceiveMessage.php
+++ b/src/Core/Usecase/Message/ReceiveMessage.php
@@ -174,7 +174,9 @@ class ReceiveMessage extends CreateUsecase
 				]
 			);
 
-			// @todo Should we actually create a new post if this happens?!
+			// Create a new post
+			$post_id = $this->createPost($incoming_message);
+			$incoming_message->setState(compact('post_id'));
 
 			// But always save the message anyway - otherwise its lost forever
 			return $this->repo->create($incoming_message);
@@ -274,15 +276,11 @@ class ReceiveMessage extends CreateUsecase
 			// @todo decouple this by moving to a listener
 			$id = $this->createTargetedSurveyMessages($entity);
 		} else {
-			$post_id = null;
-			// don't throw an event
 			// ... create post for message
 			// @todo decouple this by moving to a listener
 			$post_id = $this->createPost($entity);
+			$entity->setState(compact('post_id'));
 			// ... persist the new message entity
-			if ($post_id) {
-				$entity->setState(compact('post_id'));
-			}
 			$id = $this->repo->create($entity);
 		}
 		return $id;
--- a/src/Core/Usecase/Message/ReceiveMessage.php
+++ b/src/Core/Usecase/Message/ReceiveMessage.php
@@ -114,7 +114,7 @@ class ReceiveMessage extends CreateUsecase
 	{
 		// Update state with post id
 		$incoming_message->setState(['post_id' => $survey_state_entity->post_id]);
-		$incomingMessageId = $this->repo->create($incomingMessage);
+		$incomingMessageId = $this->repo->create($incoming_message);
 
 		return $incomingMessageId;
 	}
@@ -163,6 +163,7 @@ class ReceiveMessage extends CreateUsecase
 		$messageInSurveyState = $this->repo->get($surveyStateEntity->message_id);
 
 		// If we didn't find an outgoing message
+		// I'm not convinced this is right - if someone sends multiple incoming messages before the next outgoing one. But the survey isn't finished... shouldn't we add the messages to the original post?
 		if (!$messageInSurveyState || $messageInSurveyState->direction !== \Ushahidi\Core\Entity\Message::OUTGOING) {
 			// We can't save it as a message of the survey
 			// ... log an error because we should probably never end up here
--- a/src/Core/Usecase/Message/ReceiveMessage.php
+++ b/src/Core/Usecase/Message/ReceiveMessage.php
@@ -166,7 +166,8 @@ class ReceiveMessage extends CreateUsecase
 		if (!$messageInSurveyState || $messageInSurveyState->direction !== \Ushahidi\Core\Entity\Message::OUTGOING) {
 			// We can't save it as a message of the survey
 			// ... log an error because we should probably never end up here
-			\Kohana::$log->add(\Log::ERROR,
+			\Kohana::$log->add(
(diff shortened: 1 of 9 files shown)


rjmackay:
@rowasc @kinstelli I don't know if this actually fixes anything yet. I need to write some tests before it lands... but it rebased perfectly 🎉

rjmackay:
@rowasc @kinstelli I wrote another spec test. I found a minor bug however other than that, I'm not sure I proved anything. Spec tests are awkward for hitting all the edge cases, and mostly just showed that we need to decouple the code more.

src/Core/Usecase/Message/ReceiveMessage.php
```diff
@@ -163,6 +163,7 @@ private function createTargetedSurveyMessages($incoming_message)
 		$messageInSurveyState = $this->repo->get($surveyStateEntity->message_id);
 
 		// If we didn't find an outgoing message
+		// I'm not convinced this is right - if someone sends multiple incoming messages before the next outgoing one. But the survey isn't finished... shouldn't we add the messages to the original post?
```
rowasc: mmm. I think this might be weird if we add it to the contact's active post for the survey since it'd be linked to the survey?

src/Core/Usecase/Message/ReceiveMessage.php
```diff
@@ -111,118 +107,142 @@ public function setTargetedSurveyStateRepo(Entity\TargetedSurveyStateRepository
...
-		// ... attempt to load the entity
-		$messageInSurveyState = $messageInSurveyState->get($surveyStateEntity->message_id);
+		// Load the survey state for this contact
+		$surveyStateEntity = $this->targeted_survey_state_repo->getByContactId($incoming_message->contact_id);
+
+		// Attempt to load the previous message
+		$messageInSurveyState = $this->repo->get($surveyStateEntity->message_id);
+
+		// If we didn't find an outgoing message
+		// I'm not convinced this is right - if someone sends multiple incoming messages before the next outgoing one. But the survey isn't finished... shouldn't we add the messages to the original post?
```
rowasc: mmm. I think this might be weird if we add it to the contact's active post for the survey, since it'd be linked to the survey but not be a response to a question

(re-adding because I initially added the comment in a commit)

rjmackay:
Updated to merge from develop and fix tests

src/Core/Usecase/Message/ReceiveMessage.php
```diff
@@ -145,89 +125,129 @@ private function createIncomingMessage($incoming_message, $contact_id, $survey_s
...
 	 * @param $next_form_attribute
-	 * @return int|$incomingMessageId
-	 * @throws HTTP_Exception_400
+	 * @return int|$outgoingMessageId
 	 */
 	private function createOutgoingMessage($contact_id, $survey_state_entity, $next_form_attribute)
 	{
         // @FIXME: Message type should be configurable per deployment,survey
         $message_type = 'sms';
-        $data_provider = \DataProvider::getEnabledProviderForType($message_type);
```
rjmackay: This was breaking phpspec since I added a more thorough spec :(
I moved this handling into data providers themselves so that unassigned messages get sent via the default source

src/Core/Usecase/Message/ReceiveMessage.php
```diff
@@ -145,89 +125,131 @@ private function createIncomingMessage($incoming_message, $contact_id, $survey_s
...
+
+		// If we have another question to send
 		if ($next_form_attribute->getId() > 0) {
-			$newMessageId = $this->createOutgoingMessage($contact_id, $surveyStateEntity, $next_form_attribute);
+			// Queue next question to be sent
+			$outgoingMessageId = $this->createOutgoingMessage(
+				$incoming_message->contact_id,
+				$surveyStateEntity,
+				$next_form_attribute,
+				$incoming_message->data_provider
```
rjmackay: We're now using the incoming data provider when sending the outgoing. ie. if we receive a message from Twilio, always reply with Twilio

rjmackay:
@kinstelli you might want to double check my changes in the last couple of commits on this branch
I've merged anyway because this was blocking for lumen work

src/Core/Usecase/Message/ReceiveMessage.php
kinstelli: Yeah, I like that better. I was thinking of doing the same thing, but was also debating about specifying SMS provider for each survey, as described in my first comment [here](https://github.com/ushahidi/platform/issues/2842), though it potentially breaks a lot of things and requires a migration to `forms`.

Issue #1617, opened 2017-03-06
USH-022 — Role Creation and Permission Assignment


willdoran:
### Expected behaviour
- Verify if a user is a paid or non-paid user.
### Actual behaviour
- It is possible to create roles and make permission assignment in the FREE version, that feature only would be available on RESPONDER version. With this request is possible to create the user and then in the platform is possible to make the assignment of permissions.

caharding:
I'm sorry I don't understand this issue as written

rowasc:
@willdoran is this in the right column?

willdoran:
Yeah

rowasc:
ah ok. @willdoran is it blocked by lumen (to deploy)?

rjmackay:
Dropped back to QA to test the rebased version

PR #2851, opened 2018-04-26
Adding fix for Role feature state check

willdoran:
### Notes
I'm not certain how to test this in the integration tests since we can't alter the features via request and I don't immediately see a way to do it from the tests.

- Fixes case where users could create a role on any plan other than Responder

Test checklist:
- [ ] With Roles feature enabled set to false
- [ ] Attempt to create/update a new role via HTTP Post
- [ ] You should receive a 422

Fixes ushahidi/platform#1617

Ping @ushahidi/platform


rjmackay:
> I'm not certain how to test this in the integration tests since we can't alter the features via request and I don't immediately see a way to do it from the tests.

@willdoran if you use the `@rolesEnabled` tag in behat tests does that work? https://github.com/ushahidi/platform/blob/develop/tests/integration/bootstrap/FeatureContext.php#L43

src/App/Validator/Role/Update.php
```diff
@@ -36,6 +38,15 @@ protected function getRules()
 		];
 	}
 
+	public function checkRolesEnabled(\Kohana\Validation\Validation $validation, $permissions)
```
rjmackay: Couple of things:
1. Don't you need to add this into `getRules` above?
2. Should updates fail entirely? or should we be able to rename roles in the free tier?

willdoran:
@rjmackay Yeah I think that will work, will add it.

src/App/Validator/Role/Update.php
rjmackay: Just realised this is going back to the lumen branch. You need to change the interface to rely on `Ushahidi\Core\Tool\ValidationEngine` instead. It should work as written otherwise

willdoran: > Should updates fail entirely? or should we be able to rename roles in the free tier?
nope no changes to roles at all allowed

>Don't you need to add this into getRules above?
yep

willdoran: > You need to change the interface to rely on Ushahidi\Core\Tool\ValidationEngine instead.
Can you link an example from the other validators, not sure I follow.

willdoran:
@rjmackay could you peak at this again when you have a second?

tests/unit/App/Validator/Role/UpdateTest.php
```diff
@@ -0,0 +1,48 @@
...
+use Mockery as M;
+use Ushahidi\Core\Entity\PermissionRepository;
+
+/**
+ * @backupGlobals disabled
+ * @preserveGlobalState disabled
+ */
+class RoleUpdateTest extends TestCase
+{
+
```
rjmackay: Ideally we should add a test that actually calls Update::check() with some data too.

PR #2857, opened 2018-04-26
Bugfix/2838

rowasc:
- Filter out fields that are not in the selected list of fields for an export ( when there are selected fields)
- Add all fields even if they are filtered out. 
- Add "post status" to headers
- This depends on a platform-client PR that sends attribute keys instead of attribute ids.  https://github.com/ushahidi/platform-client/pull/1121
- Fixes issue where the join for attribute headers would return an empty set since it was matching form_id and form_stage_id fields. So basically it'd look like it was working only if you happened to have the same form_stage ids and form_ids by chance ...which is gross. 

### Pre-requisites
-  Use this PR's branch in the client https://github.com/ushahidi/platform-client/pull/1121
-  Have two surveys with some data. Make sure that for both of the surveys, the stage_id is not the same as the form_id

## Test 1 - no fields selected
- [x] Create an export job without selecting any fields or filters
- [x] Download the job. You should see all data and all available headers for both surveys. 

## Test 1.1 - no fields selected, with filters
- [x] Create an export job selecting some filters in data mode
- [x] Download the job. You should see the same number of posts that you had in the data mode list, and all available headers for both surveys. 

## Test 2 - two survey specific fields selected
- [x] Create an export job with the following requirements: 
    - [x] no filters
    - [x] one field of survey 1 selected
    - [x] one field of survey 2 selected
- [x] Download the job. 
- [x] You should see all posts for both surveys
- [x] You should see the following "base" headers  Post ID | Post Status | Created (UTC) | Updated (UTC) | Post Date (UTC) | Contact ID | Contact | Sets 
- [x] You should see the headers for the two fields you selected after the list of base headers
- [x] if the fields have values, those values should be in the correct field

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/application/classes/Ushahidi/Repository/Form/Attribute.php
+++ b/application/classes/Ushahidi/Repository/Form/Attribute.php
@@ -248,7 +248,7 @@ class Ushahidi_Repository_Form_Attribute extends Ushahidi_Repository implements
 			"INNER JOIN forms ON form_stages.form_id = forms.id ";
 		if (!empty($include_attributes))
 		{
-			$sql .= " AND form_attributes.id IN :form_attributes ";
+			$sql .= " AND form_attributes.key IN :form_attributes ";
 		}
 		$sql .= "ORDER BY form_stages.priority, form_attributes.priority ";
 		$results = DB::query(Database::SELECT, $sql)
--- a/application/classes/Ushahidi/Repository/Form/Attribute.php
+++ b/application/classes/Ushahidi/Repository/Form/Attribute.php
@@ -241,7 +241,8 @@ class Ushahidi_Repository_Form_Attribute extends Ushahidi_Repository implements
 	 * that does not directly depend on the rows we are fetching at the time but on the
 	 * list of form ids that match a specific query
 	 */
-	public function getByForms($form_ids, $include_attributes = []) {
+	public function getByForms($form_ids, array $include_attributes = null)
+	{
 		$sql = "SELECT DISTINCT form_attributes.*, form_stages.priority as form_stage_priority, form_stages.form_id as form_id " .
 			"FROM form_attributes " .
 			"INNER JOIN form_stages ON form_attributes.form_stage_id = form_stages.id " .
--- a/application/classes/Ushahidi/Repository/Form/Attribute.php
+++ b/application/classes/Ushahidi/Repository/Form/Attribute.php
@@ -244,13 +244,13 @@ class Ushahidi_Repository_Form_Attribute extends Ushahidi_Repository implements
 	public function getByForms($form_ids, $include_attributes = []) {
 		$sql = "SELECT DISTINCT form_attributes.*, form_stages.priority as form_stage_priority, form_stages.form_id as form_id " .
 			"FROM form_attributes " .
-			"INNER JOIN form_stages ON form_attributes.form_stage_id = form_stages.form_id " .
-			"INNER JOIN forms ON form_stages.form_id = forms.id ";
+			"INNER JOIN form_stages ON form_attributes.form_stage_id = form_stages.id " .
+			"INNER JOIN forms ON form_stages.  = forms.id ";
 		if (!empty($include_attributes))
 		{
-			$sql .= " AND form_attributes.id IN :form_attributes";
+			$sql .= " AND form_attributes.id IN :form_attributes ";
 		}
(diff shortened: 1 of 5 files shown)


rowasc:
@rjmackay can you review (on your monday)? I'm not sure why travis is failing :/  this fixes a bunch of issues related to 2838. 

In the strange scenario where this is approved, please land it to QA so Stacy can test when she starts her day on monday. Thanks!

migrations/20180427203643_export_job_field_size.php
```diff
@@ -0,0 +1,62 @@
...
+				['null' => true, 'limit' => 16777215, 'default' => null]
+			)
+			->changeColumn(
+				'filters',
+				'text',
+				['null' => true, 'limit' => 16777215, 'default' => null]
+			)
+			->update();
+    }
+    public function down()
```
rjmackay: Its failing because something is wrong with your migration. Have you tested downgrading to verify it works? 
Does it work with the test data in the DB?

rjmackay: >   SQLSTATE[01000]: Warning: 1265 Data truncated for column 'fields' at row 1

application/classes/Ushahidi/Repository/Form/Attribute.php
```diff
@@ -241,28 +241,41 @@ public function getByForm($form_id)
 	 * that does not directly depend on the rows we are fetching at the time but on the
 	 * list of form ids that match a specific query
 	 */
-	public function getByForms($form_ids) {
+	public function getByForms($form_ids, $include_attributes = []) {
 		$sql = "SELECT DISTINCT form_attributes.*, form_stages.priority as form_stage_priority, form_stages.form_id as form_id " .
```
rjmackay: FYI - we should really really be using the query builder for this. Don't refactor it now, but also try not to write new code like this.

application/classes/Ushahidi/Repository/Form/Attribute.php
```diff
@@ -241,28 +241,41 @@ public function getByForm($form_id)
 	 * that does not directly depend on the rows we are fetching at the time but on the
 	 * list of form ids that match a specific query
 	 */
-	public function getByForms($form_ids) {
+	public function getByForms($form_ids, $include_attributes = []) {
```
rjmackay: Could you type hint this as an array?

migrations/20180427203643_export_job_field_size.php
```diff
@@ -0,0 +1,54 @@
...
+			->changeColumn(
+				'filters',
+				'text',
+				['null' => true, 'limit' => MysqlAdapter::TEXT_MEDIUM, 'default' => null]
+			)
+			->update();
+    }
+    public function down()
+	{
+		// Noop - don't rollback because it would truncate use data
```
rjmackay: Changed this to fix rolling back migrations

migrations/20180427203643_export_job_field_size.php
```diff
@@ -0,0 +1,54 @@
...
+		 * Changing fields and filters to the MEDIUM_TEXT
+		 * length. We would not (always) be able to handle
+		 * very large deployments with many fields
+		 * with a 255 varchar.
+		 */
+		$this->table('export_job')
+			->changeColumn(
+				'fields',
+				'text',
+				['null' => true, 'limit' => MysqlAdapter::TEXT_MEDIUM, 'default' => null]
```
rjmackay: Using `MysqlAdapter::TEXT_MEDIUM` since I discovered you can do this while reading phinx docs :)
Much nicer

rjmackay:
> Add all fields even if they are filtered out.

^ I don't quite understand what you mean?

application/classes/Ushahidi/Repository/Form/Attribute.php
rjmackay: Fixed

migrations/20180427203643_export_job_field_size.php
rowasc: yea I didn't use it specifically because I didn't know if we were "allowed" to use the reference to the adapter here directly ...

rowasc:
@willdoran can you review?

migrations/20180427203643_export_job_field_size.php
willdoran: @rowasc Can you add this?

PR #2896, opened 2018-05-02
Adding ratelimiter for signups

willdoran:
- Adding ratelimiter for registration

Test checklist:
- [ ] You should not be able to make more than 3 registration attempts per minute

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/src/App/Init.php
+++ b/src/App/Init.php
@@ -47,32 +47,66 @@ $di->set('memcached', $di->lazy(function () use ($di) {
 	return $memcached;
 }));
 
+// Set up register rate limiter
+$di->set('ratelimiter.register.flap', $di->lazyNew(
+	'BehEh\Flaps\Flap',
+	array(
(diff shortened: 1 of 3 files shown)


src/App/Init.php
```diff
@@ -47,32 +47,66 @@
 	return $memcached;
 }));
 
-// Set up login rate limiter
-$di->set('ratelimiter.login.flap', $di->lazyNew('BehEh\Flaps\Flap'));
+// Set up register rate limiter
+$di->set('ratelimiter.register.flap', $di->lazyNew(
+	'BehEh\Flaps\Flap',
+	array(
```
rjmackay: Can we always use `[]` style arrays? not `array()`?
I need to update linting rules to enforce that

rjmackay:
@willdoran this is just waiting for you to land it

PR #2901, opened 2018-05-03
Fixes for post exporter cli in lumen 

rowasc:
- fix: ob_get_clean needs to have an ob_get_length to make sure it doesn't fail in more strict error reporting settings
- fix: export --job call from CLI should be called with job id. Job id should be a required argument for post exporter, not an option 

Test checklist:
- [ ] run ` bin/phpunit --no-coverage`
- [ ] run  `php artisan export --limit=200 --offset=0 --include-header=1 1` (replace the last `1` with a job ib) => this should work
- [ ] run  `php artisan export --limit=200 --offset=0 --include-header=1`=> since it does not have a job_id, this should fail 

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/Http/Controllers/API/Exports/External/CliController.php
+++ b/app/Http/Controllers/API/Exports/External/CliController.php
@@ -28,15 +28,15 @@ class CliController extends RESTController
 
     public function show(Request $request)
     {
+    	$route_params = $this->getRouteParams($request);
         // this is a trick to convert 'false' to falsy (which would be true),
         // 'true' to true, and an unset param to false
         $include_header = json_decode($request->input('include_header', 1)) == true ? 1 : 0;
-
-        // Run export command
+		// Run export command
         $exitCode = Artisan::call('export', [
+			'job' => $route_params['id'],
             '--limit' => $request->input('limit', 0),
             '--offset' => $request->input('offset', 0),
-            '--job' => $request->input('id'),
             '--include-header' => $include_header,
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/API/Exports/External/CliController.php
```diff
@@ -28,15 +28,15 @@ protected function getResource()
 
     public function show(Request $request)
     {
+    	$route_params = $this->getRouteParams($request);
```
rowasc: I tried using `show(Request $request, $id)` but that didn't work :/ not sure why tbh

src/Console/Command/PostExporter.php
```diff
@@ -88,20 +88,19 @@ public function handle()
...
-		if ($job_id) {
-			// Load the export job
-			$job = $this->exportJobRepository->get($job_id);
 
-			$this->session->setUser($job->user_id);
-			// Merge the export job filters with the base filters
-			if ($job->filters) {
-				$filters = array_merge($filters, $job->filters);
-			}
+		// Load the export job
```
rowasc: Job_id is a mandatory parameter in the cli now. It'll fail if you don't provide it

rowasc: This was a bug hidden by the fact that we always send a job id . But it was bad anyway, and was hiding a bug in CliController where id was null

Issue #2879, opened 2018-05-01
 [Platform API] Refactor CSV Export into a usecase that we can extend to support HDX export


rowasc:
### TODO
- [ ] src/Console/Command/PostExporter's handle should only deal with getting params and calling the ExportPost usecase
- [ ] an ExportPost use case should be created to handle: 
    - Getting posts with correct permissions
    - Creating header rows
    - Adding columns with data
    - Running validator and formatter
    - Returning the result
- [ ] Add tests to validate the Post Exporter results

Acceptance criteria:
- [x] the post exporter works from the CLI as before 
- [x] generating a CSV without the refactor and with the refactored usecase results in the same output (minus any bugs fixed and documented of course)
- [x] the post exporter works when called with the lumen queues repo as before
- [x] usage examples are documented

### Usage: 
Call the exporter with : 
php artisan export --limit={limit} --offset={offset} --include-header={1|0} {jobid}

offset/limit : a number, the offset and limit of rows to fetch
include-header: 1 or 0 . 1 to include it, 0 to not include it. It's used to be able to generate chunks of a csv without headers and then join them later with the first chunk with header
jobId: the export job id 

#### branch in platform api
feature/2879

PR #2910, opened 2018-05-04
Feature/2879 - refactor of console exporter to usecase

rowasc:
- Moved console exporter code to usecase
- Added a couple of tests. Once testing the right methods are called from the usecase and the other testing the CSV formatter itself for field order and contents
- Brings fixes from https://github.com/ushahidi/platform/pull/2908 to lumen
Test checklist:
- [ ]

Fixes ushahidi/platform#2879 .

Ping @ushahidi/platform


src/App/Repository/Form/AttributeRepository.php
```diff
@@ -276,6 +271,16 @@ public function getByForms($form_ids, array $include_attributes = null)
...
 			],
+			[
+				'label' => 'Survey',
+				'key' => 'form_name',
+				'type' => 'form_name',
+				'input' => 'text',
+				'form_id' => 0,
+				'form_stage_id' => 0,
+				'form_stage_priority' => 0,
+				'priority' => 2
```
willdoran: I think Priority for all these attributes should be 0 otherwise they will always move around amongst the custom attributes, What do you think?

src/Console/Command/PostExporter.php
```diff
@@ -59,93 +41,33 @@ class PostExporter extends Command
...
-		 */
-		$file = $formatter($posts, $keyAttributes);
+		return $this->usecase;
+	}
 
-		$response = [
-			[
-				'file' => $file->file,
-			]
+	public function handle()
```
willdoran: Much shorter, very great.

src/Core/Usecase/Post/Export.php
```diff
@@ -11,69 +11,153 @@
...
+
+
+	// provides setFormatter to assign the formatter the usecase will use
+	use FormatterTrait;
+
 	// - VerifyParentLoaded for checking that the parent exists
 	use VerifyParentLoaded;
 
+	public function setExportJobRepository(ExportJobRepository $repo)
+    {
```
willdoran: something weird with this indent or is it github?

src/Core/Usecase/Post/Export.php
```diff
@@ -11,69 +11,153 @@
...
+		$this->exportJobRepository = $repo;//service('repository.export_job');
+	}
+
+	public function setFormAttributeRepository(AttributeRepository $repo)
+    {
```
willdoran: same

src/Core/Usecase/Post/Export.php
```diff
@@ -11,69 +11,153 @@
...
+		$this->formAttributeRepository = $repo; //service('repository.form_attribute');
+	}
+
+	public function setPostExportRepository(ExportRepository $repo)
+    {
```
willdoran: same

src/Core/Usecase/Post/Export.php
```diff
@@ -11,69 +11,153 @@
...
 	 */
-	protected function getPagingFields()
-	{
-		return [
-			'orderby' => 'created',
-			'order'   => 'asc',
-			'limit'   => null,
-			'offset'  => 0
+	public function constructFilters($payload, $job_filters = null)
+    {
```
willdoran: same

src/Core/Usecase/Post/Export.php
```diff
@@ -11,69 +11,153 @@
...
+	}
+
+	/**
+	 * @param $job
+	 * @param $filters
+	 * @return mixed
+	 * Construct a Search Data object to hold the search info
+	 */
+	public function constructSearchData($job, $filters)
+    {
```
willdoran: same

willdoran: same

src/App/Repository/Form/AttributeRepository.php
rowasc: ah. No because we use the stage priority and form id as well to sort. So a stage priority 0 with priority 2 is still lower than custom attributes, and will always appear first

src/Core/Usecase/Post/Export.php
rjmackay: I clearly need to fix the linters indenting and reformat all the code to match. We have mixed spaces and tabs in too many places

tests/unit/Core/Usecase/Post/ExportTest.php
```diff
@@ -0,0 +1,117 @@
...
+    {
+		// set CLI params to be the payload for the usecase
+		$payload = [
+			'job_id' => 1,
+			'limit' => 100,
+			'offset' => 0,
+			'add_header' => true,
+		];
+
+		$post1 = new MockPostEntity();
```
rjmackay: In most cases its much easier and just as effective to just use the Entity as-is rather than mocking it. ie. if you just use `new Post()` here instead, does that just work?

src/Console/Command/PostExporter.php
```diff
@@ -59,93 +41,33 @@ class PostExporter extends Command
...
 		];
 
-		$this->line(json_encode($response));
+		// Get the usecase and pass in authorizer, payload and transformer
+		$file  = $this->getUsecase()
+			->setPayload($payload)
+			->setAuthorizer(service('authorizer.export_job'))
+			->setFormatter(service('formatter.entity.post.csv'))
+			->interact();
+		$this->line(json_encode($file));
```
rjmackay: We can just call the usecase directly in the controller now right? and then stop returning JSON from a CLI command I think

rowasc: Oh. I think that's correct yea :D I'll try

rowasc:
@rjmackay @willdoran can you review again?thanks!

app/Http/Controllers/API/Exports/External/CliController.php
```diff
@@ -26,29 +26,35 @@ protected function getResource()
         return 'export_jobs';
     }
 
+	protected function getUsecase()
+	{
+		if (!$this->usecase) {
+			// @todo inject
+			$this->usecase = service('factory.usecase')
```
rjmackay: usecase factory is already injected, use `this->usecaseFactory`

src/App/Formatter/Post/CSV.php
```diff
@@ -103,11 +103,9 @@ public function createHeading($attributes)
 	 *
 	 * @return array
 	 */
-	protected function generateCSVRecords($records, $attributes)
+	public function generateCSVRecords($records, $attributes)
```
rjmackay: Why public?

src/Console/Command/PostExporter.php
```diff
@@ -59,95 +41,32 @@ class PostExporter extends Command
...
+			'limit' => $this->option('limit'),
+			'offset' => $this->option('offset'),
+			'add_header' => $this->option('include-header'),
 		];
 
-		$this->line(json_encode($response));
+		// Get the usecase and pass in authorizer, payload and transformer
+		$file  = $this->getUsecase()
+			->setPayload($payload)
+			->setAuthorizer(service('authorizer.export_job'))
```
rjmackay: I think setAuthorizer and setFormatter should all happen in `getUsecase` or update the DI  setup so they're set correctly by the usecase factory first

app/Http/Controllers/API/Exports/External/CliController.php
```diff
@@ -26,29 +26,35 @@ protected function getResource()
...
+		// set CLI params to be the payload for the usecase
+		$payload = [
+			'job_id' => $route_params['id'],
+			'limit' => $request->input('limit', 0),
+			'offset' => $request->input('offset', 0),
+			'add_header' => $include_header
+		];
+		// Get the usecase and pass in authorizer, payload and transformer
+		$this->usecase = $this->getUsecase()
+			->setPayload($payload)
```
rjmackay: It seems like these should be filters/search params as they normally are in a search usecase

src/Core/Usecase/Post/Export.php
```diff
@@ -11,69 +11,155 @@
...
-	// Usecase
 	public function interact()
 	{
-		// Fetch an empty entity...
-		$entity = $this->getEntity();
 
-		// ... verify that the entity can be searched by the current user
-		$this->verifySearchAuth($entity);
+		//FIXME inject
+		$this->data = service('factory.data');
```
rjmackay: If you're extending the search usecase you already have `this->search` which is a search data object. Does that do what you need?

src/Core/Usecase/Post/Export.php
```diff
@@ -11,69 +11,155 @@
...
+		$this->formatter->setHeading($header_row);
+		$formatter = $this->formatter;
+		/**
+		 * KeyAttributes is sent instead of the header row because it contains
+		 * the attributes with the corresponding features (type, priority) that
+		 * we need for manipulating the data
+		 */
+		$file = $formatter($posts, $keyAttributes);
+		return [
+			'results' => [
```
rjmackay: Could we make this less weird shaped? ie. `['file' => $file->file];`

src/Core/Usecase/Post/Export.php
```diff
@@ -11,69 +11,155 @@
...
+		// set the filters that should be used
+		foreach ($filters as $key => $filter) {
+			$data->$key = $filter;
+		}
+
+		return $data;
 	}
 
-	// Usecase
 	public function interact()
```
rjmackay: I think this needs to look more at what the SearchUsecase already does... it seems like a chunk of this could reuse that, or at least follow a more standard pattern.

src/App/Formatter/Post/CSV.php
rowasc: my bad. I was wrangling with tests and forgot to change this back to protected. Thanks

rowasc:
@rjmackay thanks Robbie. I some tidy up and changed things to match your feedback mostly, but I'd appreciate it if we can pair today about the usecase's interact itself to make it cleaner/more like search .

src/Core/Usecase/Post/Export.php
```diff
@@ -91,46 +91,70 @@ public function constructSearchData($job, $filters)
...
 		$this->session = service('session');
-
-		$job_id = $this->payload['job_id'];
-		$add_header = $this->payload['add_header'];
 		// Load the export job
-		$job = $this->exportJobRepository->get($job_id);
-
+		$job = $this->exportJobRepository->get($this->filters['job_id']);
 		// load the user from the job into the 'session'
 		$this->session->setUser($job->user_id);
```
rjmackay: We should probably do an Auth check based on that user, just in case?

Issue #2807, opened 2018-04-20
[Platform API] User based datasource configuration for HDX


rowasc:
## Acceptance criteria/Deliverable
- [ ] New endpoints listed in the ticket
- [ ] Postman collection for testing the endpoints 

## Expected behavior
- A user with access to the HDX feature should be able to configure an API Key in the new "Configure HDX API" page.
- This API Key will be saved and used only for their own HDX connections. 
- The API Key is not shared amongst other admins in the deployment and each HDX user needs their own key set up. 
- There is a way to get the API key in the backend to use it when requesting or creating data in the CKAN HDX API
- We need a new user data provider configuration that does not rely on the global `config` table . 

## Endpoints
GET /user/:id/settings
Returns the configuration for the user id

POST /user/:id/settings
Create a new setting for the user id

PUT /user/:id/settings/:settingId

Update user settings

### New table `user_settings`
id: PK
user_id: ref#user.id
config_key: string (initial entry options with group hdx: hxd.api_key, hdx.default_maintainer, hdx.default_source, hdx.default_org)
config_value: string (text)
updated: timestamp

## TASKS
- [x] Create new table
- [ ] Add create, update, list operations for user config

rowasc:
@caharding YUP.
cc @Shadrock I understand this is the requirement. If not: then we need a per-deployment key

Erioldoesdesign:
Can I get an update today on whether I need to look into any UI/UX design work for the configuring of the API?

So if we decide to move it to user account or webhooks or it's own settings section and if the setting up of the API requires anything (UI) that we don't have already?

This section, once in v0 will be heavily user tested to improve it for v1 but as it stands, we don't have time to test where in a deployment to place this from a user pov.

rowasc:
@Erioldoesdesign I think this was resolved with the conversations yesterday and the new screen, but lmk if that's not the case and there are more questions

rowasc:
@rjmackay @Erioldoesdesign burning question/suggestion: maybe we should never return the hdx api key .It's useless in the frontend and why would we do that* when we can just have a way to update it or create it instead ? 

*obviously this means we can't show it in the client.. but it's pretty common not to show api keys after you set them up. Maybe just the last 4 digits or something  like that ?

rjmackay:
@rjmackay I like the idea of not returning the API key once set. I don't have a strong opinion on it - it'd be a neat addition though and easy enough to implement in the API

Angamanga:
[...]

PR #2920, opened 2018-05-08
Adding User Setting Entity and associated modules

willdoran:
- Add User Setting entity

Test checklist:
- [ ] Run tests

Fixes ushahidi/platform#2807

Ping @ushahidi/platform


routes/web.php
```diff
@@ -477,6 +477,16 @@
         // Public access
         $router->get('/', 'UsersController@index');
         $router->get('/{id:[0-9]+}', 'UsersController@show');
+        $router->get('/settings', 'SettingsController@index');
+
+        // Sub-user routes
+        $router->group(['prefix' => '/{user_id:[0-9]+}'], function () use ($router) {
```
rowasc: @willdoran I think the routes need to be inside the hxl feature flag check ((Features::isEnabled('hxl')))

src/App/Repository/User/SettingRepository.php
```diff
@@ -0,0 +1,115 @@
...
+	// UserSettingRepository
+	public function existsInUserSetting($user_id)
+	{
+		return (bool) $this->selectCount(compact('user_id'));
+    }
+    
+	public function create(Entity $entity)
+	{
+		$record = $entity->asArray();
+		$record['created'] = time();
```
rowasc: @willdoran why not use a default timestamp instead?

tests/integration/users/settings.feature
```diff
@@ -0,0 +1,138 @@
...
+                "config_key":"key",
+                "config_value":"value"
+            }
+            """
+        When I request "/users/9999/settings"
+        Then the response is JSON
+        And the response has a "errors" property
+        Then the guzzle status code should be 404
+
+    Scenario: Creating a new Settings on a Stage with the wrong Form ID
```
rowasc: is this supposed to be `wrong user id` instead?

tests/integration/users/settings.feature
```diff
@@ -0,0 +1,138 @@
...
+                "priority":1
+            }
+            """
+        When I request "/users/2/Settingss"
+        Then the response is JSON
+        And the response has a "errors" property
+        Then the guzzle status code should be 400
+
+    Scenario: Updating a Settings
+        Given that I want to update a "Settings"
```
rowasc: I don't think updating a key should be possible, can you explain this one ? I'm probably just extra slow today

rowasc: ok yea extra slow it it. Ignore.

routes/web.php
willdoran: @rowasc 
I don't think so, I think this could legitimately be considered a separate feature since we're likely to use user_settings elsewhere. Would it not deserve its own flag or no?

But I can hide it for now behind that flag with a comment.

src/App/Repository/User/SettingRepository.php
willdoran: Only for consistency with other Repos

willdoran: But can change

willdoran:
@rowasc I didn't add the feature flag because I want to see if I can extract the routes easily and place them in a features route file so that we can easily see what is and isn't feature flagged. i'll try that this morning

rowasc:
@willdoran ah that's ok. LMK if you want me to review again

app/Http/Middleware/CheckFeature.php
```diff
@@ -0,0 +1,28 @@
...
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @param  mixed  ...$features
+     * @return mixed
+     */
+    public function handle($request, Closure $next, ...$features)
+    {
+        foreach ($features as $feature) {
+            if (!Features::isEnabled($feature)) {
+                abort(403, "Forbidden.");
```
rowasc: Robbie asked for a 404 when a feature is not enabled. But I'm not sure if we want to do it for everything or just the hdx feature. 

cc @rjmackay can you check this?

src/App/Validator/User/Setting/Update.php
```diff
@@ -0,0 +1,48 @@
...
+
+	public function __construct(UserRepository $user_repo)
+	{
+		$this->user_repo = $user_repo;
+	}
+
+	protected function getRules()
+	{
+		return [
+			'user_id' => [
```
rowasc: is it github or is there something wrong with the spaces/tabs here?

tests/integration/bootstrap/RestContext.php
```diff
@@ -279,9 +279,9 @@ public function iRequest($pageUrl)
 
 		// Get response object
 		$this->response = $response;
-		// $data = $this->response->getBody(true);
-		// $data = explode("\n", $data);
-		// \Log::info(print_r($data, true));
+		$data = $this->response->getBody(true);
+		$data = explode("\n", $data);
```
rowasc: I'm not sure I understand why we do this here for the print r?

app/Http/Middleware/CheckFeature.php
willdoran: Cool we can make it 404 instead with no issue, though it would be nice if this wasn't English it looks weird having a string attached to it. Surely, it can just be 404 and the browser figures it out

tests/integration/bootstrap/RestContext.php
willdoran: oh sorry that should be commented out, I forgot to revert it

bootstrap/lumen.php
```diff
@@ -104,6 +102,7 @@
     'namespace' => 'Ushahidi\App\Http\Controllers',
 ], function ($router) {
     require __DIR__.'/../routes/web.php';
+    require __DIR__.'/../routes/features.php';
```
rjmackay: Maybe we should split up routes per entity endpoints and require each one within the `web.php` file?

app/Http/Middleware/CheckFeature.php
```diff
@@ -0,0 +1,28 @@
...
+	 *
+	 * @param  \Illuminate\Http\Request  $request
+	 * @param  \Closure  $next
+	 * @param  mixed  ...$features
+	 * @return mixed
+	 */
+	public function handle($request, Closure $next, ...$features)
+	{
+		foreach ($features as $feature) {
+			if (!Features::isEnabled($feature)) {
```
rjmackay: TBH I actually preferred this being a simple if statement within routes. My thinking was that feature toggle shouldn't be complicated, and also that routes simply shouldn't exist if the feature is disabled (vs existing but having middleware). 

That said: I'm not strongly attached to my approach. If you/@rowasc prefer this way, then lets do it. As long as we pick a method and stick with it.

routes/features.php
```diff
@@ -0,0 +1,55 @@
...
+
+	// HXL
+	$router->group([
+		'prefix' => 'hxl',
+		'middleware' => ['feature:hxl']
+	], function () use ($router) {
+		$router->get('/', "HXLController@index");
+	});
+
+	// User Settings
```
rjmackay: I kind prefer keeping these within the rest of the routes, so we don't have to guess which file something is in depending on if it has a feature toggle or not. I think in general feature flagged code shouldn't move because it has a feature flag or not. (maybe this is different for permanent feature flags though? ie. thing that will never be generally available?)

I do agree that `routes/web.php` is too hard to scan though. Would it work to just split it into groups ie. routes/user, routes/form, routes/post etc? (that should happen on its on PR though)

app/Http/Controllers/API/Users/SettingsController.php
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	public function index(Request $request)
+	{
+		$params = $this->getRouteParams($request);
+		
+		$this->usecase = $this->usecaseFactory
+			->get($this->getResource(), 'search');
+		
+		if (is_array($params) && array_key_exists('user_id', $params)) {
```
rjmackay: This isn't necessary because `user_id` isn't optional right?

migrations/20180508154706_add_user_settings_table.php
```diff
@@ -0,0 +1,37 @@
...
+use Phinx\Migration\AbstractMigration;
+
+class AddUserSettingsTable extends AbstractMigration
+{
+    /**
+    * Migrate Up.
+    */
+    public function up()
+    {
+        $this->table('user_settings')
```
rjmackay: This needs a unique key constraint ie. you can't have duplicate config_key's for the same user right?

routes/features.php
```diff
@@ -0,0 +1,55 @@
...
+		$router->group([
+			'prefix' => 'settings',
+			'middleware' => ['feature:user-settings']
+		], function () use ($router) {
+			$router->get('/', 'SettingsController@index');
+			$router->get('/{id:[0-9]+}', 'SettingsController@show');
+			$router->post('/', 'SettingsController@store');
+			$router->put('/{id:[0-9]+}', 'SettingsController@update');
+			$router->delete('/{id:[0-9]+}', 'SettingsController@destroy');
+		});
```
rjmackay: I'm not sure about the structure of this. I think it matches the spec but may have issues.

1. I think ideally user settings shouldn't have IDs, rather user user_id+config_key as the primary key. The urls become `user/:user/settings/:config_key` instead.
2. I kind of think we should allow a create via PUT in this instance ie. `PUT user/1/settings/hxl_api_key` either creates or overwrites the existing API key. But doesn't really care if it existed already or not. I think that will match to how its used in the client

routes/web.php
```diff
@@ -471,13 +467,23 @@
...
     $router->group([
+        'namespace' => 'Users',
         'prefix' => 'users',
         'middleware' => ['scope:users']
     ], function () use ($router) {
         // Public access
         $router->get('/', 'UsersController@index');
         $router->get('/{id:[0-9]+}', 'UsersController@show');
 
+        // Sub-user routes
```
rjmackay: Why is this here, but repeated in features routes?

src/App/Init.php
```diff
@@ -627,6 +634,12 @@
 $di->params[Ushahidi\App\Validator\User\Reset::class] = [
 	'repo'    => $di->lazyGet('repository.user')
 ];
+$di->params[Ushahidi\App\Validator\User\Setting\Update::class] = [
+	'user_repo'    => $di->lazyGet('repository.user')
+];
+$di->params[Ushahidi\App\Validator\Contact\Update::class] = [
```
rjmackay: Why did this change in this PR?

src/App/Repository/Config/features.php
```diff
@@ -85,4 +85,9 @@
 	'hxl' => [
 		'enabled' => false,
 	],
+
+	// Enable or disable User Settings feature
+	'user-settings' => [
+		'enabled' => true,
```
rjmackay: This should be disabled by default. We need to add a way to override per environment but for now we should always be keeping things disabled unless we want them on in production.

src/App/Repository/User/SettingRepository.php
```diff
@@ -0,0 +1,115 @@
...
+	{
+		$query = $this->search_query;
+
+		if ($search->user_id) {
+			$query->where('user_id', '=', $search->user_id);
+		}
+	}
+
+	// UserSettingRepository
+	public function updateCollection(array $entities)
```
rjmackay: Is this used anywhere yet? I really like this idea ie. that you can overwrite all user settings at once, since that seems likely how it'll get used in the UI anyway

app/Http/Controllers/API/Users/SettingsController.php
```diff
@@ -0,0 +1,40 @@
...
+			$this->usecase
+				->setFilters($request->query() + [
```
rjmackay: I wonder if we could rewrote this slightly and avoid treating user_id as a filter... its really an identifier ie. part of the primary key, not an optional filter

src/App/Repository/User/SettingRepository.php
```diff
@@ -0,0 +1,115 @@
...
+	// UserSettingRepository
+	public function getByUser($user_id)
+	{
+		$query = $this->selectQuery(compact($user_id));
+		$results = $query->execute($this->db);
+
+		return $this->getCollection($results->as_array());
+	}
+
+	public function deleteAllForForm($user_id)
```
rjmackay: deleteAllForUser?

src/App/Validator/User/Setting/Update.php
```diff
@@ -0,0 +1,48 @@
...
+				['digit'],
+				[[$this->user_repo, 'exists'], [':value']],
+			],
+			'config_key' => [
```
rjmackay: Add a list of allowed keys? or maybe we don't even care

src/Core/Entity/UserSetting.php
```diff
@@ -0,0 +1,43 @@
...
+    protected $config_key;
+    protected $config_value;
+    protected $created;
+    protected $updated;
+
+	// DataTransformer
+	protected function getDefinition()
+	{
+		return [
+			'id'       => 'int',
```
rjmackay: I think ideally these don't have an ID, just a combined config_key+user_id as primary key

src/Core/Entity/UserSettingRepository.php
```diff
@@ -0,0 +1,27 @@
...
+ * @copyright  2018 Ushahidi
+ * @license    https://www.gnu.org/licenses/agpl-3.0.html GNU Affero General Public License Version 3 (AGPL3)
+ */
+
+namespace Ushahidi\Core\Entity;
+
+use Ushahidi\Core\Entity\Repository\EntityGet;
+use Ushahidi\Core\Entity\Repository\EntityExists;
+
+interface UserSettingRepository extends
```
rjmackay: I haven't done this everywhere yet but this should probably extend Create and Update Repository interfaces I think

src/Core/Tool/Authorizer/UserSettingAuthorizer.php
```diff
@@ -0,0 +1,82 @@
...
+		// These checks are run within the user context.
+		$user = $this->getUser();
+
+		// Only logged in users have access if the deployment is private
+		if (!$this->canAccessDeployment($user)) {
+			return false;
+		}
+
+		// Role with the Manage Users permission can manage all users
+		if ($this->acl->hasPermission($user, Permission::MANAGE_USERS)) {
```
rjmackay: I think maybe this shouldn't be true. ie. a user should be the only person who can edit their own settings? Admins don't need to be able to edit a users settings

bootstrap/lumen.php
willdoran: Yeah, splitting per endpoint would be ideal. We'll need a ticket for it,

rowasc: Agree on splitting per endpoints, that'd be nice. We could keep the CheckFeature and separate everything per endpoint , instead of having the routes/feature.php

rowasc: but another option is that everything has their own file , unless feature flagged, I guess? I kind of like the idea of being able to check which flags are still being used easily.

app/Http/Controllers/API/Users/SettingsController.php
willdoran: It was at one point but I think I removed that route. I'll check.

src/App/Init.php
willdoran: It was changed because of fiddling with white space in editors.

src/App/Repository/User/SettingRepository.php
willdoran: No it's only a placeholder so far.

src/App/Validator/User/Setting/Update.php
willdoran: Right now we don't have a list but I'll add it as a comment.

src/Core/Entity/UserSettingRepository.php
willdoran: where are those?

src/Core/Tool/Authorizer/UserSettingAuthorizer.php
willdoran: Not sure about that. Ideally, you should be able to see all user settings but I'm not sure really what benefit it is. I'll remove it.

routes/web.php
willdoran: I checked I don't see it repeated but I was wondering why is it outside auth?

routes/web.php
rjmackay: Everything being added for HXL (or anything else) should get a feature flag. That way we can eventually deploy it to prod but disabled till its 100% working. I would argue for just using the HXL feature flag - on the assumption that this feature flag is only going to be used to disable the feature during develop and QA, not a long lived feature flag that will be tied to users plans.

src/Core/Tool/Authorizer/UserSettingAuthorizer.php
```diff
@@ -0,0 +1,77 @@
...
+		// These checks are run within the user context.
+		$user = $this->getUser();
+
+		// Only logged in users have access if the deployment is private
+		if (!$this->canAccessDeployment($user)) {
+			return false;
+		}
+
+		// Admin user should be able to do anything - short of deleting self
+		if ($this->isUserAdmin($user)) {
```
rjmackay: Sorry I wasn't clear before. I don't think this should be here either.

src/Core/Entity/UserSettingRepository.php
rjmackay: `Ushahidi\Core\Usecase` I think

routes/web.php
rjmackay: Users are available to anonymous users to provide the users name for post detail - but we filter out email and other sensitive info.

migrations/20180508154706_add_user_settings_table.php
```diff
@@ -0,0 +1,37 @@
...
+            ->addColumn('config_key', 'string', [
```
rjmackay: Can you add an index since we'll be searching by config key?

willdoran: Yep I have it's just not committed just yet.

willdoran:
> I'm going to add groups as well.
What we're to do is basically have a bag of properties that belongs to a user, the likely operations we'll have are. Get me all settings by user or get me all settings related to a feature. I also don't really like config, I agree that the serialised json is awful and it solves a problem we don't really have which is trying to avoid millions of rows of settings data.

My major concern is that the keys are not going to be unique across different systems. We can make them unique ourselves on purpose but it would be much nicer to have the settings match the expected structure of either internal or external usecases. I'd rather be able to do 

`$config = User.getSettingFor('twitter');`

Than have to do 

`
$user_settings = User.getSettingFor('twitter');
$config = [
   'api_key' => $user_settings['twitter_api_key'],
....
   'some_setting_99 => $user_setting['twitter_some_setting_99']
`

The stupider the API can be about the data model the better. It doesn't really matter at this stage so I'll drop the group but it'll be a very annoying change to make later. Though I guess it could be looked at under data model. The important thing to recognise is that these aren't our settings they are at the moment 3rd party settings so we don't control the naming scheme.

src/Core/Entity/UserSettingRepository.php
willdoran: I think we can do that in a separate ticket, can you add one for it and assign it to @jrtricafort so he can track it into a later cycle?

src/App/Init.php
willdoran: Our white space is wrong everywhere.

PR #2922, opened 2018-05-09
fix: webhooks $request was missing in the controller

rowasc:
- Fix webhooks controller - was missing $request

Test checklist:
- [x] bin/behat

Ping @ushahidi/platform


rowasc:
@rjmackay I found this missing while doing another task

Issue #2881, opened 2018-05-01
[ Platform API ] Setup CKAN API Client for laravel


rowasc:
### TODO 
- [ ] Setup a CKAN client in Lumen to connect to the HDX CKAN API 

###  Acceptance criteria: 
- [ ] I can call a package from the HDX usecases and get/send data to ckan api
- [ ] I can configure the CKAN API url to be test-humdata.org or humdata.org depending on .env vars
- [ ] I can return a friendly error message to the client if CKAN is down

## Implementation notes
I found a few packages that are generic PHP packages. I also found this one  https://github.com/Germanaz0/laravel-ckan-api for laravel that looks nice enough to use and extend as needed. We could fork and add some tests to it instead of writing our own client. 
Good things about it:
- ready for laravel so probably won't have issues to issue in lumen
- easy to use
- has nice examples in docs file
- is being updated from time to time 
- uses guzzle
- Small and most of its code looks testable

Bad
- no tests provided :(

rowasc:
@rjmackay is there a CKAN client you prefer? I know you looked at some but I can't remember if you liked any of them . 
I was going to just make a quick wrapper myself but this laravel one seems ok as a starting point that we could extend as needed / contribute back to

rjmackay:
No. Basically pick something that seems stable and recently maintained

> this laravel one 

link?

rowasc:
https://github.com/Germanaz0/laravel-ckan-api

rowasc:
@kinstelli  @willdoran Can you review and provide any input you have?

willdoran:
@rowasc Approve

PR #2924, opened 2018-05-09
Adding ckan client api,

willdoran:
- Adding ckan laravel client
- The client is not lumen compatible so I had to add a work around to deal with the use of the config_path() function by autoloading a helper file.

Test checklist:
- [ ]

Fixes ushahidi/platform#2881

Ping @ushahidi/platform


willdoran:
@rjmackay @rowasc Do you have a better idea for working around the use of config_path(), we could either submit a PR to make their library lumen compatible or we could fork it or we could go with this, this is likely to crop up again though and this might be the best work around though we'd have to track which libs had the problem.

config/ckan_api.php
```diff
@@ -0,0 +1,2 @@
+CKAN_API_URL=""
+CKAN_API_KEY=""
```
rowasc: this needs to come from the database not from the .ENV since it's set up by each user

composer.json
```diff
@@ -47,7 +47,8 @@
         "nexmo/client": "^1.1",
         "sentry/sentry-laravel": "^0.8.0",
         "laravel/passport": "^4.0",
-        "giggsey/libphonenumber-for-php": "^8.9"
+        "giggsey/libphonenumber-for-php": "^8.9",
```
rowasc: why do we need this again?

willdoran: no idea, I can remove it.

config/ckan_api.php
willdoran: That is going to be more complex then because I don't think this client supports that but I'll look.

rowasc: cool. For context: this key is the one the user provides and you save in the user settings hdx.api_key config value

willdoran:
It's possible this client isn't going to work because it instantiates the API details(url, key) only from config :(

willdoran:
@rowasc I think this is correct, could you take a look and double check this does what you want. I included some example code in the HXL controller I can remove it before merging but it's just so you can see if it will do all you need. 

It's relatively simple, we import the client, pass it a guzzle client and then use it directly. So long as the config is formatted as in the example it should work as expected.

It will also return sensible errors, though those come directly from HDX we just pass them through.

rowasc:
@willdoran I think this is OK (after commenting out the create call) . We will need some validation on our side because the CKAN API implementation in HDX fails at validating some things and just results in 500 errors when it should say 422:something , but this is what we need to start coding the integration so I think it's ok.

PR #2930, opened 2018-05-11
refactor: Split routes into a file for each top level endpoint

rjmackay:
- refactor: Split routes into a file for each top level endpoint

Test checklist:
- [ ] composer test

Ping @ushahidi/platform


routes/collections.php
```diff
@@ -0,0 +1,31 @@
...
+    $router->get('/', 'CollectionsController@index');
+    $router->get('/{id}', 'CollectionsController@show');
+    $router->group(['prefix' => '/{set_id:[0-9]+}/posts'], function () use ($router) {
+        $router->get('/', 'PostsController@index');
+        $router->get('/{id}', 'PostsController@show');
+    });
+
+    // Restricted access
+    $router->group([
+        'middleware' => ['auth:api', 'scope:collections,sets']
```
willdoran: is the repetition of scope needed here?

routes/config.php
```diff
@@ -0,0 +1,20 @@
...
+    'middleware' => ['scope:config']
+], function () use ($router) {
+    // Public access
+    $router->get('/', ['uses' => 'ConfigController@index']);
+    // @todo stop using this in client, and remove?
+    $router->options('/', ['uses' => 'ConfigController@indexOptions']);
+    $router->get('/{id}', ['uses' => 'ConfigController@show']);
+
+    // Restricted access
+    $router->group(['middleware' => ['auth:api', 'scope:config']], function () use ($router) {
```
willdoran: Not sure scope is needed here since the outer route has it, I thought they inherrited.

routes/exports.php
```diff
@@ -0,0 +1,32 @@
...
+$router->group([
+    'namespace' => 'Exports',
+    'middleware' => ['auth:api', 'scope:posts'],
+    'prefix' => '/exports/jobs'
+], function () use ($router) {
+    $router->get('/', 'JobsController@index');
+    $router->post('/', 'JobsController@store');
+    $router->get('/{id}', 'JobsController@show');
+    $router->put('/{id}', 'JobsController@update');
+    $router->delete('/{id}', 'JobsController@destroy');
```
willdoran: It's not clear why {id} is specified sometimes and at other times a regex to match a numeric, is there a difference in the id being passed?

routes/posts.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
+// Posts
+$router->group([
+    'namespace' => 'Posts',
```
willdoran: Why is the use of namespace variable?

rjmackay:
> create a set of base inheritable routes and only provide those where there is a modification

We could probably loop over an array for many routes, but I preferred explicitly defining each route. It avoids accidentally defining broken/unused routes i.e. in Kohana we automatically got routes for POST/PUT/DELETE even on controllers where we only wanted GET, and then had the action map in controllers checking if they were allowed. This way: if we don't want those, we just don't define them.

In the spirit of reducing boilerplate though, maybe we could define a `resource` helper similar to laravel? In laravel you can do
`Route::resource('posts', 'PostsController');` or more complex versions like 
```
Route::resource('photos', 'PhotoController')->only([
    'index', 'show'
]);

Route::resource('photos', 'PhotoController')->except([
    'create', 'store', 'update', 'destroy'
]);
```

Lumen doesn't have that but its probably just a few lines to implement it.

routes/web.php
```diff
@@ -11,6 +11,45 @@
 |
 */
 
+// @todo move to helper
+if (!function_exists('resource')) {
+    function resource($router, $uri, $controller, array $options = [])
```
rjmackay: @willdoran this is modelled off laravels `Route::resource()` function. It needs to move elsewhere but I wanted to validate the idea first

routes/tags.php
```diff
@@ -6,15 +6,13 @@
     'middleware' => ['scope:tags']
 ], function () use ($router) {
     // Public access
-    $router->get('/', 'TagsController@index');
-    $router->get('/{id}', 'TagsController@show');
+    resource($router, '/', 'TagsController', [
```
rjmackay: This pattern happens in several places - there should clearly be either an `public-read` option or second helper for this.

routes/users.php
```diff
@@ -2,21 +2,21 @@
...
     ], function () use ($router) {
-        $router->post('/', 'UsersController@store');
-        $router->put('/{id:[0-9]+}', 'UsersController@update');
-        $router->delete('/{id:[0-9]+}', 'UsersController@destroy');
-        $router->get('/me', 'UsersController@showMe');
-        $router->put('/me', 'UsersController@updateMe');
+        resource($router, 'users', 'UsersController', [
+            'only' => ['store', 'update', 'destroy']
+        ]);
+        $router->get('/users/me', 'UsersController@showMe');
```
rjmackay: There are a lot of minor exceptions like these. I do wonder how useful the helper is if I'm constantly adding extra routes.

routes/notifications.php
```diff
@@ -1,13 +1,6 @@
 <?php
 
 // Notifications
-$router->group([
+resource($router, 'notifications', 'NotificationsController', [
     'middleware' => ['auth:api', 'scope:notifications'],
```
rjmackay: We could almost include scope middleware automatically too, with an optional override. Same with auth.
It depends how much we want to match the original laravel API

rjmackay:
> We might be better looking at something like Route::resources() to handle implicit routes and then only override non-standard ones.

I totally only saw this comment after implementing. Sorry if it seemed like I was claiming your idea like my own *shuffles out awkwardly*

app/helpers.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+if (!function_exists('resource')) {
```
rjmackay: Moved the helper here.

FYI This pattern of having a `helpers.php` file is used in laravel and lumen themselves. I've used it in platform-cloud-interface when porting a helper from laravel into lumen, I suspect we'll want to do that again in platform

rjmackay:
@willdoran I've merged from develop and moved `resource()` to a permanent home.
One final Q: Do you think it's still worth keeping all the routes in their own file now they're using the resource() helper and are much shorter?

willdoran:
@rjmackay I'm not positive, on the one hand some like forms and posts have a bit more length and I think would be justified as their own files. But I don't know if contacts and dataproviders need their own files. 

It'd be nice to have a single consistent pattern. I think forms and posts deserve their own files and on the basis of consistency I'd like all other routes to have their own files as well rather than there being two rules one for short routes and one for long.

But it's not a deeply held belief so I'm ok if you want to recombine them.

PR #2948, opened 2018-05-15
Feature/2936 hdx metadata table

rowasc:
This pull request makes the following changes:
- Implements #2936 

Test checklist:
- [ ] run ./bin/behat

- [ ] I certify that I ran my checklist

Fixes ushahidi/platform# .

Ping @ushahidi/platform

Issue #2937, opened 2018-05-12
[Platform API] HXL Meta Data Endpoint


willdoran:
### Overview
The HXL Meta Data entity will should be createable via the API for completeness

### Requirements
1. Validation
  - private: 
     - boolean
     - required
  - dataset_title: 
     - text
     - required
     - length: varchar 255
  - organisation: 
     - text
     - required
     - length: varchar 255
  - source: 
     - text
     - required
     - length: varchar 255
  - maintainer: 
     - text
     - required
     - length: varchar 255
  - license_id: 
     - license id must exist
     - required
  - export_job_id: 
     - export_job id must exist
     - required
  - user_id: 
     - user id must exist
     - required
2. Authorisation
  - Only owners are allowed to change
  - All users can search 
3. Formatting 
  - Uses standard entity formater
4. Usecase
  - Create
5. Verbs
 - POST < store>

rowasc:
@willdoran is there a reason we allow a text_medium for title? I think it might be too much.  How about a varchar with maybe 256? This is the title that users in HDX will see when they look for datasets so I think it's reasonable to keep it short .


In addition to this, we might want to save organization that the dataset belongs to, as well as a resource id that can be associated to the dataset.

willdoran:
@rowasc Updated spec

kinstelli:
@willdoran  Is there other info besides organization that we potentially need in this endpoint? 
Does this store to the HXL specific table?
Also, is this just a POST endpoint?

rowasc:
@willdoran hey sorry I still don't see `organization` there , but maybe it was missed in the comment. Can you confirm if we will have it or if you think we dont need to / why?

rowasc:
ah
>In that table there would be hdx_organistion:value and the other reusable settings

So this is technically correct, we do save defaults there for when the user does a new export, but that's just a default, not the value that matches the dataset itself. To make this clear: organization is a value that can change for each user. For instance, I'm part of Ushahidi and Comrades in hdx.

willdoran:
ah!!! I didn't realise, fixed can you recheck

rowasc:
APPROVED

kinstelli:
Can we include the verbs of this endpoint in the spec? :)

rowasc:
I'd support that :D @willdoran  @kinstelli

rowasc:
APPROVED

kinstelli:
Okay, POST is great. APPROVED post-hoc.

[...]

PR #2949, opened 2018-05-15
Feature/2937 hdx metadata endpoint

rowasc:
- Implements spec from #2937 
- Depends on PR https://github.com/ushahidi/platform/pull/2948
Test checklist:
- [ ] ./bin/behat

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -249,6 +249,7 @@ $router->group([
             // Restricted access
             $router->get('/licenses', 'HXLLicensesController@index');
             $router->get('/tags', 'HXLTagsController@index');
+            $router->post('/metadata', 'HXLMetadataController@store');
         });
     }
 
--- a/routes/web.php
+++ b/routes/web.php
@@ -240,19 +240,6 @@ $router->group([
         });
     });
 
-    if (Features::isEnabled('hxl')) {
-        $router->group([
-            'namespace' => 'HXL',
-            'prefix' => 'hxl',
-            'middleware' => ['auth:api'] //TODO which scopes do we need?
-        ], function () use ($router) {
-            // Restricted access
-            $router->get('/licenses', 'HXLLicensesController@index');
-            $router->get('/tags', 'HXLTagsController@index');
-            $router->post('/metadata', 'HXLMetadataController@store');
-        });
-    }
-
     // Layers
     $router->group([
         'prefix' => 'layers',
@@ -480,6 +467,7 @@ $router->group([
 
     // Users
     $router->group([
+        'namespace' => 'Users',
         'prefix' => 'users',
         'middleware' => ['scope:users']
     ], function () use ($router) {
@@ -496,6 +484,21 @@ $router->group([
             $router->delete('/{id:[0-9]+}', 'UsersController@destroy');
             $router->get('/me', 'UsersController@showMe');
             $router->put('/me', 'UsersController@updateMe');
+            
+            // Sub-user routes
+            $router->group(['prefix' => '/{user_id:[0-9]+}'], function () use ($router) {
+                // Settings
+                $router->group([
+                    'prefix' => 'settings',
+                    'middleware' => ['feature:user-settings']
+                ], function () use ($router) {
+                    $router->get('/', 'SettingsController@index');
+                    $router->post('/', 'SettingsController@store');
+                    $router->get('/{id}', 'SettingsController@show');
+                    $router->put('/{id}', 'SettingsController@update');
+                    $router->delete('/{id}', 'SettingsController@destroy');
+                });
+            });
         });
     });
 
@@ -511,6 +514,18 @@ $router->group([
         $router->delete('/{id:[0-9]+}', 'WebhooksController@destroy');
 
         $router->put('/posts', 'WebhookPostsController@update');
+    });
+
+    // HXL
+    $router->group([
+        'prefix' => 'hxl',
+        'middleware' => ['auth:api', 'feature:hxl'],
+        'namespace' => 'HXL'
+    ], function () use ($router) {
+        $router->get('/', "HXLController@index");
+        $router->get('/licenses', 'HXLLicensesController@index');
+        $router->get('/tags', 'HXLTagsController@index');
+        $router->post('/metadata', 'HXLMetadataController@store');
     });
 });
 
--- a/src/App/Formatter/HXL/HXLMetadata.php
+++ b/src/App/Formatter/HXL/HXLMetadata.php
@@ -1,0 +1,20 @@
+<?php
+
+/**
+ * Ushahidi API Formatter for HXL Metadata
+ *
+ * @author     Ushahidi Team <team@ushahidi.com>
+ * @package    Ushahidi\Application
+ * @copyright  2014 Ushahidi
+ * @license    https://www.gnu.org/licenses/agpl-3.0.html GNU Affero General Public License Version 3 (AGPL3)
+ */
+
+namespace Ushahidi\App\Formatter\HXL;
+
+use Ushahidi\App\Formatter\API;
+use Ushahidi\Core\Traits\FormatterAuthorizerMetadata;
+
+class HXLMetadata extends API
+{
+    use FormatterAuthorizerMetadata;
+}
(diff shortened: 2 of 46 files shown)


src/App/Validator/HXL/Metadata/Create.php
```diff
@@ -0,0 +1,91 @@
...
+        $this->license_repo = $license_repo;
+        $this->export_job_repo = $export_job_repo;
+    }
+    /**
+     * @return array|\Ushahidi\Core\Tool\ArrayValidation
+     */
+    protected function getRules()
+    {
+        return array_merge([
+            'private' => [
```
rowasc: @rjmackay  @willdoran code climate complains because of this function being 29 lines. Maybe I could have a function return the rules that are the same as the ones for organisation and dataset_title, but I'm not sure it helps with readability. Tagging you to get your opinion on this

src/App/Validator/HXL/Metadata/Create.php
```diff
@@ -0,0 +1,91 @@
...
+                ['max_length', [':value', 255]],
+                ['regex', [':value', Validator::REGEX_STANDARD_TEXT]], // alpha, number, punctuation, space
+            ],
+            'maintainer' => [
+                ['not_empty'],
+                ['min_length', [':value', 2]],
+                ['max_length', [':value', 255]],
+                ['regex', [':value', Validator::REGEX_STANDARD_TEXT]], // alpha, number, punctuation, space
+            ]
+        ], $this->getForeignKeyRules());
```
rowasc: I separated the rules for the keys to help code climate calm down. Then realized I'm still over 25 lines

src/App/Validator/HXL/Metadata/Create.php
rjmackay: I think maybe code climate is just over enthusiastic. 25 lines is really short, and theres barely any real logic in this file, so I don't think it bears splitting it.

src/App/Validator/HXL/Metadata/Create.php
rjmackay: I'm not sure it makes it any more clear either... I don't think you need to take the code climate stuff strictly right now, still just experimenting with it.
Maybe we should just relax that rule a bit?

src/App/Validator/HXL/Metadata/Create.php
rowasc: ah ok, makes sense.

rowasc:
@willdoran @rjmackay travis is timing out in behat, do you know what could be happening there?

src/App/Validator/HXL/Metadata/Create.php
rowasc: uh. Yea me either. 
I'll change it back today

rowasc:
@rjmackay can you review?

src/Core/Tool/Authorizer/HXLMetadataAuthorizer.php
```diff
@@ -0,0 +1,55 @@
...
+    // if roles are available for this deployment.
+    use AclTrait;
+
+    use OwnerAccess;
+
+
+    /* Authorizer */
+    public function isAllowed(Entity $entity, $privilege)
+    {
+        if (parent::isAllowed($entity, $privilege)) {
```
rjmackay: It'd be cleaner to express this as `parent::isAllowed() && $this->isUserOwner()`.

However if we're going to use the HXL Authorizer everywhere like this, it might be better as a DataImportExport trait that provides `$this->canImportExportData()`.

The original ideal of authorizers was that you could read them and understand the intent. So a parent::isAllowed() check doesn't provide much info about intent here.

rjmackay: I'll also caveat this with: if you'd rather just get on with it and keep coding, this isn't a blocking/must fix request.

routes/web.php
```diff
@@ -525,6 +525,7 @@
         $router->get('/', "HXLController@index");
         $router->get('/licenses', 'HXLLicensesController@index');
         $router->get('/tags', 'HXLTagsController@index');
+        $router->post('/metadata', 'HXLMetadataController@store');
```
rjmackay: I don't quite understand why this is write only? Do we never expect to read metadata again?

src/App/Formatter/HXL/HXLMetadata.php
```diff
@@ -0,0 +1,20 @@
...
+ * @copyright  2014 Ushahidi
+ * @license    https://www.gnu.org/licenses/agpl-3.0.html GNU Affero General Public License Version 3 (AGPL3)
+ */
+
+namespace Ushahidi\App\Formatter\HXL;
+
+use Ushahidi\App\Formatter\API;
+use Ushahidi\Core\Traits\FormatterAuthorizerMetadata;
+
+class HXLMetadata extends API
```
rjmackay: I've got to figure out how to remove the need for these classes, since they're all the same!

src/App/Init.php
```diff
@@ -768,30 +768,63 @@
...
 $di->params['Ushahidi\Factory\AuthorizerFactory']['map']['hxl'] =
     $di->lazyGet('authorizer.hxl');
+
+// hxl meta_data
+$di->set('repository.hxl_meta_data', $di->lazyNew(Ushahidi\App\Repository\HXL\HXLMetadataRepository::class));
+$di->set('formatter.entity.hxl_meta_data', $di->lazyNew(Ushahidi\App\Formatter\HXL\HXLMetadata::class));
+$di->set('authorizer.hxl.meta_data', $di->lazyNew('Ushahidi\Core\Tool\Authorizer\HXLMetadataAuthorizer'));
+
+$di->setter[Ushahidi\App\Formatter\HXL\HXLMetadata::class]['setAuth'] = $di->lazyGet("authorizer.hxl.meta_data");
+$di->setter[Ushahidi\App\Formatter\HXL\HXLMetadata::class]['setAuth'] =
```
rjmackay: Repeated line here?

tests/integration/hxl/metadata.feature
```diff
@@ -0,0 +1,88 @@
...
+          """
+          {
+              "license_id": 1,
+              "organisation": "ushahidi",
+              "dataset_title": "ushahidi-dataset",
+              "source": "other",
+              "maintainer": "ushahidi-maintainer",
+              "private": true,
+              "export_job_id":1,
+              "user_id":2
```
rjmackay: Do we actually need to pass `user_id`? Is it always the current user? If so, we should be pulling it from user context.

src/App/Repository/HXL/HXLMetadataRepository.php
```diff
@@ -0,0 +1,53 @@
...
+class HXLMetadataRepository extends OhanzeeRepository implements
+    HXLMetadataRepositoryContract
+{
+    // OhanzeeRepository
+    protected function getTable()
+    {
+        return 'hxl_meta_data';
+    }
+
+    public function getSearchFields()
```
rjmackay: Minor ... but if this is write only, then we don't need the search methods, and possibly should be extending the SearchRepository interface in the contract

tests/integration/hxl/metadata.feature
rjmackay: For reference you can just do this to get it from the session instead: https://github.com/ushahidi/platform/blob/develop/src/Core/Usecase/Contact/CreateContact.php#L19

tests/integration/hxl/metadata.feature
```diff
@@ -0,0 +1,88 @@
+@hxl
```
rjmackay: This should probably have `@hxlEnabled` to ensure the feature is always enabled.

routes/web.php
rowasc: (answered in slack but adding here to not forget)
Right now, we don't expect to read metadata from an endpoint, so this was spec'd as write only.

tests/integration/hxl/metadata.feature
rowasc: eh , I guess it should always be the current user yea. Changing

src/App/Repository/HXL/HXLMetadataRepository.php
rowasc: The endpoint is write only. Which doesn't mean we don't search data in that reopo. We use the metadata for sending to hdx and search for dataset_title  to verify if we need to update or create a dataset.

tests/integration/hxl/metadata.feature
rowasc: done

src/App/Init.php
rowasc: yep thanks

PR #2950, opened 2018-05-15
wip: Ruleset changes to check doc comments, throw tags, disallow short array 

rowasc:
- Sets ruleset for PHP code 

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


src/ruleset.xml
```diff
@@ -4,7 +4,19 @@
 
     <rule ref="PSR2" />
 
+    <!-- Checks that the opening PHP tag is the first content in a file -->
```
rowasc: @willdoran this adds some of what we discussed today for the new code review guidelines. We shouldn't merge until there's agreement on the rules, just sending this PR so you have it as reference

src/ruleset.xml
```diff
@@ -4,7 +4,19 @@
...
     <rule ref="Generic.PHP.CharacterBeforePHPOpeningTag" />
+    <!-- Checks format of doc blocks -->
+    <rule ref="Generic.Commenting.DocComment" />
+    <!-- forbids TODO commemts -->
+    <rule ref="Generic.Commenting.Todo"/>
```
rjmackay: I'm torn on this one. Todo's can be helpful future pointers... but we don't want to accidentally leave them in either

rjmackay:
👍 I'd definitely like consistent comment requirements.

src/ruleset.xml
rowasc: yeah. So Will had the idea that TODOs should just become an issue and not live in the code, I think that'd be ideal, since we don't usually go back to cleanup TODOs otherwise. 

Maybe a decent approach would be that instead of a TODO we have an issue link with some other flag like "ISSUE" that we can check for format with a custom linter (there might be something that does this already) ? another idea is to forbid TODOs but only if they are not followed by an issue link. Not sure. Throwing ideas around :)

rowasc:

PR #2954, opened 2018-05-16
[wip-self-review] Feature/2521 csv hxl row

rowasc:
-

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


rowasc:
@willdoran can you take a look at this when you have a chance? There's a TODO in there but that's just for when we actually have/need multicolumn fields in regular CSVs*, not something that we could do in this epic 
I think I should create a ticket but I'm not sure how to phrase it since right now it isn't needed (maybe a  "CSV multicolumn field support" ticket? )

PR #2955, opened 2018-05-16
WIP: adding initial API calls to HDX

kinstelli:
- adds some HDX API calls using CKAN API wrapper library
- adds some tests

### DIFF
--- a/src/App/ExternalServices/HDXInterface.php
+++ b/src/App/ExternalServices/HDXInterface.php
@@ -1,0 +1,141 @@
+<?php
+
+namespace Ushahidi\App\ExternalServices;
+
+/**
+ * HDX Interface -
+ *
+ * @author    Ushahidi Team <team@ushahidi.com>
+ * @package   ExternalServices\HDX
+ * @copyright 2018 Ushahidi
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU General Public License Version 3 (GPLv3)
+ */
+
+use Ushahidi\Core\Usecase\HXL\SendHXLUsecase;
+use Germanazo\CkanApi\CkanApiClient;
+use GuzzleHttp\Client;
+use Log;
+
+class HDXInterface
+{
+    protected $ckanClient;
+    protected $userAPIKey;
+    protected $ckanURL;
+
+    public function __construct($url, $key)
+    {
+         $this->ckanURL = $url;
+         $this->userAPIKey = $key;
+    }
+
+    public function setServer($url)
+    {
+        $this->ckanURL = $url;
+    }
+
+    public function setUserAPIKey($key)
+    {
+        $this->userAPIKey = $key;
+    }
+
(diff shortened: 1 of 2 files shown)


rowasc:
@kinstelli added HDX_URL in .env.default since that's a global value. 
Injected usecase/repos where needed. Now working on the $handler setup

.env.example
```diff
@@ -16,3 +16,4 @@ QUEUE_DRIVER=sync
 
 REDIS_HOST=127.0.0.1
 REDIS_PORT=6379
+HDX_URL=https://test-data.humdata.org/api/action
```
rowasc: @kinstelli I think it should just be https://test-data.humdata.org but I'm not sure.

src/Core/Usecase/HXL/SendHXLUsecase.php
```diff
@@ -0,0 +1,129 @@
...
+    {
+        $this->jobID = $jobID;
+    }
+
+    public function interact()
+    {
+        // get job by job_id
+        $job = $this->exportJobRepository->get($this->getIdentifier('job_id'));
+        // get user settings by user id
+        $user_settings = $this->userSettingRepository->getByUser($job->user_id);
```
rowasc: FIXME: need to call getConfigKeyByUser instead (I'll add it in 30)

src/Core/Usecase/HXL/SendHXLUsecase.php
```diff
@@ -0,0 +1,129 @@
...
+        } else {
+            $job->setState(['status' => 'SUCCESS']);
+        }
+        $this->exportJobRepository->update($job);
+        return $job;
+    }
+
+    private function setHDXInterface($user_settings)
+    {
+        $this->hdxInterface = new HDXInterface(
```
rowasc: FIXME should be using $user_settings->config_value since it receives the config key with api_key in it

adding this in 30

src/App/ExternalServices/HDXInterface.php
```diff
@@ -0,0 +1,159 @@
...
+            // @TODO: be graceful here
+            $createResult = ['error' => 'Unable to create dataset on HDX server.'];
+            Log::error('Unable to create dataset on HDX server: '.print_r($e, true));
+        }
+        return $createResult;
+    }
+
+    public function getAllOrganizationsForUser()
+    {
+        $apiClient = $this->getApiClient();
```
rowasc: FIXING this now to make the get_organizations_for_user request which is the one hdx has available to get a specific list of organisations for this user api key

.env.example
kinstelli: Not sure either. Maybe production is okay, since users still need a key.

src/App/ExternalServices/HDXInterface.php
kinstelli: @rowasc - oops, sorry if that wasn't the right API call. Let me know if you want me to grab this and/or update the tests.

PR #2956, opened 2018-05-16
CLI Export Send to Export Broker

willdoran:
Adding short term cli that will send jobs to the export broker until it is fully integrated into core

- Adds Export Send CLI

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/src/Console/Command/SendExport.php
+++ b/src/Console/Command/SendExport.php
@@ -12,10 +12,9 @@
 namespace Ushahidi\Console\Command;
 
 use Illuminate\Console\Command;
-
 use Ushahidi\Core\Entity\ExportJobRepository;
 
-class Webhook extends Command
+class SendExport extends Command
 {
     private $db;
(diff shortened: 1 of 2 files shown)


src/Console/Command/SendExport.php
```diff
@@ -0,0 +1,93 @@
...
+ * @license    https://www.gnu.org/licenses/agpl-3.0.html GNU Affero General Public License Version 3 (AGPL3)
+ */
+
+namespace Ushahidi\Console\Command;
+
+use Illuminate\Console\Command;
+
+use Ushahidi\Core\Entity\ExportJobRepository;
+
+class Webhook extends Command
```
kinstelli: Should this be called something more like SendExport?

willdoran: Yep

PR #2963, opened 2018-05-17
fix: metadata should be created once per job

rowasc:
- adds check for metadata to not allow more than 1 metadata object per job

Test checklist:
- [ ] Create a metadata object for a job
- [ ] Try to create a second metadata object for the same job
- [ ] You should get an error  "Cannot create a metadata entity for a job that already has one"
- [x]  I certify that I ran my checklist

Fixes ushahidi/platform# .

Ping @ushahidi/platform


rowasc:
@willdoran @kinstelli this was not part of the original spec but realized it could happen today while checking in with Anna. It would be problematic for the client if we allow it without a good error response. LMK if you think this should/shouldn't land.

PR #2967, opened 2018-05-17
fix: Fix webhooks/posts route and controller

rjmackay:
- Update webhooks/posts route to use signature middleware
- Update webhooks/posts route to include post id
- Update WebhookPostsController to fix class name and method name
- Update WebhookPostsController to extend PostsController so it includes
  correct post type identifier
- Update WebhookUpdatePost Usecase to skip auth check for reading post
  since we have no user session and it will always fail
- Add tests for webhooks/posts

Test checklist:
- [x] bin/phpunit tests/unit/API/WebhooksPostsUpdateAPI.php 

Ping @ushahidi/platform

### DIFF
--- a/src/Core/Usecase/Post/WebhookUpdatePost.php
+++ b/src/Core/Usecase/Post/WebhookUpdatePost.php
@@ -35,17 +35,11 @@ class WebhookUpdatePost extends UpdateUsecase
         // ... persist the changes
         $this->repo->updateFromService($entity);
 
-        // ... check that the entity can be read by the current user
-        if ($this->auth->isAllowed($entity, 'read')) {
-            // ... and either load the updated entity from the storage layer
-            $updated_entity = $this->getEntity();
+        // ... and either load the updated entity from the storage layer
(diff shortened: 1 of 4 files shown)


src/Core/Usecase/Post/WebhookUpdatePost.php
```diff
@@ -35,17 +35,11 @@ public function interact()
         // ... persist the changes
         $this->repo->updateFromService($entity);
 
-        // ... check that the entity can be read by the current user
-        if ($this->auth->isAllowed($entity, 'read')) {
```
rjmackay: @willdoran this check was always failing, because we never have a user session on this request. Is the desired result to always return an empty 204 response? or to always return a 200 with the updated post? I've gone with 200, but I suspect it was previously a 204

willdoran: A 200 is fine, now that we have lumen the service proxy will hopefully go away completely and live inside Core. Right now it does nothing with the response unless it fails in which case it will retry 3 times.

Issue #3027, opened 2018-05-30
User not receiving reset email after resetting password


StaicyG:
### Expected behaviour
Requests reset password
### Actual behaviour
Receives notification below but never receives a reset email
![image.png](https://images.zenhubusercontent.com/5a8e72064b5806bc2bc7edf1/508de79a-a9ec-4d4d-b19b-cf3b867b52b1)
### Steps to reproduce the behaviour/error
Clicked on forgot password 
Keyed in my email address
Got the notification similar to the one on the screenshot
Did not receive a password reset email

#### Where does this issue happen?

- [ x ] Ushahidi.io / SaaS solution (example https://juantest.ushahidi.io/)

Public as well as private deployments.

jrtricafort:
@StaicyG which deployment is experiencing this issue? Can you provide a link and a username that we can use to test? We would need this so we can help debug. Thanks!

jrtricafort:
@rowasc Could we ask someone on the team to look into this? I believe this is P0 since there is no workaround if the user forgot their password.

Btw, this looks similar to issue #2608 except that 2608 is particularly for private deployments. In this case, I was able to reproduce the bug even on a non-private deployment: https://juantest.ushahidi.io/

rowasc:
Checking it . I'll report back in a bit

rowasc:
@jrtricafort also to clarify, this is not like #2608 since in this ticket we just don't get an email (at least for me ), while in 2608 the email was received but the link did not work. I'm not sure what's causing it yet

jrtricafort:
@rowasc Got it. Thanks for investigating.

rowasc:
@rjmackay can you take a look? I can't find any logs pointing to the issue

rjmackay:
It turns out that way back in September 2017, somewhere near this commit https://github.com/ushahidi/platform-cloud-ansible/commit/2e730c483c35b00ec26e470f0bbb168cfb77895d#diff-04a78849dbc24e615f3b8487782d6a2b

We moved the postfix email configuration from the `ushahidi.cloud-common` role into `ushahidi.common`. However we failed to configure `smtp_user` and `smtp_password` correctly and so postfix configuration was skipped. This didn't break anything until we rebuilt the servers for PHP7.

I've fixed this and rerun playbook on v3-qa, staging and production to install postfix

[...]

PR #3032, opened 2018-05-31
fix(UserRepository): Generate password reset token using laravel classes

rjmackay:
Use laravel Str and Hasher to generate password reset tokens

Test checklist:
- [x] bin/phpunit tests/unit/App/Repository/UserRepositoryTest.php

Refs ushahidi/platform#3027

Ping @ushahidi/platform


src/App/Repository/UserRepository.php
```diff
@@ -188,9 +190,7 @@ public function register(Entity $entity)
 	// ResetPasswordRepository
 	public function getResetToken(Entity $entity)
     {
-		// Todo: replace with something more robust.
-		// This is predictable if we don't have the openssl mod
-		$token = Security::token(true);
+		$token = Hash::make(Str::random(40));
```
rjmackay: I'm not really sure if hashing is necessary here, since I'm not verifying the hash. We could probably just use a 40 character random string.

The laravel built in reset tokens use a hash, but they also verify it. We could do that too, ie:
- Generate a string.
- Hash it and save that to the db
- Return the original token string to the caller (ie. usecase that then sends an email)
- Update our verification to verify the hash, not simply a token match

rjmackay:
@rowasc @willdoran I originally sent this PR against the lumen release branch. I'm not sure if thats still correct. Do we plan to release just lumen? or branch a new release off of develop when we're ready?

src/App/Repository/UserRepository.php
willdoran: So I think like with reset tokens it's best to gen a really really really random string, then save the hash in the db. This is probably best so that the token can't just be read out of the DB. We need to make sure that random() is legitimately random. I think we need to make sure openssl_random_pseudo_bytes is installed otherwise rand just uses quickRandom which is a bit meh.

So that's a long way of saying, yeah the above list of steps sounds good.

willdoran:
@rjmackay my understanding was we would work with a release branch built from develop is that correct @rowasc ?

rjmackay:
I'll rebase this off develop then :)

rowasc:
@rjmackay @willdoran yes we will need a new release branch off develop

rjmackay:
@rowasc adding this to cycle 8 because it needs fixing before lumen deploys to prod, and Will requested further work in the comments

PR #3098, opened 2018-06-20
fix(ResetUserPassword) 

rowasc:
- Before checking if a token is valid , it runs urldecode on it since the email goes out with a urlencode 

Test checklist:
- [ ] Click on the Login button
- [ ] In the login modal, click on "forgot your password? "
- [ ] In the "Forgot your password" modal, set your email and hit the "RESET MY PASSWORD" button.
- [ ] Wait for the password reset email. When you get the email, click on the "Reset your password" button.
- [ ] In the Reset password modal, try to setup your new password.
- [ ]  You should be able to reset your password and login with it

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/src/Core/Usecase/User/ResetUserPassword.php
+++ b/src/Core/Usecase/User/ResetUserPassword.php
@@ -61,8 +61,7 @@ class ResetUserPassword implements Usecase
 
 	public function interact()
 	{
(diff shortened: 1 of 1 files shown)


willdoran:
Looks fine

PR #3118, opened 2018-06-28
feat(HDXInterface): create slug with str_slug

rowasc:
-

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


src/App/ExternalServices/HDXInterface.php
```diff
@@ -91,12 +95,11 @@ public function getDatasetIDByName($title)
      * @return array
      * Create dataset object based on the parameters we received from create/update
      */
-    private function formatDatasetObject(array $metadata, $license, $tags = [])
+    public function formatDatasetObject(array $metadata, $license, $tags = [])
```
rowasc: sooo, I made this public because I wanted to test the format was OK.. but there might be a better way to do it without this

rowasc: @willdoran I'm not sure what to do here. Usually, I would not test this and leave it private, but since we want to check the slug generates correctly, and the function that calls it requires that we mock the request, it seemed like a good idea to test it directly. Thoughts?

willdoran: I think that should be fine all the other functions in that class are public anyway so I don't think it makes that much of a difference

rowasc:
@willdoran I updated to fix the tests after merging develop. Can you review again ?

Issue #2986, opened 2018-05-23
[Platform API] Move Maintainer field to User Settings


willdoran:
### Overview
The maintainer field should be moved to be closely linked with the API key.

### Requirements
- [x] Maintainer field is removed HXL Meta Data table
- [x] Maintainer field is removed from HXL Meta Data entity and Validation
- [x] Maintainer field is renamed as hdx_user_id 
- [x] hdx_user_id is substituted throughout API
- [x] hdx_user_id is now retrieved from User settings.

## Acceptance criteria / Test script
- [x] hdx_user_id is returned correctly from User settings

rowasc:
@jrtricafort I think this should be part of cycle 6, as not doing it raises a requirement to ask the user to re-validate their user id when generating an HDX export, which is more complex than just moving this to the correct screen. 
We still  need to improve our export validation but this is an 'easy' improvement that would make it way more usable and allow users to make mistakes without having to re-map their export . 
cc @willdoran

jrtricafort:
Agreed, added. to Cycle 6.

rowasc:
@willdoran we should add that the server will validate the API Key against HDX  (so far I figured out that if I do a /user_show for a user and it returns the API key + all data, it means it is the same user) and return an error if they don't match

willdoran:
@rowasc Can you review please?

rowasc:
@willdoran APPROVED.

PR #3123, opened 2018-06-29
Changing to use hdx_maintainer_id from settings

willdoran:
-

Test checklist:
- [ ]

Fixes ushahidi/platform#2986

Ping @ushahidi/platform


src/App/Repository/Config/features.php
```diff
@@ -82,11 +82,11 @@
...
     // We will need a new 'hxl-download' flag when we do the HXL downloads for P1
     'hxl' => [
-        'enabled' => false,
+        'enabled' => true,
     ],
 
     // Enable or disable User Settings feature
     'user-settings' => [
-        'enabled' => false,
+        'enabled' => true,
```
rowasc: should be false for hxl and user-settings here

src/App/Validator/HXL/Metadata/Create.php
```diff
@@ -60,12 +60,6 @@ protected function getRules()
                 ['min_length', [':value', 2]],
                 ['max_length', [':value', 255]],
                 ['regex', [':value', Validator::REGEX_STANDARD_TEXT]], // alpha, number, punctuation, space
```
rowasc: this still applies. Why was it removed?

src/App/ExternalServices/HDXInterface.php
```diff
@@ -21,12 +21,14 @@ class HDXInterface
 {
     protected $ckanClient;
     protected $userAPIKey;
+    protected $hdx_user_id;
```
rowasc: I know I approved the spec, and this matches it, so feel free to ignore this and add a tech debt task for me... but I really think this should be called hdx_maintainer. There's an id in HDX for users that they return in their API (different than maintainer) and it opens things up for more confusion if we use $hdx_user_id for what is actually the maintainer.

src/Core/Usecase/HXL/Organisations/GetByUser.php
```diff
@@ -67,21 +67,26 @@ public function isSearch()
...
         return $this->formatter->__invoke($organisations);
     }
 
-    private function setHDXInterface($user_settings)
+    private function setHDXInterface($user_settings_key, $user_settings_user_id)
     {
         $this->hdxInterface = new HDXInterface(
             getenv('HDX_URL'),
-            $user_settings->config_value
+            $user_settings_key->config_value,
```
rjmackay: I think it might be better to restructure these values before passing them to `setHDXInterface` ie. only pass the config_value through, not the entire object. 
It's sort of weird that this method has to know about the config structure at all.

rjmackay:
I know there's lots of detail in the spec on #2986 but can we make sure pull requests still actually have a relevant description? and if not completing section - ie. test checklist - either delete them or reference a checklist elsewhere?

src/Core/Usecase/HXL/Organisations/GetByUser.php
willdoran: That falls under refactor which is covered by a different ticket so I'm not going to change that here.

src/App/ExternalServices/HDXInterface.php
willdoran: ok that's a little annoying but it's just a find and replace so I can do it today.

rowasc:
@willdoran this looks good, I'll test with the client changes and get back to you today

willdoran:
@rowasc thank you!

PR #3127, opened 2018-06-30
WIP - adding a console script to obfuscate data within current deployment DB

kinstelli:
- Adds a console script executable by `artisan db:obfuscatedata` that:
      - overwrites contacts
      - overwrites authors of posts
      - removes dataprovider values
      - overwrites some site settings
      - deletes all user accounts
      - adds a default admin user

Test checklist: 
WIP - Tests forthcoming


src/Console/Command/ObfuscateData.php
```diff
@@ -0,0 +1,225 @@
...
+            $this->db->commit();
+        } else {
+            $this->info("This script will only run on test or staging deployments.\n\n");
+            exit;
+        }
+    }
+
+    private function isTestDeployment()
+    {
+        return (strtolower(getenv('APP_ENV')) == 'test');
```
rjmackay: You should use app()->environment for these checks

rjmackay: Maybe add a static property of `allow_environments` and check against that?

src/Console/Command/ObfuscateData.php
```diff
@@ -0,0 +1,225 @@
...
+    {
+        $this->db = service('kohana.db');
+        $this->contactRepository = service('repository.contact');
+        $this->postRepository = service('repository.post');
+        $this->configRepository = service('repository.config');
+        $this->userRepository = service('repository.user');
+
+        if ($this->isTestDeployment() || $this->isStagingDeployment()) {
+            //confirm acknowledgement
+            $this->info("This script will wipe data from the current deployment.\n\n");
```
rjmackay: This should probably be a confirm too right. ie. `This script will wipe user, contacts, post author and data source data. Do you want to continue?`

src/Console/Command/ObfuscateData.php
```diff
@@ -0,0 +1,225 @@
...
+        $faker = Faker\Factory::create();
+        $query = DB::select('contacts.id', 'contacts.type', 'contacts.contact')
+            ->distinct(true)
+            ->from('contacts');
+        $results = $query->execute($this->db);
+        $overwrittenCount = 0;
+        $resultsCount = count($results);
+        foreach ($results as $row) {
+            $contactEntity = $this->contactRepository->getEntity($row);
+            $fakePhone = $faker->e164PhoneNumber;
```
rjmackay: Not sure if you just haven't got to this yet, but contacts can be emails or twitter handles too

src/Console/Command/ObfuscateData.php
```diff
@@ -0,0 +1,225 @@
...
+    {
+
+    }
+    */
+
+    private function overwriteContacts()
+    {
+        $this->info("Overwriting contact data...");
+        $faker = Faker\Factory::create();
+        $query = DB::select('contacts.id', 'contacts.type', 'contacts.contact')
```
rjmackay: Do you think this should all move into contacts repo? or stay here as a read/write step so that ContactRepo doesn't have to depend on Faker?
I can't actually decide which makes more sense

src/Console/Command/ObfuscateData.php
```diff
@@ -0,0 +1,225 @@
...
+            return false;
+        }
+    }
+
+    private function overwriteSiteConfig()
+    {
+        $this->info("\nORemoving general settings data...");
+        $siteConfig = $this->configRepository->get('site');
+        $siteConfig->setState([
+            'name' => 'Test site',
```
rjmackay: Not sure if we want to wipe the name and description. Maybe just prefix them with `SANITIZED` or similar. It'd be useful to still know what deployment you're looking at data from.

src/Console/Command/ObfuscateData.php
```diff
@@ -0,0 +1,225 @@
...
+            'email'=> [],
+        ]);
+        $this->configRepository->update($dataProviderConfig);
+        $this->info("\nSet all dataproviders to false.");
+    }
+
+    private function deleteUsers()
+    {
+        $this->info("Deleting users...");
+        $query = DB::delete('users');
```
rjmackay: I think we should replace users with junk data, rather than wiping them outright. It makes for much more useful test data ie. many users not just a single user.

src/Console/Command/ObfuscateData.php
```diff
@@ -0,0 +1,225 @@
...
+    //TODO: unhardcode this if this is a feature we want
+    private function addAdminUser()
+    {
+        $newUserEntity = $this->userRepository->getEntity();
+        $newUserEntity->setState([
+            'email' => 'admin@ushahidi.com',
+            'realname' => 'Admin User',
+            'role' => 'admin',
+            'password' => 'admin' ]); // NOTE: pw is implicitly hashed via create
+        $id = $this->userRepository->create($newUserEntity);
```
rjmackay: Just make sure to handle the state where admin@ushahidi.com already exists

src/Console/Command/ObfuscateData.php
kinstelli: I'm not sure either. I suppose each of these repos could have an obfuscation method, and possibly even a fake data generation method.

src/Console/Command/ObfuscateData.php
kinstelli: True. I'm trying to decide if we should then just scramble passwords, or set them all to one obvious thing.

src/Console/Command/ObfuscateData.php
kinstelli: I'd assumed admin@ushahidi.com would have been wiped along with all other users, but I suppose we could just *not* wipe that account if it's there -- or ask in-line what the person running the script prefers.

src/Console/Command/ObfuscateData.php
kinstelli: Left this method in place, in case we want to add this as an option.

src/Console/Command/ObfuscateData.php
kinstelli: @rjmackay - Not sure if app() is available from within a Console command, but I can access with `$this->getLaravel()->environment();`
Am I doing that wrong?

rjmackay: I think `app()` is available everywhere as a helper function. I don't know if `this->app` works in commands, but getLaravel() seems reasonable. 
Otherwise check the docs here: https://lumen.laravel.com/docs/5.1#custom-configuration-files

src/Console/Command/ObfuscateData.php
rjmackay: I think scramble passwords unless its too slow to work. In which case set them all to the same things but make it long and random

src/Console/Command/ObfuscateData.php
rjmackay: True. If we've already scrambled the email it won't be there. That said if we import say 10% of production deployments to staging and then all of them have a admin user with password admin... thats basically leaving the data open to the public. It might be easier just not to add the user since we have another console command that can do that if we need to.

Issue #1618, opened 2017-03-06
USH-023 — Internal Server Error While Uploading Photo


willdoran:
### Expected behaviour
- Don't show (detailed) error messages to users.
- Validate the input.
### Actual behaviour
- An internal server error appears when using a NULL byte in the filename. This reveals the underlying host where the images are located.
- Server throws a 500, not 422

```
{"errors":[{"status":0,"title":"Client error response\n[status code] 412\n[reason
 phrase] Precondition Failed\n[url] https:\/\/storage101.ord1.clouddrive.com
\/v1\/MossoCloudFS_e606dc27-4d2f-4bbb-8ea9-ff12ba538332\/platform-cloud\/svink.api.ush
ahidi.io\/5\/8\/58a6821538ad0-test.png%00","message":"Cli
ent error response\n[status code] 412\n[reason phrase] Precondition Failed\n[url]
 https:\/\/storage101.ord1.clouddrive.com\/v1\/MossoCloudFS_e606dc2
7-4d2f-4bbb-8ea9-ff12ba538332\/platform-cloud\/svink.api.ushahidi.io\/5\/8\/58a6821538
ad0-test.png%00"}]}
```

### Steps to reproduce the behaviour/error
- [ ] Simulate a file upload where the file name is `test.png%00`
- [ ] Server throws an error
- [ ] Error shouldn't contain paths.

## Implementation

- Validate file name in Media Create validator
- Modify Ushahidi\Core\Tool\Uploader
    - Add try/catch blocks around `fs` calls where needed and ensure we are catching error
    - Throw appropriate errors ie. validation or client errors.
- Modify CreateMedia usecase:
    - Move `$this->uploader->upload` to try/catch block. 
    - If it fails, trigger a Validator exception (or something else where appropriate)

Caveat:
- Uploader is modified/deprecated in lumen. If we can resolve this issue solely within the usecase - do that.

rowasc:
Does this need a more detailed spec or is it good to go?
@rjmackay @willdoran ?

rjmackay:
I think it's straight forward. Error messages shouldn't include paths, in this case I think its because we send the flysystem error back in the response. We can drop that.

rjmackay:
@rowasc added more to the spec. Is that sufficient?

rjmackay:
I'll bump it to a 2 just to account for checking the develop and master branches for the issue

rowasc:
[...]

PR #3133, opened 2018-07-05
Fix server error while uploading photos

rjmackay:
- Sanitize filenames when uploading files
- Handle errors from flysystem properly when uploading
- Fix some namespacing issues in CreateMedia

Test checklist:
- [ ] Create a post with an image
- [ ] Image should upload normally
- [ ] Send a media upload request with a null byte in the filename - see audit report for details
- [ ] Image should upload normally or throw an error about invalid extension.
- [ ] No path data should be leaked in the response

Fixes ushahidi/platform#1618

Ping @ushahidi/platform

### DIFF
--- a/src/Core/Tool/Uploader.php
+++ b/src/Core/Tool/Uploader.php
@@ -44,9 +44,10 @@ class Uploader
 			$filename = uniqid() . '-' . $file->name;
 		}
 
-		// Avoid any possible issues with case sensitivity by forcing all files
-		// to be made lowercase.
-		$filename = strtolower($filename);
+		// Avoid possible issues with case sensitivity by forcing all files
+		// to be made lowercase. Also replace possibly invalid characters in filename
+		$filename = strtolower(preg_replace('/[^\pL\pN\-\_\s\.]+/u', '', $filename));
+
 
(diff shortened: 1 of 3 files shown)


src/Core/Tool/Uploader.php
```diff
@@ -44,9 +44,10 @@ public function upload(UploadData $file, $filename = null)
 			$filename = uniqid() . '-' . $file->name;
 		}
 
-		// Avoid any possible issues with case sensitivity by forcing all files
-		// to be made lowercase.
-		$filename = strtolower($filename);
+		// Avoid possible issues with case sensitivity by forcing all files
+		// to be made lowercase. Also replace possibly invalid characters in filename
+		$filename = strtolower(preg_replace('/[^\pL\pN\-\_\s\.]+/u', '', $filename));
```
rowasc: this made my brain hurt, but it seems to be correct. Just wondering...are we planning to use the same for lumen?

rjmackay: The regex is actually straight forward but the character classes are confusing: `\p` and `\N` are shorthand for all unicode characters and numbers, not just latin ones. So it replaces everything that isn't letters, number - _ or whitespace

PR #3136, opened 2018-07-06
Feature/hdx dataset tags

rowasc:
- Test with https://github.com/ushahidi/platform-client/pull/1196
- A CSV with geo tags should be generated. The lat/lon fields should each be aligned with their hdx tag+attribute. 

Test checklist:
- [x] Generate a csv with a field that points a location attribute to  #geo+lat (you'll automatically get #geo+lon if doing it from the client). When the export is complete, the fields should be correctly aligned in the CSV (+lat matching .lat, +lon matching .lon).
- [x] Generate a csv with a field that points a location attribute to  #geo+lat+elevation, #geo+lon+elevation from the client. When the export is complete, the fields should be correctly aligned (+lat matching .lat, +lon matching .lon)

Fixes ushahidi/platform# .

Ping @ushahidi/platform


src/App/Formatter/Post/CSV.php
```diff
@@ -67,26 +67,64 @@ public function generateHXLRows($header_rows, $hxl_rows)
...
         return $tag;
     }
 
+    private function shouldProcessLatLonTag($key_set, $hxl_row)
+    {
+        $should_process_lat_lon =
+            $hxl_row['tag_name'] === 'geo' &&
+            $hxl_row['type'] === 'point' &&
+            $hxl_row['input'] === 'location';
+        if (!$should_process_lat_lon) {
```
willdoran: Why is this an inverse?

rowasc: ah good point.. should just be
+        $should_not_process_lat_lon =
+            $hxl_row['tag_name'] !== 'geo' &&
+            $hxl_row['type'] !== 'point' &&
+            $hxl_row['input'] !== 'location';
I'll change it

rowasc:
@willdoran fixed negated condition, can you check again?

src/App/Formatter/Post/CSV.php
```diff
@@ -67,26 +67,64 @@ public function generateHXLRows($header_rows, $hxl_rows)
...
-                $tag = $tag.$attribute;
+            //contains key + index of the key
+            $key_set = explode('.', $form_attribute_key);
+            // the heading type (sets, contact, title)
+            $heading_key = isset($key_set[0]) ? $key_set[0] : null;
+
+            // checks that the hxl field matches the form attribute we received in $form_attribute_key
+            $key_matches_hxl = $hxl_row['key'] === $heading_key;
+            $should_process = $key_matches_hxl && $this->shouldProcessLatLonTag($key_set, $hxl_row);
+            if ($should_process) {
```
willdoran: let's move the above line into the conditional, I think we don't need the var $should_process

src/App/Formatter/Post/CSV.php
```diff
@@ -67,26 +67,64 @@ public function generateHXLRows($header_rows, $hxl_rows)
...
-                if ($tag === '') {
-                    $tag = '#' . $hxl_row['tag_name'];
-                }
-                $attribute = $hxl_row['attribute'] && !empty($hxl_row['attribute']) ?
-                        ' +'. $hxl_row['attribute'] : '';
-                $tag = $tag.$attribute;
+            //contains key + index of the key
+            $key_set = explode('.', $form_attribute_key);
+            // the heading type (sets, contact, title)
+            $heading_key = isset($key_set[0]) ? $key_set[0] : null;
```
willdoran: Can these two lines move outside the loop?

src/App/Formatter/Post/CSV.php
```diff
@@ -67,26 +67,64 @@ public function generateHXLRows($header_rows, $hxl_rows)
...
         return $tag;
     }
 
+    private function shouldProcessLatLonTag($key_set, $hxl_row)
+    {
+        $should_not_process_lat_lon =
+            $hxl_row['tag_name'] !== 'geo' &&
+            $hxl_row['type'] !== 'point' &&
+            $hxl_row['input'] !== 'location';
+        if ($should_not_process_lat_lon) {
```
willdoran: I think the above can move into the if's conditional. I see the purpose of the var but we should use a comment instead if the conditional is complex.

rowasc: Alright. Changing now. I disagree as I prefer to read a conditional as a var instead of relying on a comment that may or may not be obsolete, but also see the other side. ..

src/App/Formatter/Post/CSV.php
rowasc: yup

src/App/Formatter/Post/CSV.php
rowasc: sure

rowasc:
@willdoran updated, thanks!

willdoran:
I approved as the comments are only suggestions up to you if you want to do anything about them!

rowasc:
@willdoran thanks, I followed the review advice anyway :)

Issue #3082, opened 2018-06-15
Backport fixes to master


rjmackay:
## Overview
Backport fixes from develop to master, before releasing audit report.

## Requirements
- [x] Backport #2851 
- [x] Backport #2876
- [x] Backport #2896

### Roles:
- [x] With Roles feature enabled set to false
- [x] Attempt to create/update a new role via HTTP Post
- [x] You should receive a 422

### Password reset
Using link
- [x] Open a deployment where you have an account and go to login
- [x] Select "forgot password"
- [x] Enter email and trigger a password reset
- [x] Receive password reset email
- [x] Click link in password reset email
- [x] Set password to 5 characters
- [x] Returns an error that password is too short
- [ ] Try again with 8 characters
- [ ] Password is set

Using copy/paste:
- [x] Open a deployment where you have an account and go to login
- [x] Select "forgot password"
- [x] Enter email and trigger a password reset
- [x] Receive password reset email
- [x] Copy the token from in password reset email into reset form
- [x] Set password to 5 characters
- [x] Returns an error that password is too short
- [ ] Try again with 8 characters
- [ ] Password is set

### Registration:
- [x] You should not be able to make more than 3 registration attempts per minute

rjmackay:
@rowasc I've added an acceptance checklist.
Are you happy to approve this as is?
I'd suggest point score might be a 3 or 5 because of the complexity of backporting

rowasc:
@rjmackay looks good.
Small fix: " With Roles feature enabled set to falseWith Roles feat" => I think this should be " With Roles feature enabled set to false"


I'd also add that the user should be able to copy and paste the token to the screen where you have the option to paste the token and add your new password (right after the forgot password modal) . It's a bit obvious but since we had a bug with the reset token not being decoded and failing if you just clicked the link, it might make sense to add the opposite behavior (copy-pasting the token) in the test script here as well. 



-----
Other than those small comments, I'd approve this , but probably a 5 because it's still 3 issues that need to be backported and tested and released.

rjmackay:
- Registration rate limit tests pass
- password reset validation works, but not exactly as in the test script:
  - client validation prevent reset with the wrong thing
  - it looks like maybe you can't reuse the token after the first attempt to reset

PR #3137, opened 2018-07-06
Adding reset password validation

rjmackay:
Backport https://github.com/ushahidi/platform/pull/2876 

Fixes ushahidi/platform#3082

Ping @ushahidi/platform


src/Core/Usecase/User/ResetUserPassword.php
```diff
@@ -62,22 +64,28 @@ public function setRepository(ResetPasswordRepository $repo)
...
-		} else {
-			// Otherwise throw an exception
-			throw new ValidatorException('Invalid or expired token', ['Invalid or expired reset token']);
-		}
+		$this->verifyValid($entity_array);
+			
+		$this->repo->setPassword($token, $password);
+
+		// And delete the token
+		$this->repo->deleteResetToken($token);
```
willdoran: I'm not sure how much of an issue it is but may be we should delete the token before we save the password, or does it matter?

Issue #1607, opened 2017-03-06
USH-012 — Lack of Bruteforce Protection new user.


willdoran:
### Expected behaviour
- Captcha
- IP limit or lockout for a certain amount of time
### Actual behaviour
- There is no protection against the automatic creation of useraccounts. It was possible to create 10 accounts in 10 seconds.

### Test script
- Go to a deployment (ie. qa.ushahididev.com)
- In under 1 minute:
    - Register a new user
    - Register a second new user
    - Register a third new user
    - Register a fourth new user
- Assuming this was done quickly, the fourth user should fail

rjmackay:
We have a rate limiter for auth, we should add that to this endpoint.

rjmackay:
Tagged this to cycle 4 since its current in progress

rowasc:
@willdoran what should we do about this ticket?  move to ready for dev to port to kohana?

rjmackay:
It still needs to land to develop. We should fix the conflicts and land it. Then close this issue. Backports are tracked in their own issue.

rjmackay:
I've rebased the existing PR. It just needs tests to run then it can land and go to QA

rjmackay:
@Eve-wanderi I've added a test script. It's possible this might be too hard to test manually - ie. registration is too slow through the UI so you never hit the rate limit. If so let me know and I can try to test directly on the API

Eve-wanderi:
@rjmackay Yea its quite a challenge for me to Do all four Registrations for a new user in less than 1 minute. Since am doing this manually. Could you please assist testing this directly on the API.

PR #3155, opened 2018-07-10
fix: Handle ThrottlingException and throw a HTTP 429

rjmackay:
Catch ThrottlingExceptions and turn them into HTTP 429. Previously we were just throwing a 500 error

Refs #1607

Test checklist:
- [x] bin/behat tests/integration/users.feature:220 --format=pretty


rjmackay:
I fixed this directly in master to unblock release https://github.com/ushahidi/platform/commit/56063fdbd55ff51bef95306d594cd8e5884507b8 - this takes the same fix to the lumen codebase

rjmackay:
@rowasc @willdoran @crcommons could someone review this and merge if you're happy?

Issue #2601, opened 2018-03-14
BUG: CSV Export history should ONLY show exports from the logged in user


rowasc:
### Expected behaviour
- When I see my exports history, I only see the ones that belong to me
### Actual behaviour
- When I see my exports history as an admin, I see all the exports

### Steps to reproduce the behavior/error
- Login to http://csvexportsforever.v3-qa.ush.zone/settings/data-export or another qa deployment from v3-qa.ush.zone
- Go to the exports page
- Go to the history tab
- Expected behavior: You should only see your exports
- Current behavior: You see all exports

##### Where
- [x] [Local setup with Vagrant ](https://www.ushahidi.com/support/install-ushahidi#installing-for-development)
- [ ] [Local setup from platform-release ](https://www.ushahidi.com/support/install-ushahidi#installing-the-latest-release)
- [ ] Ushahidi.io / SaaS solution 
- [x] Ushahidi's QA environment (in v3-qa.ush.zone) 
- [ ] Other (explain):

rowasc:
@Angamanga I couldn't reproduce this.

Angamanga:
@rowasc @caharding Unfortunately, yes I can reproduce it on .io. I created a new user and went to the data-export in settings and started an export. In the export-tab I see all export-jobs that is finished. 

I also see all the export-jobs in the response from the api so its not a caching-issue.

willdoran:
### Spec:

- [ ] Export Job History should show only export jobs from the currently logged in user
- [ ] Export Job Repository should filter export jobs by current user

I believe the reason it was like this is because in order to perform an Export you must be an admin and by default all admins can see everything.

@caharding @Angamanga Instead of restricting the jobs to the current user would it be better to add information to the Job showing which user started the job:
'Created by X'

Angamanga:
@willdoran As long as its only admins that can see other users exports thats ok i think.

Angamanga:
@willdoran Looks good, moving to ready for dev!

jrtricafort:
Needs a new spec. @willdoran

rowasc:
@jrtricafort do you remember why this was flagged as "needs a new spec" ?

jrtricafort:
@rowasc That comment was to capturing something from a conversation I had with @willdoran , but I don't recall the specifics. Will do you remember? If not, I think we can probably ignore that for now, since we have prioritized this as P2 so not as high a priority.

rowasc:
[...]

PR #3161, opened 2018-07-12
2601: Bugfix - show exports only for logged in user

crcommons:
fix: this commit allows users with permisions to access only their own exports, rather than all exports + tests

- Checks who the user is and only gets export jobs for that user
- adds tests 

Test checklist:
- [ ] run behat tests

Fixes ushahidi/platform#2601

Ping @ushahidi/platform

### DIFF
--- a/src/App/Repository/ExportJobRepository.php
+++ b/src/App/Repository/ExportJobRepository.php
@@ -73,11 +73,9 @@ class ExportJobRepository extends OhanzeeRepository implements ExportJobReposito
             $query->where('hxl_meta_data_id', '=', $search->hxl_meta_data_id);
         }
 
-        // Limit search to user's records unless they are admin
-        // or if we get user=me as a search param
-        if (! $this->isUserAdmin($user) || $search->user === 'me') {
-            $search->user = $this->getUserId();
-        }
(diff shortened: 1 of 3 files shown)


src/App/Repository/ExportJobRepository.php
```diff
@@ -73,11 +73,9 @@ public function setSearchConditions(SearchData $search)
             $query->where('hxl_meta_data_id', '=', $search->hxl_meta_data_id);
         }
 
-        // Limit search to user's records unless they are admin
-        // or if we get user=me as a search param
-        if (! $this->isUserAdmin($user) || $search->user === 'me') {
-            $search->user = $this->getUserId();
-        }
+        // get user ID so that we only ever get jobs from that user
+        $search->user = $this->getUserId();
```
crcommons: Just wondering if there's any reason to check the query string, rather than just getting the logged in user? The original test had a query string with a user ID.

willdoran: We should always get the logged in user not trust a query string, though conceivably an Admin could run a request for a user's stuff and in that case we would care about the query string, however, that's not a feature we support at the moment.

PR #3162, opened 2018-07-13
fix(HXLAttributes Migration): remove elevation as an option for hxl tag #geo

rowasc:
- delete 'elevation' from the hxl_tag_attributes table

Test checklist:
- [ ] Select #geo as a tag in a location type field
- [ ]   Check the available attributes. You should not see elevation as an option

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/migrations/20180713183206_remove_elevation_from_hxl_attributes.php
+++ b/migrations/20180713183206_remove_elevation_from_hxl_attributes.php
@@ -5,17 +5,16 @@ use Phinx\Migration\AbstractMigration;
 class RemoveElevationFromHxlAttributes extends AbstractMigration
 {
     /**
-     * Remove the 'elevation' HXL attribute.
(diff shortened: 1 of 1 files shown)


migrations/20180713183206_remove_elevation_from_hxl_attributes.php
```diff
@@ -0,0 +1,21 @@
...
+{
+    /**
+     * Remove the 'elevation' HXL attribute.
+     * This attribute will be added back if we decide to implement elevation correctly
+     * in platform + csv
+     */
+    public function change()
+    {
+        $this->execute(
+            "DELETE FROM hxl_tag_attributes where attribute_id IN (
```
rowasc: I only deleted the reference so it doesn't appear. We are planning to add elevation at some point but we shouldn't see it in the list as an available option

PR #3165, opened 2018-07-16
Bulk exports in master

crcommons:
- Ports changes made in develop to master for csv exports; user can now only see their own csv exports;
- Changes permission to include exports, not just imports

*note*: needs to be QA'd along with changes in the Front End https://github.com/ushahidi/platform-client/pull/1209

Test checklist:
- [ ] Run behat tests

Fixes ushahidi/platform# .

Ping @ushahidi/platform


crcommons:
@rowasc Do I need to update the add permissions migration to "Bulk Data Import and Export"?  https://github.com/ushahidi/platform/blob/master/migrations/20160202115439_add_permissions.php

rowasc:
https://github.com/ushahidi/platform/blob/develop/migrations/20180510172759_data_import_permission_name_change.php This is the migration that adds the permission and renames it

PR #3170, opened 2018-07-20
fix(HXL): hxl tags should not have spaces

rowasc:
-

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/src/App/Formatter/Post/CSV.php
+++ b/src/App/Formatter/Post/CSV.php
@@ -125,7 +125,7 @@ class CSV extends API
             $tag = '#' . $hxl_row['tag_name'];
         }
         $attribute = $hxl_row['attribute'] && !empty($hxl_row['attribute']) ?
-            ' +'. $hxl_row['attribute'] : '';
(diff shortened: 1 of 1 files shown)


src/App/Formatter/Post/CSV.php
```diff
@@ -125,7 +125,7 @@ private function returnHXLTagAndAttributeString($tag, $hxl_row)
             $tag = '#' . $hxl_row['tag_name'];
         }
         $attribute = $hxl_row['attribute'] && !empty($hxl_row['attribute']) ?
-            ' +'. $hxl_row['attribute'] : '';
+            '+'. $hxl_row['attribute'] : '';
```
rowasc: having an extra space made the preview not be completely correct specifically in the map view of HDX . While HDX was indeed rendering points in the map, they were not always correct (ie first point was wrong)

rowasc: #readthedocs, romina

PR #3180, opened 2018-07-25
feat: adds console command to destructively obfuscate selected data f…

kinstelli:
…rom a platform database

Adds a console script executable by artisan db:obfuscatedata that:
- overwrites contacts
- overwrites authors of posts
- removes dataprovider values
- overwrites some site settings
- deletes all user accounts
- adds a default admin user

Test checklist:
WIP - Tests forthcoming

To use manually...
**NOTE: This _WILL_ destroy your data**
- For a single deployment:
`php artisan db:obfuscatedata`
- For shared code on a multisite deployment, this command will provide an error unless you include a HOST envvar, such as in the following:
`HOST=somedeployment.sharedhost.com php artisan db:obfuscatedata `


src/Console/Command/ObfuscateData.php
```diff
@@ -0,0 +1,305 @@
...
+    protected $description = 'Obfuscates selected PII and removes various config keys from database';
+
+    public function __construct()
+    {
+        parent::__construct();
+    }
+
+    public function handle()
+    {
+        //set up repos
```
rjmackay: Not significant, but even though we can't do constructor injection, I'd generally put the `service()` calls in __construct. Just to separate them from the actual handler code

src/Console/Command/ObfuscateData.php
```diff
@@ -0,0 +1,305 @@
...
+        $this->contactRepository = service('repository.contact');
+        $this->postRepository = service('repository.post');
+        $this->configRepository = service('repository.config');
+        $this->userRepository = service('repository.user');
+
+        $this->db = service('kohana.db');
+
+        if ($this->isThisAMultisiteInstall()) {
```
rjmackay: Is there something we can call usefully on the Multisite service to answer this? and check if we have `HOST` define?

If not, maybe move this code there instead.

src/Console/Command/ObfuscateData.php
```diff
@@ -0,0 +1,305 @@
...
+            $this->overwriteContacts();
+            $this->overwritePostAuthors();
+            $this->removeDataProviderValues();
+            $this->overwriteSiteConfig();
+            $this->deleteUsers();
+            $this->addAdminUser();
+            $this->db->commit();
+        } else {
+            $this->info("This script will only run on test or staging deployments.\n\n");
+            exit;
```
rjmackay: I wonder if this should be exit or `return`? Does laravel do any handling after the `handle` call that gets skipped by `exit`?

src/Console/Command/ObfuscateData.php
```diff
@@ -0,0 +1,305 @@
...
+        }
+    }
+
+    protected function isThisAMultisiteInstall()
+    {
+        // @TODO: is this the correct way to check against multisite
+        return (config('multisite.domain'));
```
rjmackay: It should check `multisite.enabled`.

src/Console/Command/ObfuscateData.php
```diff
@@ -0,0 +1,305 @@
...
+                $fakeTwitter = '@fake'.preg_replace("/[^A-Za-z0-9]/", '', $faker->realText(15));
+                $contactEntity->setState(['contact'=> $fakeTwitter]);
+            } else {
+                $fakeText = preg_replace("/[^A-Za-z0-9]/", '', $faker->realText(20));
+                $contactEntity->setState(['contact'=> $fakeText]);
+            }
+            $overwrittenCount += $this->contactRepository->update($contactEntity);
+        }
+        if ($overwrittenCount == $resultsCount) {
+            $this->info("updated ".$overwrittenCount." records.\n");
```
rjmackay: Minor but copy should probably be `Updated` not `updated`.

src/Console/Command/ObfuscateData.php
```diff
@@ -0,0 +1,305 @@
...
+                'frontlinesms' => false
+            ],
+            'frontlinesms'=> [],
+            'nexmo'=> [],
+            'smssync'=> [],
+            'twilio'=> [],
+            'email'=> [],
+        ]);
+        $this->configRepository->update($dataProviderConfig);
+        $this->info("\nSet all dataproviders to false.");
```
rjmackay: I might be wrong but I think artisan add newlines for you so you can probably drop all the `\n`s

src/Console/Command/ObfuscateData.php
```diff
@@ -0,0 +1,305 @@
...
+        $this->configRepository->update($siteConfig);
+        $this->info("\nDone.");
+    }
+
+    private function removeDataProviderValues()
+    {
+        $this->info("\nORemoving dataProvider data...");
+        $dataProviderConfig = $this->configRepository->get('data-provider');
+        //TODO: walk through each key instead?
+        //TODO:  otherwise set them to defaults?
```
rjmackay: If you can easily get defaults, that would be better. But I think this is fine if not.

src/Console/Command/ObfuscateData.php
```diff
@@ -0,0 +1,305 @@
...
+        $dataProviderConfig->setState([
+            'url' => '',
```
rjmackay: Don't think `url` is actually part of the data right?

src/Console/Command/ObfuscateData.php
```diff
@@ -0,0 +1,305 @@
...
+            $this->info("failed to overwrite all records.\n");
+            return false;
+        }
+    }
+
+    private function overwritePostAuthors()
+    {
+        $this->info("Overwriting post author data...");
+        $faker = Faker\Factory::create();
+        $results = DB::select('posts.*')
```
rjmackay: I know we couldn't decide on where this code should live before, but looking at it now I think we'd be better to move all the DB code into an `obfuscate` method on the repo, that just returns the number of records processed (or an exception if it fails).

src/Console/Command/ObfuscateData.php
```diff
@@ -0,0 +1,305 @@
...
+    }
+    */
+
+    /* TODO: would this be useful?
+    private function scrambleLongitudeLatitude($distanceFromOrigin = null)
+    {
+    }
+    */
+
+    private function overwriteContacts()
```
rjmackay: This method is particularly dense. A bit of spacing/comments would help just for easier reading

kinstelli: updated

src/Console/Command/ObfuscateData.php
kinstelli: updated

src/Console/Command/ObfuscateData.php
kinstelli: fixed

src/Console/Command/ObfuscateData.php
kinstelli: fixed

src/Console/Command/ObfuscateData.php
```diff
@@ -122,7 +133,39 @@ private function scrambleLongitudeLatitude($distanceFromOrigin = null)
...
+
+        $resultOtherCount = LaravelDB::table('contacts')
+            ->where('type', '!=', 'twitter')
+            ->where('type', '!=', 'email')
+            ->where('type', '!=', 'phone')
+            ->update(['contact' => $randomTextGenerator]);
+
+        $totalChangedCount = ($resultEmailCount + $resultPhoneCount + $resultTwitterCount + $resultOtherCount);
+
+        $this->info("Updated ".print_r($totalChangedCount, true)." contact records.");
```
rjmackay: Why the `print_r`?

src/Console/Command/ObfuscateData.php
```diff
@@ -0,0 +1,360 @@
...
+    public function handle()
+    {
+        if ($this->isThisAMultisiteInstall()) {
+            if (!getenv('HOST') || strlen(getenv('HOST')) < 1) {
+                $this->alert("ERROR: A host must be specified for a multisite deployment.");
+                return;
+            }
+        }
+
+        if ($this->isTestEnvironment() || $this->isStagingEnvironment()) {
```
rjmackay: We might want to replace this with a similar method to whats used in deployments:import

ie.
```
            if (!$this->getLaravel->environment($this->allowedEnvironments) && $this->options('overwrite-force')) {
                $this->warn('Overwriting existing deployments in a production environment');
            }
```

src/Console/Command/ObfuscateData.php
```diff
@@ -0,0 +1,354 @@
...
+    public function handle()
+    {
+        if ($this->isThisAMultisiteInstall()) {
+            if (!getenv('HOST') || strlen(getenv('HOST')) < 1) {
+                $this->alert("ERROR: A host must be specified for a multisite deployment.");
+                return;
+            }
+        }
+
+        if ($this->getLaravel()->environment($this->allowedEnvironments)) {
```
kinstelli: Still refusing to allow this to run anywhere but staging* and test*, with no overrides.

src/Console/Command/ObfuscateData.php
```diff
@@ -0,0 +1,354 @@
...
+     */
+    protected $name = 'db:obfuscatedata';
+
+    /**
+     * The console command signature.
+     *
+     * @var string
+     */
+    protected $signature = 'db:obfuscatedata
+        {--use-faker : Uses the slower Faker library to generate realistic data }';
```
rjmackay: Nice :)

src/Console/Command/ObfuscateData.php
rjmackay: I think thats OK for now. I might need to have the environment value in ansible to include staging or test in it

src/Console/Command/ObfuscateData.php
```diff
@@ -0,0 +1,354 @@
...
+            if (!getenv('HOST') || strlen(getenv('HOST')) < 1) {
+                $this->alert("ERROR: A host must be specified for a multisite deployment.");
+                return;
+            }
+        }
+
+        if ($this->getLaravel()->environment($this->allowedEnvironments)) {
+            //confirm acknowledgements
+            $this->alert("WARNING: This script will wipe user, contacts, post author and data source data.");
+            if (!$this->confirm("Do you want to continue?")) {
```
rjmackay: Can we have a `--force` option to skip this and next check too?

src/Console/Command/ObfuscateData.php
```diff
@@ -0,0 +1,354 @@
...
+                $this->info("Request canceled.");
+                return;
+            }
+            if (!$this->confirm('Do you acknowledge that this data will still contain sensitive data?')) {
```
rjmackay: TBH I think we could skip the confirm here, and just make it an warn()

src/Console/Command/ObfuscateData.php
```diff
@@ -0,0 +1,354 @@
...
+    private function scrambleLongitudeLatitude($distanceFromOrigin = null)
+    {
+    }
+    */
+
+    protected function overwriteContacts()
+    {
+        $this->info("Overwriting post author data...");
+        // @codingStandardsIgnoreLine
+        $randomEmailGenerator = LaravelDB::raw("CONCAT(lpad(conv(floor(rand()*pow(26,8)), 10, 36), 8, 0),'@',LEFT(UUID(), 8),'.example.com')");
```
rjmackay: Can you add quick comments to explain what each of these does?

PR #3184, opened 2018-07-27
299/feat maintenance mode

willdoran:
- Adds fix for CORS header issue when Exceptions are raised before the Middleware stage of the Lumen Request cycle is reached
- Adds more specific error messages to explain failures on the API related to migration or deployment status
- Adds ability to place the API into Maintenance Mode and return this information to the client.

## Acceptance criteria

Test .env param for maintenance - on a local deployment
1. Set `MAINTENANCE_MODE=1` in your `.env` file
2. Load `api.ushahidi.test/api/v3/config/site` 
    - Should return a 503 with "SITE is down for maintenance"
3. Load client UI
    - See "SITE is down for maintenance" message.
4. Remove MAINTENANCE_MODE from .env
5. Load `api.ushahidi.test/api/v3/config/site` 
    - Should return 200 and config results

Test new deployment building messages:
1. Go to ushahididev.com/create (or equivalent)
2. Create a new deployment
3. Immediately try to load http://mydeployment.api.ushahididev.com/api/v3/config
    - Should return a 503 with "SITE is still being created"
4. Reload http://mydeployment.api.ushahididev.com/api/v3/config in 5 mins (ie. when it should be built)
    - Get 200 and site config data
5. Check cloud db deployments table and confirm `deployment_date` was set

Test maintenance mode:
1. Edit deployment entry in cloud.deployments and set status=maintenance
2. Try to load http://mydeployment.api.ushahididev.com/api/v3/config
    - Should return a 503 with "Site is down for maintenance"
3. Edit deployment entry in cloud.deployments and set status=migrating
4. Try to load http://mydeployment.api.ushahididev.com/api/v3/config
    - Should return a 503 with "Site is down for maintenance"
3. Edit deployment entry in cloud.deployments and set status=deployed
4. Try to load http://mydeployment.api.ushahididev.com/api/v3/config
    - Get 200 and site config data

Fixes [ushahidi/299](https://github.com/ushahidi/platform-cloud-interface/issues/299)

Ping @ushahidi/platform


tests/integration/cors.feature
```diff
@@ -1,30 +1,30 @@
...
+    
     Scenario: Preflight config request
         Given that I want to make an OPTIONS request
         And that the request "Origin" header is "http://local.dev"
         And that the request "Access-Control-Request-Method" header is "POST"
         And that the request "Access-Control-Request-Headers" header is "Authorization, Content-Type, Accept"
         When I request "/posts"
         Then the guzzle status code should be 200
-        And the "Access-Control-Allow-Origin" header should be "*"
+        And the "Access-Control-Allow-Origin" header should be "http://local.dev"
```
willdoran: The new version of the cors lib enforces returning the origin instead of *, this is correct according to the w3c spec but so is returning *

composer.json
```diff
@@ -40,7 +40,7 @@
         "beheh/flaps": "dev-master",
         "doctrine/cache": "^1.6.2",
         "ramsey/uuid": "^3.4.1",
-        "barryvdh/laravel-cors": "^0.9.2",
+        "barryvdh/laravel-cors": "^0.11.0",
```
willdoran: We should add
"asm89/stack-cors": "1.2.0" since we are dependent on it now separate from laravel-cors

app/Exceptions/Handler.php
```diff
@@ -108,9 +113,24 @@ public function render($request, Exception $e)
...
+            // it is necessary to force the inclusion of the CORS headers.
+
+            // This uses the app's CORS config.
+            // This config must be loaded in the bootstrap process
+            // before exception handlers are expected to be used.
+            $options = config('cors');
+
+            // This CorsService relies on Asm89\Stack
+            $cors = new CorsService($options);
+            $response = $cors->addActualRequestHeaders($response, $request);
```
willdoran: We could add a check here to see if the headers are already added.

app/Exceptions/Handler.php
```diff
@@ -108,9 +113,24 @@ public function render($request, Exception $e)
...
+
+            return $response;
         }
 
         return parent::render($request, $e);
```
willdoran: Since we're not changing this one, the change will only take effect for the json requests...we might want to fix this for other requests but it's not crucial.

src/App/Multisite.php
```diff
@@ -135,4 +125,35 @@ public function getClientUrl($host = null)
...
+
+        $status = $deployment['status'];
+        $deployed_date = $deployment['deployed_date'];
+
+        if ($status === 'migrating' && !$deployedDate) {
+            abort(503, "Deployment not ready");
+        }
+        if (($status === 'migrating' && $deployedDate) || $status === 'maintenance') {
+            abort(503, $this->domain . "is down for maintenance");
+        }
```
willdoran: we can rewrite these two conditionals as
```
if ($status === 'migrating') {
    if (!$deployedDate) {
            abort(503, "Deployment not ready");
    } else if ($deployedDate || $status === 'maintenance') {
           abort(503, $this->domain . "is down for maintenance");
    }
}
```

src/App/Multisite.php
```diff
@@ -135,4 +125,35 @@ public function getClientUrl($host = null)
...
     }
+    protected function checkDeploymentStatus($deployment)
+    {
+        // No deployment? throw a 404
+        if (! count($deployment)) {
+            abort(404, "Deployment not found");
+        }
+
+        $status = $deployment['status'];
+        $deployed_date = $deployment['deployed_date'];
```
willdoran: You could drop these two lines and just use the array access in the conditionals. That should reduce some of the codeclimate's complaints about complexity

src/App/Multisite.php
```diff
@@ -135,4 +125,35 @@ public function getClientUrl($host = null)
...
+
+        if ($status === 'migrating' && !$deployedDate) {
+            abort(503, "Deployment not ready");
+        }
+        if (($status === 'migrating' && $deployedDate) || $status === 'maintenance') {
+            abort(503, $this->domain . "is down for maintenance");
+        }
+        if ($status === 'pending') {
+            abort(503, $this->domain . "is not ready");
+        }
```
willdoran: We could combine this into the one above as an else if, I don't know if that's better or not but at least it groups things on the basis of status checks

tests/integration/cors.feature
rjmackay: How do we set what origins are allowed? ie. for .io we need to allow `*.ushahidi.io` for the client to work right?

app/Http/Middleware/MaintenanceMode.php
```diff
@@ -0,0 +1,34 @@
...
+    * Handle an incoming request.
+    *
+    * @param  \Illuminate\Http\Request  $request
+    * @param  \Closure  $next
+    * @param  string|null  $guard
+    * @return mixed
+    */
+    public function handle($request, Closure $next, $guard = null)
+    {
+        $maintenanceMode = getenv('MAINTENANCE_MODE');
```
rjmackay: It's probably better to use laravel's `env('MAINTENANCE_MODE', false)` helper. It also handles turning 'true' in a boolean true.
https://laravel.com/docs/5.6/configuration#environment-variable-types

app/Http/Middleware/MaintenanceMode.php
```diff
@@ -0,0 +1,34 @@
...
+        if ($maintenanceMode === '1') {
```
rjmackay: It'd probably be better just to check for any truthy value, thats much more normal behaviour for boolean environment vars in laravel

bootstrap/lumen.php
```diff
@@ -80,6 +81,11 @@
 |
 */
 
+// Configure CORS package
```
rjmackay: You could probably move this to just after `app->withEloquent`. Otherwise its a bit weird that its in the 'Register Service Provider' block. I actually think semi-standard lumen setup was to have config calls at the start of the bootstrap for issues like this

src/App/Multisite.php
```diff
@@ -135,4 +125,35 @@ public function getClientUrl($host = null)
...
+
+        $status = $deployment['status'];
+        $deployedDate = $deployment['deployed_date'];
+
+        if ($status === 'migrating' && !$deployedDate) {
+            abort(503, "Deployment not ready");
+        }
+        if (($status === 'migrating' && $deployedDate) || $status === 'maintenance') {
+            abort(503, $this->subdomain . " is down for maintenance");
```
rjmackay: Shouldn't this use the site name not subdomain?

rjmackay: ie. `deployment['deployment_name']`

app/Exceptions/Handler.php
```diff
@@ -56,6 +58,9 @@ public function report(Exception $e)
      */
     public function render($request, Exception $e)
     {
+        // @todo we should try app('request') first but we can't guarantee its been created
```
rjmackay: Does this happen when multisite calls `abort()` before everything is actually started?

app/Exceptions/Handler.php
```diff
@@ -108,9 +113,24 @@ public function render($request, Exception $e)
...
             }
 
-            return response()->json([
+            $response = response()->json([
                 'errors' => $errors
             ], $statusCode, $headers);
+
+            // In the circumstance where an exception is raised
+            // before the lumen request cycle reaches the middleware stage,
+            // it is necessary to force the inclusion of the CORS headers.
```
rjmackay: :( I guess we should plan to rework the Multisite stuff into middleware in a future iteration. It's kind of gross that this can ever happen, but clearly out of scope for this PR.

tests/integration/cors.feature
willdoran: It works the same way as before but in the previous version it would return * as the allowed origin where as in this version it returns the request origin - provided it passes the test. We set the origins via the config file, that has remained the same thankfully.

app/Exceptions/Handler.php
willdoran: Yeah, the request is not fully formed at that stage.

app/Exceptions/Handler.php
crcommons: @willdoran Do you mean execute addActualRequestHeaders and then check to see if it worked? Or do you mean to check if they're already there and only run if they're not?

willdoran: Sorry to be more clear, I think we can do something like
if (! $response->headers->has('Access-Control-Allow-Origin')) {
            $response = $cors->addActualRequestHeaders($response, $request);
        }

willdoran: @crcommons ^

crcommons: Thanks! I couldn't tell what you meant from the location of the comment.

tests/integration/cors.feature
rjmackay: Ah ok. That makes sense.
So we would configure an origin like *.ushahidi.io, then a request from qa.ushahidi.io would pass the test and the response would return qa.ushahidi.io?

PR #3190, opened 2018-08-07
DATA_IMPORT_EXPORT

rowasc:
Adding the legacy data import permission so that they don't break the release. This will need to be reverted in the future to stop using the legacy permission entirely

- Renames DATA_IMPORT const to match the name off the permission
- Adds LEGACY_DATA_IMPORT

Test checklist:
- [ ] Test checklist in https://github.com/ushahidi/platform-client/pull/1223 
- [ ] Make sure you test the client and api together

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/src/Core/Entity/Permission.php
+++ b/src/Core/Entity/Permission.php
@@ -15,37 +15,38 @@ use Ushahidi\Core\StaticEntity;
 
 class Permission extends StaticEntity
 {
-	protected $id;
-	protected $name;
-	protected $description;
+    protected $id;
+    protected $name;
+    protected $description;
+    // FIXME: this LEGACY_DATA_IMPORT has to be removed after the prod release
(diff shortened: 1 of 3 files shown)


src/Core/Entity/Permission.php
```diff
@@ -15,37 +15,38 @@
...
-	// StatefulData
-	protected function getImmutable()
-	{
-		return array_merge(parent::getImmutable(), ['name']);
-	}
+    protected $id;
+    protected $name;
+    protected $description;
+    // FIXME: this LEGACY_DATA_IMPORT has to be removed after the prod release
+    const LEGACY_DATA_IMPORT    = 'Bulk Data Import';
```
rowasc: I will remove LEGACY_DATA_IMPORT after the release is tested and complete, and do another release with the correct DATA_IMPORT_EXPORT permission only

PR #3206, opened 2018-08-15
Targeted and Crowdsourced survey stats

crcommons:
- Adds new targeted survey stat to get total response recipients; adds stats for crowdsourced posts; adds ability to filter for both.

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/Http/Controllers/API/Forms/StatsController.php
+++ b/app/Http/Controllers/API/Forms/StatsController.php
@@ -24,6 +24,8 @@ class StatsController extends RESTController
     public function index(Request $request)
     {
         $params = $this->getRouteParams($request);
+        $params['created_after'] = $request->input('created_after');
+        $params['created_before'] = $request->input('created_before');
         $this->usecase = $this->usecaseFactory
             ->get($this->getResource(), 'search')
             ->setIdentifiers($params)
--- a/app/Http/Controllers/API/Forms/StatsController.php
+++ b/app/Http/Controllers/API/Forms/StatsController.php
@@ -24,7 +24,7 @@ class StatsController extends RESTController
     public function index(Request $request)
     {
         $params = $this->getRouteParams($request);
-        $filters = getFilters($request);
+        $filters = $this->getRouteParams($request);
         $filters['created_after'] = $request->input('created_after');
         $filters['created_before'] = $request->input('created_before');
         $this->usecase = $this->usecaseFactory
--- a/app/Http/Controllers/API/Forms/StatsController.php
+++ b/app/Http/Controllers/API/Forms/StatsController.php
@@ -24,6 +24,7 @@ class StatsController extends RESTController
     public function index(Request $request)
     {
         $params = $this->getRouteParams($request);
+        $filters = getFilters($request);
         $filters['created_after'] = $request->input('created_after');
         $filters['created_before'] = $request->input('created_before');
         $this->usecase = $this->usecaseFactory
--- a/app/Http/Controllers/API/Forms/StatsController.php
+++ b/app/Http/Controllers/API/Forms/StatsController.php
@@ -24,12 +24,13 @@ class StatsController extends RESTController
     public function index(Request $request)
     {
         $params = $this->getRouteParams($request);
-        $params['created_after'] = $request->input('created_after');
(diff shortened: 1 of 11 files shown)


src/App/Repository/Form/StatsRepository.php
```diff
@@ -260,12 +281,62 @@ public function getRecipients($form_id)
         ];
         $query = $this->selectQuery($where)
             ->resetSelect()
-            ->select([DB::expr('COUNT(distinct contact_id)'), 'total']);
+            ->select([DB::expr('COUNT(contacts.id)'), 'total']);
+
         $query = $this->targetedSurveyStateJoin($query);
+        if ($created_after || $created_before) {
```
rowasc: I am thinking that we may need this $query to be outside the `  if ($created_after || $created_before) {`  since you still want to do the join with messages and filter only outgoing even without a created_after/before filter .

src/App/Repository/Form/StatsRepository.php
```diff
@@ -260,12 +281,62 @@ public function getRecipients($form_id)
...
+            ->where('created', '>=', $created_after);
+        }
+
+        if ($created_before) {
+            $query = DB::select()
+            ->from([$query, 'selected_posts'])
+            ->where('created', '<=', $created_before);
+        }
+
+        $query = DB::select('contact_id')
```
rowasc: this overrides the $query instances before it . Is it intended to add to it and then execute, instead?

src/App/Repository/Form/StatsRepository.php
```diff
@@ -279,4 +350,99 @@ private function targetedSurveyStateJoin($query)
...
+            'web' => $this->queryForWeb($form_id, $created_after, $created_before),
+            'all' => $this->queryForAllPosts($form_id, $created_after, $created_before)
+        ];
+
+        return $result;
+    }
+
+    private function queryByDataSource($form_id, $created_after, $created_before, $source)
+    {
+        $query = DB::select([DB::expr('COUNT(messages.id)'), 'total'])
```
rowasc: I just realized that we could do a GROUP_BY('messages.type') and return all the datasource counts in 1 query /faster. LMK if you want to do that, we can pair a bit and do it. Of course this only applies to the queryByDatasource calls

src/Core/Usecase/Form/SearchFormStats.php
```diff
@@ -37,20 +36,60 @@ public function interact()
         // ... and get the search filters for this entity
         $search = $this->getSearch();
         $search->setFilter('form_id', $this->getIdentifier('form_id'));
+        $search->setFilter('created_after', $this->getIdentifier('created_after'));
```
rowasc: note that this would change if you change the code so that created_after is a filter instead of an identifier

src/App/Repository/Form/StatsRepository.php
crcommons: I think that because (when we are not filtering by date) we just want the total number of recipients, we're just looking up contacts for that targeted Survey in the query in `targetedSurveyStateJoin`. But if we want to filter by date, we need to join the messages table so we can only get the contacts for messages sent in that date range.

src/App/Repository/Form/StatsRepository.php
crcommons: I removed the unnecessary ones in the conditionals that check for created_after/before, but we still need the final one. it won't overwrite it because you're passing it in as the first argument in the `from`.

src/App/Repository/Form/StatsRepository.php
rowasc: I think you're correct. My impression was that maybe that was a bug caused by me when I coded that originally but now that I reviewed the other joins I think its actually correct. Thanks !

app/Http/Controllers/API/Forms/StatsController.php
```diff
@@ -24,10 +24,13 @@ protected function getResource()
     public function index(Request $request)
     {
         $params = $this->getRouteParams($request);
+        $filters['created_after'] = $request->input('created_after');
```
rowasc: $filters should be defined before

Issue #3193, opened 2018-08-08
Unable to recover forgotten password


tuxpiper:
In the develop branch (lumen), I'm unable to recover a user's forgotten password. The e-mail is apparently never sent. This information is left in the api logs:

```
[2018-08-08 15:02:34] lumen.ERROR: Symfony\Component\Debug\Exception\FatalThrowableError: Class 'Ushahidi\App\Repository\Security' not found in /var/www/platform/src/App/Repository/UserRepository.php:193
Stack trace:
#0 /var/www/platform/src/Core/Usecase/User/GetResetToken.php(85): Ushahidi\App\Repository\UserRepository->getResetToken(Object(Ushahidi\Core\Entity\User))
#1 /var/www/platform/app/Http/Controllers/RESTController.php(216): Ushahidi\Core\Usecase\User\GetResetToken->interact()
#2 /var/www/platform/app/Http/Controllers/API/PasswordResetController.php(31): Ushahidi\App\Http\Controllers\RESTController->executeUsecase(Object(Illuminate\Http\Request))
#3 [internal function]: Ushahidi\App\Http\Controllers\API\PasswordResetController->store(Object(Illuminate\Http\Request))
#4 /var/www/platform/vendor/illuminate/container/BoundMethod.php(29): call_user_func_array(Array, Array)
```

kinstelli:
This should be fixed with #3032

rowasc:
Eve to retest in passage environment

Eve-wanderi:
Failed QA i cannot reset my password i get the below error.

![image](https://user-images.githubusercontent.com/36038029/44257526-f055ca80-a214-11e8-8b02-d2176bf57c99.png)

rowasc:
@kinstelli can you investigate?

kinstelli:
For me:
1. Email was never sent from the passage server, but message is logged on the server.
2. Reset token is invalid.
<img width="478" alt="screen shot 2018-08-21 at 2 26 32 pm" src="https://user-images.githubusercontent.com/6600708/44424283-4052cb00-a54e-11e8-9f04-ecf088323fc7.png">

kinstelli:
It appears that when the user clicks the reset password link, this value is not urldecoded before being compared to the value stored in the database. 

However, when the user pastes the reset token into the Reset Password dialog, the reset process works, since that version is not urlencoded in the first place.

[...]

PR #3224, opened 2018-08-22
fix: adding features to support mailgun configurations

kinstelli:
- adds config/services.php
- adds MAILGUN_SECRET, MAILGUN_DOMAIN to env.deploy

Partial fix for ushahidi/platform#3193


config/services.php
```diff
@@ -0,0 +1,38 @@
...
+        'secret' => env('SES_SECRET'),
+        'region' => env('SES_REGION', 'us-east-1'),
+    ],
+
+    'sparkpost' => [
+        'secret' => env('SPARKPOST_SECRET'),
+    ],
+
+    'stripe' => [
+        'model' => App\User::class,
```
rjmackay: Does this cause any issues? I don't know if we have this class

.env.deploy
```diff
@@ -18,13 +18,16 @@ MULTISITE_CLIENT_DOMAIN=""
 REDIS_HOST=""
 REDIS_PORT=""
 
+MAILGUN_DOMAIN=""
```
rjmackay: These should just be `MAILGUN_DOMAIN=`
ie. don't provide a default value, because they *must* be configured

PR #3225, opened 2018-08-23
CSV ONLY speedups

rowasc:
- makes import faster and large imports feasible

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


values.json
```diff
@@ -0,0 +1 @@
+{"form": 9}
```
rowasc: need to remove this

PR #3235, opened 2018-08-23
Fixing session length

willdoran:
- Extends session time out to be 1 day

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/application/config/init.php
+++ b/application/config/init.php
@@ -28,7 +28,7 @@ return array(
 	'base_url'    => '/',
 	'index_file'  => FALSE,
 	'charset'     => 'utf-8',
-	'errors'      => FALSE,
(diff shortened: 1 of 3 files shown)


application/config/init.php
```diff
@@ -28,7 +28,7 @@
 	'base_url'    => '/',
 	'index_file'  => FALSE,
 	'charset'     => 'utf-8',
-	'errors'      => FALSE,
+	'errors'      => true,
```
rowasc: should be false

willdoran: no no , i think it should be true I accidentally reversed it earlier so this is tot fix it

PR #3241, opened 2018-08-27
Adding validator to check for duplicate description/title-fields

Angamanga:
- Refactoring form-attribute-repository to add a function to get attributes by type
- Adding a validator that checks for duplicate description and title-fields

##Test-script
Smoke-test to make sure saving and duplicating surveys work:
1. Create a new survey
2. Save
- [ ] Should save without error
3. Duplicate a survey
4. Save
- [ ] Should save without error

Test to make sure validator works (this need manipulation of the database)
1. Add another field of the type "description" to a survey in the database
2. duplicate this survey
3. Save
- [ ] You should get an error-message that say "There cannot be more than one field with the type description"
4. Repeat but add another field of the type "title" to a survey in the databse
5. duplicate this survey
6. Save
- [ ] You should get an error-message that say "There cannot be more than one field with the type title"

Do not forget to remove the added duplicate fields from the database!

Fixes ushahidi/account-management#42 .

Ping @ushahidi/platform


Angamanga:
Moving to QA

PR #3246, opened 2018-08-29
Bring imports from master 2

rowasc:
-

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


mapping.json
```diff
@@ -0,0 +1,9 @@
+{
```
rjmackay: @willdoran @rowasc is this actually used?

willdoran: No this is redundant.

PR #3258, opened 2018-08-30
Develop to production - solving issue with exports and datasources not workking

rowasc:
- Merges fixes from develop to production
- Fix saving message dates. They're objects now

Ping @ushahidi/platform

### DIFF
--- a/src/App/Repository/MessageRepository.php
+++ b/src/App/Repository/MessageRepository.php
@@ -185,9 +185,28 @@ class MessageRepository extends OhanzeeRepository implements
     // CreateRepository
     public function create(Entity $entity)
     {
-        return parent::create($entity->setState([
-            'created'   => time(),
-        ]));
+
+        $message = $entity->asArray();
+        // Set default value for post_date
+        if (!empty($message['datetime'])) {
+            $message['datetime'] = $message['datetime']->format("Y-m-d H:i:s");
+        }
+        $message['created'] = time();
+        // Create the post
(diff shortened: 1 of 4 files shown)


src/App/Repository/MessageRepository.php
```diff
@@ -185,9 +185,28 @@ public function getTotalMessagesFromContact($contact_id)
     // CreateRepository
     public function create(Entity $entity)
     {
-        return parent::create($entity->setState([
-            'created'   => time(),
-        ]));
+
+        $message = $entity->asArray();
+        // Set default value for post_date
+        if (!empty($message['datetime'])) {
```
rjmackay: is this *always* a date object or do we need to check that too?

willdoran: I think now it is always a date object, yes. I think the problem was that the code got ported for other things like post_date but this never got completed. It probably got lost in the shuffle of everything. If it changes from a date though I think it breaks a lot of things and the database will physically reject it. We can check but it might be redundant.

willdoran: I think checking would fall under better validation with better validation flow so that this stuff can be fixed or mitigated. The field can be null though according to the schema so it could be ignored if it's not a datetime since it's internally created.

willdoran: Also, I forgot it is checked in validation so it shouldn't get here if it's not a datetime object. I feel that is enough protection generally. BUT there is no guarantee an object goes through validation...see csv work around.

Issue #3254, opened 2018-08-30
Unable to Post Responses if user is logged in


Eve-wanderi:
**Describe the bug**
If a user is logged IN i cannot post any responses i get an error "There was a problem posting your response".

**Where was the bug observed**
Mobile App (https://rominacsvsms.ushahidi.io)

**To Reproduce**
1. Go to 'https://rominacsvsms.ushahidi.io' mobile App.
2. Log in as 'Admin' (Please note if certain user permissions, like Admin or Manage Data, are required to reproduce this bug. Also include if there is a support@ or admin@ account for this deployment. If not, contact the platform slack channel with further instructions.)
3. Click on ' Submit Survey Response'
4. Scroll down to 'Basic Post.'
5. Fill in all details

**Expected behavior**
User should get a notification "Post submitted successfully"

**Is there a workaround? What is it.**
For android phones No.

**Screenshots**

![screenshot_20180830-182746_ushahidi](https://user-images.githubusercontent.com/36038029/44864427-4a7c7400-ac88-11e8-8401-27ff62aa0037.jpg)
![screenshot_20180830-185738_ushahidi](https://user-images.githubusercontent.com/36038029/44864428-4b150a80-ac88-11e8-8419-e0985bc8eca7.jpg)

**URL / Environment where this happened**
[https://rominacsvsms.ushahidi.io]

**If the bug is datasource related (e.g. SMS or Targeted SMS), please specify the configuration details (e.g. Frontline + Clickatell, etc):**

**Desktop Hardware Details (please complete the following information):**

**Smartphone Hardware Details (please complete the following information):**
 - Device: [Samsung S8+]
 - OS: [Android]
 - Version [Version 8.0.0]

![screenshot_20180830-194457_settings](https://user-images.githubusercontent.com/36038029/44866399-cb8a3a00-ac8d-11e8-91de-c1402756787c.jpg)

**Additional context**
This functionality however works in IOS and another Android version i.e 4 .

See video below
[20180830_191729_gif.zip](https://github.com/ushahidi/platform/files/2337037/20180830_191729_gif.zip)

dalezak:
@rowasc looks like the error is `User 0 is not allowed to create resource posts #0`.

```
{"errors":[{"status":403,"message":"User 0 is not allowed to create resource posts #0"}]}
```

Any idea why the API would return that?

dalezak:
*POST*
```
https://rominacsvsms.api.ushahidi.io/api/v3/posts
```
[...]

PR #3259, opened 2018-08-31
feat: Re-add expires value to oauth responses for backward compatibility

rjmackay:
Refs: ushahidi/platform#3254

Test checklist:
- [ ] Log in or manually generate a token
- [ ] Check it returns an `expires` value 


rjmackay:
Turns out this wasn't needed for that bug. I'm going to close it

PR #3271, opened 2018-09-03
Remove unused queryForAllPosts method

rowasc:
- Remove unused methohd

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


crcommons:
Why are we removing this? @rowasc

rowasc:
@crcommons we don't use it. The total posts is the sum of all the posts categories, otherwise you are counting translations as different posts and doing extra queries

Issue #3272, opened 2018-09-03
Responding to a post from SMS or other datasource does not work


rowasc:
**Describe the bug**
The send message form does not appear for posts coming from twitter/email datasources (this probably happens for SMS too)

**Where was the bug observed**
rominacsvsms.ushahidi.io
Dreams partner here https://ushahidi.slack.com/archives/C03UU507R/p1535972733000100

**To Reproduce**
Data view - Sending messages
-    Go to Data View
-    Select a post from SMS or email source
-    Open the Send New Message screen for the post

**Expected behavior**
-  The send message form does not appear.

**Is there a workaround? What is it.**
No

**If the bug is datasource related (e.g. SMS or Targeted SMS), please specify the configuration details (e.g. Frontline + Clickatell, etc):**
In the one I tried in rominacsvsms.ushahidi.io we had twitter and email enabled. This also happened to a DREAMS partner (https://ushahidi.slack.com/archives/C03UU507R/p1535972733000100)

**Desktop Hardware Details (please complete the following information):**
 - OS: OSX
 - Browser Chrome
 - Version 68

rowasc:
Same as https://github.com/ushahidi/platform/issues/3277

rowasc:
Closing as there's another issue tracking this, and I see the send new message panel

PR #3273, opened 2018-09-03
When a message is saved, set the id in the entity before we send it to the CreatePostFromMessage listener

rowasc:
- Ensures that when we send a message to createPostFromMessages, it gets there with its ID .
- fixes https://github.com/ushahidi/platform/issues/3272
Test checklist:
- [ ] Configure a datasource and fetch (`php artisan datasource:incoming`) from it
- [ ] With platform-client open, check that the tweets/emails have a contact and that you can see the "Conversation with author" panel in the post details

- [x] I certify that I ran my checklist ( tested with twitter)

## Important: 
- The entity id was not getting to where it should get in createPostFromMessage. It works, and I think it should not impact targeted surveys at all because of the `        if ($this->targetedSurveyStateRepo->isContactInActiveTargetedSurveyAndReceivedMessage($message->contact_id)) {` in  CreatePostFromMessage::handle method... but it's a pretty naive fix and it may impact other things  in some way that I didn't realize while checking it , so any advice is appreciated. 
-  is there a better way to do this? I wasn't sure.

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/src/Core/Usecase/Message/ReceiveMessage.php
+++ b/src/Core/Usecase/Message/ReceiveMessage.php
@@ -91,6 +91,8 @@ class ReceiveMessage extends CreateUsecase
         // ... persist the new message entity
         $id = $this->repo->create($entity);
 
+        $entity->setState(compact('id'));
+
         $this->dispatch('message.receive', [
             'id' => $id,
             'entity' => $entity,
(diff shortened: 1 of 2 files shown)


rowasc:
Probably a @willdoran review , I can test this more and check that targeted surveys are not impacted when I wake up tomorrow, but sent the PR in case @rjmackay wants to follow up since its impacting a dreams partner.

Issue #3269, opened 2018-08-31
Notifications for posts added to a collection are not being sent


rowasc:
**Describe the bug**
I didn't receive notifications when a post was added to a collection I was subscribed to

**Where was the bug observed**
rominacsvsms.ushahidi.io

**To Reproduce**
1. Go to the collections list. 
2. Click a collection. It will take you to the map view . 
3. Click the "3 dots"  button for the collection in the mode bar. 
4. Click the "notifications" button to turn them on.
5. Add a post to the collection

**Expected behavior**
5. You should receive a notification

**Is there a workaround? What is it.**
no

**Screenshots**
<img width="212" alt="screen shot 2018-08-31 at 5 24 47 pm" src="https://user-images.githubusercontent.com/2434401/44934401-0ac69280-ad43-11e8-9daf-e88503e62c22.png">

rowasc:
Note: email datasource was configured for this deployment .

rowasc:
@jrtricafort

jrtricafort:
@rowasc This is a Lumen regression or existing bug?

rowasc:
Regression, it was passing

jrtricafort:
p1

rjmackay:
@rowasc What user and deployment was this tested with?
If the user doesn't have an contacts with the `can_notify` flag set, then their notifications are ignored. It's a weird behaviour, but not new.

rjmackay:
Ah. It looks suspiciously like messages are queued as `incoming` not `outgoing`. It looks suspiciously like this might been the case in the old code too, but it might have been getting fixed somewhere else in the data flow.

rowasc:
>If the user doesn't have an contacts with the can_notify flag set, then their notifications are ignored. It's a weird behaviour, but not new.
I don't know if my user had that flag. It's `romina+csvdeployment@ushahidi.com`

PR #3288, opened 2018-09-11
Fix notifications for posts added to collections

rjmackay:
- Adds more debug info when queuing notifications
- Queue messages as `outgoing`

Test checklist:
- [ ] Go to the collections list.
- [ ] Click a collection. It will take you to the map view .
- [ ] Click the "3 dots" button for the collection in the mode bar.
- [ ] Click the "notifications" button to turn them on.
- [ ] Add a post to the collection
- [ ] Confirm message is queued for sending
- [ ] Confirm queued message has `outgoing` direction

Fixes ushahidi/platform#3269

Ping @ushahidi/platform


rjmackay:
@rowasc it turns out that the email datasource isn't enabled by default anymore. We had a hack that did that previously but it enabled both the incoming and outgoing email, just so we could send outgoing mail.

I've tried to come up with a less hacky approach by splitting the email data source. Then to avoid dupe datasources in the config UI, I made it possible to mark it as no configurable and hide it from the API. It's not great. I'm curious if you have better ideas? I'd love this to be less ugly

src/App/DataSource/DataSource.php
```diff
@@ -26,4 +26,5 @@ public function getOptions();
     public function getInboundFields();
     public function getInboundFormId();
     public function getInboundFieldMappings();
+    public function isUserConfigurable();
```
rjmackay: Added this to define which sources can be configured, and thus which appear in the API

src/App/DataSource/Email/Email.php
```diff
@@ -21,7 +21,7 @@
 use Ushahidi\Core\Entity\Contact;
 use Log;
 
-class Email implements IncomingAPIDataSource, OutgoingAPIDataSource
+class Email extends OutgoingEmail implements IncomingAPIDataSource
```
rjmackay: Still extends outgoing email, partly because I wasn't sure what exactly might break if we had to send and receive through different sources

src/App/Repository/Config/data-provider.php
```diff
@@ -14,6 +14,7 @@
         // List of data providers key=provider value=enabled
         'smssync' => false,
         'email' => false,
+        'outgoingemail' => true,
```
rjmackay: Defaults to enabled

src/App/Repository/DataproviderRepository.php
```diff
@@ -65,22 +65,12 @@ public function getEntity(array $data = null)
...
-            // Returns all *enabled* providers.
-            return $this->datasources->getEnabledSources();
-        } else {
-            // Returns all providers, even if they are disabled.
-            return $this->datasources->getSource();
-        }
-    }
-
-    // DataProviderRepository
-    public function all($enabled = false)
```
rjmackay: Dropped this because nothing used it

src/App/Repository/DataproviderRepository.php
```diff
@@ -65,22 +65,12 @@ public function getEntity(array $data = null)
      * Get all enabled providers, with their configuration data.
      * @return Array
      */
-    protected function getAllProviders($enabled = false)
+    protected function getAllProviders()
     {
-        if ($enabled) {
```
rjmackay: Dropped the enabled filter because again, nothing used it

src/App/Repository/DataproviderRepository.php
```diff
@@ -65,22 +65,12 @@ public function getEntity(array $data = null)
...
-    }
-
-    // DataProviderRepository
-    public function all($enabled = false)
-    {
-        $providers = $this->getAllProviders($enabled);
-        return $this->getCollection($providers);
+        // Returns all providers, even if they are disabled.
+        return array_filter($this->datasources->getSource(), function ($source) {
+            return $source->isUserConfigurable();
```
rjmackay: filter out anything that can't actually be configured

rjmackay:
🎉 successfully got notification emails on sad-charlie (this branch is deployed there)

src/Console/Command/Notification.php
```diff
@@ -130,16 +137,19 @@ private function generateMessages($notification)
                     'message' => trans('notifications.' . $messageType . '.message', $subs),
                     'type' => $messageType,
                     'data_source' => $data_source,
```
rjmackay: Need to make this $data_source->getId() otherwise we're trying to save an object and failing

rjmackay:
@rowasc I've updated with a fix for notifications - there was a bug that only happened when email was configured - and also a fix for incoming email.

src/App/DataSource/Email/Email.php
```diff
@@ -219,11 +179,25 @@ public function fetch($limit = false)
...
-            $last_uid = $this->messageRepo->getLastUID('email');
-            $max_range = $last_uid + $limit;
-            $search_string = $last_uid ? $last_uid + 1 . ':' . $max_range : '1:' . $max_range;
+            $mailboxinfo = imap_check($connection);
+
+            Log::info("Connected to $inbox", [$username, $password, $mailboxinfo]);
 
-            $emails = imap_fetch_overview($connection, $search_string, FT_UID);
+            $last_uid = $this->messageRepo->getLastUID('email');
+            if ($last_uid > 0) {
```
rjmackay: This might still fail if you change you email settings, because it won't reset the last uid. We could update it to save that to config and reset similar to the twitter provider.

rowasc: that'd be cool. This could happen with admins switching the admin email most likely, I think.

PR #3293, opened 2018-09-14
Recursive diff should do a strict comparison to account for bool/truthy vals like 1 not going as true

rowasc:
- Fixes issue with some config features not being set because the  value would compare as true and return that nothing changed when changing ints/strings that had a default value of true in the site config. 

Test checklist:
- [ ] Test https://github.com/ushahidi/platform-cloud-interface/issues/389
- [ ] Need to add some kind of test suite to verify this didn't break half the app updates/creates of entities

Fixes ushahidi/platform# .

Ping @ushahidi/platform


rowasc:
Hold merging until we decide when this goes live

rowasc:
I think this could get merged now that pricing is live, I could test it next week in sad-charlie and deploy it (doing a lot of other higher priority things this week )

rowasc:
This is still very relevant. I think it should be tested and merged after the next release is live.

tuxpiper:
> This is still very relevant as a general fix. I think it should be tested and merged after the next release is live.

Shall we label it to signify it's planned to land on the release after next?

rowasc:
good idea yea

rowasc:

PR #3316, opened 2018-10-06
fix: typo in service initialization while setting params

ozdemirburak:
- Fixed the typo in service  initialization while setting params

Greetings, 

I've a question also, have you guys discussed or considered about upgrading `aura/di` to `~3.0`?
I actually tried to do that, and was actually going to PR that if I'd succeeded but I got stuck, since the new version locks after calling `get()` method only once.


ozdemirburak:
[clabot:check]

rowasc:
@ozdemirburak regarding the aura question, if you open an issue we could follow up there  (please ping me in it). Thanks for your contribution, and happy hacktoberfest :)

PR #3348, opened 2018-10-22
Twitter refactor data

rowasc:

Attempts to solve Twitter compliance issues

- Tweet content and related data that we had from before we fixed this issue should be removed
    - For posts where the `posts.content` field matched a `messages.message` that was created from a tweet, replaced the message with the tweet's URL. 
    - For posts where the `posts.title` field matched a `messages.message` that was created from a tweet, replaced the title with 'From Twitter on {date}'` 
    - For posts that had location data extracted from Twitter, removed the location data
    - Removing tweet's contents from the messages database after the previous steps are done. 
- Twitter Usernames should not be saved
    - Removing contact usernames that we stored from twitter
    - For future tweets, saving the tweet user ID, instead of tweet username 

- New tweets that we extract with `php artisan datasource:incoming`
    - Do not save the tweet's content. Instead, use  'From Twitter on {date}'` as the title of the post created and the Tweet URL as the content. 
    - Do not save location data from Twitter

- Exports: 
    - Add the data_source_message_id field to the CSV exports to be able to extract the tweet id
    - Do not export content for unstructured posts from Twitter. 

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


migrations/20181023140057_remove_contact_usernames.php
```diff
@@ -0,0 +1,35 @@
...
+     *    addIndex
+     *    addForeignKey
+     *
+     * Remember to call "create()" or "update()" and NOT "save()" when working
+     * with the Table class.
+     */
+    public function change()
+    {
+        $this->table('contacts')
+            ->changeColumn('contact', 'string', ['null' => true])
```
rowasc: ugh no this should just be the opposite, null out the contact in message , and keep contact null constraint .

PR #3349, opened 2018-10-22
Run CSV exports in laravel jobs [WIP]

rjmackay:
- Create job queues for handling CSV export:
    - [x] initialize
    - [x] batch
    - [x] combine
- Remove old export/external routes
- Add entity and table to track each export batch
- Dispatch event for *all* create usecase runs
- Create a base repo for using eloquent instead of kohana DB 🎉 
- ~[ ] Combine export init actions into single usecase (count, generate headers, etc)~

Test checklist:
- [x] composer test

Fixes ushahidi/account-management#48

Ping @ushahidi/platform


app/Jobs/ExportPostsBatchJob.php
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * The job failed to process.
+     *
+     * @param  Exception  $exception
+     * @return void
+     */
+    public function failed(Exception $exception)
+    {
+        $exportJobRepo = app(ExportJobRepository::class);
```
rjmackay: Need to test this. Not sure how exactly

app/Jobs/CombineExportedPostBatchesJob.php
```diff
@@ -0,0 +1,147 @@
...
+            Log::debug('Job already completed', ['jobId' => $this->jobId]);
+            // All done here
+            return;
+        }
+
+        // Check if all batches exported
+        if (!$exportJobRepo->isJobFinished($this->jobId)) {
+            Log::debug('Batches not finished, requeueing', ['jobId' => $this->jobId]);
+            // Batches aren't done yet, wait 5 minutes and try again
+            dispatch(
```
rjmackay: I'm not sure how we'd ever get here, but just in case

app/Jobs/CombineExportedPostBatchesJob.php
```diff
@@ -0,0 +1,147 @@
...
+
+        // Combine the files
+        $destinationFile = $this->combineFiles($fileNames);
+
+        // Clean up partial files
+        Storage::delete($fileNames);
+
+        // Set status = completed
+        $job->setState([
+            'url' => Storage::url($destinationFile),
```
rjmackay: I'm not sure this should actually be a URL, or a filename that gets formatted to a URL in the Formatter.

app/Jobs/CombineExportedPostBatchesJob.php
```diff
@@ -0,0 +1,147 @@
...
+            'url_expiration' => null, // $urlExpiration, // @todo fix me
```
rjmackay: No idea how expirations worked. It looked like we were setting CDN expiries at all!?

migrations/20181018034914_add_export_job_batches.php
```diff
@@ -0,0 +1,30 @@
...
+    {
+        $this->table('export_batches') // @todo revisit name?
+            ->addColumn('export_job_id', 'integer', ['null' => false])
+            ->addColumn('status', 'string', [
+                'null' => false,
+                'default' => 'pending'
+            ])
+            ->addColumn('batch_number', 'integer', ['default' => 0])
+            ->addColumn('filename', 'string')
+            ->addColumn('url', 'string', ['null' => true]) // Do we need both?
```
rjmackay: I'm not using URL so I should drop this

src/App/Repository/EloquentRepository.php
```diff
@@ -0,0 +1,228 @@
...
+     * Converts an array/collection of results into an collection
+     * of entities, indexed by the entity id.
+     *
+     * Included directly instead of using Ushahidi\Core\Traits\CollectionLoader
+     * because this implementation returns an Illuminate\Support\Collection
+     *
+     * @param  Array|Iterable $results
+     * @return \Illuminate\Support\Collection
+     */
+    protected function getCollection($results)
```
rjmackay: Added this directly to this repo, rather than using the existing trait. Laravel Collections are much more useful than arrays.

src/App/Repository/EloquentRepository.php
```diff
@@ -0,0 +1,228 @@
...
+    /**
+     * Check if an entity with the given id exists
+     * @param  int $id
+     * @return bool
+     */
+    public function exists($id)
+    {
+        return $this->connection->table($this->getTable())
+            ->select($this->getTable() . '.*') // @todo do we need this?
+            ->where($where) // @todo do we need to handle whereIn here too?
```
rjmackay: Don't need to resolve this right now... but maybe I should so it doesn't trip someone else up?

app/Jobs/CombineExportedPostBatchesJob.php
rowasc: yea, we were. It's in the ExportBroker.php in the lumen_csvexport_queues repo 
$url_expiration = time() + getenv('TEMP_URL_LIFESPAN');

rjmackay: I finally found it in the old code it calls something like `fs->getTempUrl()` to get an expiring URL from RS. I need to figure out how to do that. And how to make sure the bucket isn't public to start with

kinstelli: I'm happy to review that with you. From their docs, I don't think there was a way to lock down that bucket otherwise. Any chance we're changing providers?

app/Jobs/ExportPostsJob.php
```diff
@@ -0,0 +1,92 @@
...
+
+use Exception;
+use Ushahidi\Core\Usecase\Export\Job\PostCount;
+use Ushahidi\Core\Entity\ExportJob;
+use Ushahidi\Core\Entity\ExportJobRepository;
+use Illuminate\Support\Facades\Log;
+
+class ExportPostsJob extends Job
+{
+    protected $batchSize = 200;
```
kinstelli: could this live in a config, but have 200 as a default?

app/Jobs/CombineExportedPostBatchesJob.php
kinstelli: And is this code exclusively for multisite?

src/App/Repository/ExportJobRepository.php
```diff
@@ -108,40 +107,29 @@ public function create(Entity $entity)
...
         return parent::create($entity->setState($state));
     }
 
-    // overriding the update method here to intercept 'EXPORT_COMPLETED'
-    //   and decide if we still need to send something to HDX or mark this
-    //   as SUCCESSful
+    // Overriding the update method here to handle state transitions
     public function update(Entity $entity)
     {
-        //check for new status of 'EXPORTED_TO_CDN'
```
rjmackay: This was terrifying so I moved it into the entity itself

app/Jobs/CombineExportedPostBatchesJob.php
kinstelli: is this code meant to double for multisite and OSS? Would Storage::url be more versatile?

src/Core/Entity/ExportJob.php
```diff
@@ -76,4 +89,32 @@ protected function getImmutable()
...
+        return in_array(
+            $this->status,
+            [self::STATUS_SUCCESS, self::STATUS_EXPORTED_TO_CDN, self::STATUS_PENDING_HDX]
+        );
+    }
+
+    /**
+     * Handle state transitions
+     */
+    public function handleStateTransition()
```
rjmackay: Moved here so that we're not hiding business logic in the repo

src/Core/Entity/ExportJob.php
```diff
@@ -76,4 +89,32 @@ protected function getImmutable()
...
+    {
+        // Check for new status of 'EXPORTED_TO_CDN'
+        if ($this->hasChanged('status') && $this->status == ExportJob::STATUS_EXPORTED_TO_CDN) {
+            if ($this->send_to_hdx) {
+                // Jump to next state PENDING_HDX
+                $this->setState(['status' => ExportJob::STATUS_PENDING_HDX]);
+                // if sending to HXL is required, then we spawn an event to do that
+                Event::fire(new SendToHDXEvent($this->getId()));
```
rjmackay: Makes the entity dependent on Lumen, but the app is cleaner overall

app/Jobs/ExportPostsJob.php
rjmackay: It could. I'll see if I have time since I think it was hard coded before

app/Jobs/CombineExportedPostBatchesJob.php
rjmackay: OSS only right now. But it'll do both in future.
So if we store a URL then you can't move files from 1 storage driver to another, because the URL is hard coded. If we store only the path, then we can run it through Storage::url() in the formatter.

I've looked into it and we can actually make the RS driver build the Temp url for us, using a different URL for every request.

app/Jobs/CombineExportedPostBatchesJob.php
rjmackay: OSS right now (Jobs won't work with multisite). 
I'm not changing providers right now... but it should work with either of them

app/Jobs/ExportPostsJob.php
```diff
@@ -0,0 +1,105 @@
...
+                $this->jobId,
+                $batchNumber,
+                $batchNumber * $this->batchSize,
+                $this->batchSize,
+                ($batchNumber === 0)
+            ));
+        }
+
+        // Set status = queued
+        $job->setState([
```
rjmackay: This has a slightly bizarre result if QUEUE_DRIVER=sync.
The jobs above run before this update. So the final job status = queued, instead of completed.

I can swap the order, but its just a bit weird. @rowasc @kinstelli thoughts?

src/App/Formatter/ExportJob.php
```diff
@@ -12,8 +12,42 @@
...
+        $url_path = explode("/", $value);
+        $filename = rawurlencode(array_pop($url_path));
+        array_push($url_path, $filename);
+        $path = implode("/", $url_path);
+
+        $expiration = Carbon::now()->add(CarbonInterval::fromString(config('media.temp_url_lifespan')));
+
+        // Try to get a temporary URL
+        try {
+            return Storage::temporaryUrl($path, $expiration);
```
rjmackay: @rowasc @kinstelli I think this does this job? Although it'll generate a new URL each time

src/App/Formatter/ExportJob.php
```diff
@@ -12,8 +12,42 @@
...
     use FormatterAuthorizerMetadata;
+
+    protected function formatUrl($value)
+    {
+        if (empty($value)) {
+            return $value;
+        }
+
+        // If we already have a URL, just return it
+        if (filter_var($value, FILTER_VALIDATE_URL)) {
```
rjmackay: This handles the existing entries with full URLs in the DB

rjmackay:
@rowasc I changed the base for this to `spinoff-yerevan-daisy-with-develop`... but I think that needs the latest develop merged into it too?

PR #3357, opened 2018-10-25
Adding limit check for demo

willdoran:
- Should only see first 25 posts when set to demo_1 tier

Test checklist:
- [x] For deployment with 25+ posts you should only ever see the first 25
- [x] You should not be able to load more
- [x] You should be able to see that you are viewing 25 of real number of posts
- [x] You should not be able to get more posts by using the offset

Fixes ushahidi/platform# .

Ping @ushahidi/platform


rowasc:
Export is returning all posts, but I think that was by design , right? The checklist passed.

PR #3370, opened 2018-10-30
Upgrade aura/di

rjmackay:
- Upgrade aura/di to v3.4
- Fix DI configuration as needed
- Remove unused and broken feature() helper
- Move all the DI configuration into Config classes

Test checklist:
- [x] composer test

Ping @ushahidi/platform


app/Providers/AppServiceProvider.php
```diff
@@ -26,8 +26,6 @@ public function register()
         $this->registerFilesystem();
         $this->registerMailer();
 
-        $this->configureAuraDI();
```
rjmackay: All of this moved to LumenAuraConfig.php

src/Init.php
```diff
@@ -31,436 +31,16 @@ function service($what = null)
 {
     static $di;
     if (!$di) {
-        $di = new Aura\Di\Container(new Aura\Di\Factory);
+        $builder = new Aura\Di\ContainerBuilder();
+        $di = $builder->newConfiguredInstance([
+            'Ushahidi\Core\CoreConfig',
```
rjmackay: All the aura config now happens in these 4 Config classes

src/Console/ConsoleConfig.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Ushahidi\Console;
+
+use Aura\Di\Container;
+use Aura\Di\ContainerConfig;
+
+class ConsoleConfig extends ContainerConfig
```
rjmackay: This probably could just merge to something else but one thing at a time

src/Console/ConsoleConfig.php
```diff
@@ -0,0 +1,18 @@
...
+{
+    public function define(Container $di)
+    {
+        $di->set('authorizer.console', $di->lazyNew('Ushahidi\Console\Authorizer\ConsoleAuthorizer'));
+    }
+
+    public function modify(Container $di)
+    {
+    }
+}
```
rjmackay: These are very similar to ServiceProviders for Laravel. 
define ~= register
modify ~= boot

app/Providers/LumenAuraConfig.php
```diff
@@ -0,0 +1,137 @@
...
+
+class LumenAuraConfig extends ContainerConfig
+{
+    public function define(Container $di)
+    {
+        $this->configureAuraServices($di);
+        $this->injectAuraConfig($di);
+    }
+
+    public function modify(Container $di)
```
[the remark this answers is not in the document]
rjmackay: IDN. From the docs

> Modify service objects after the Container is locked.

However... we can actually remove it because it's defined in the parent class. I thought I had to keep it for the interface.

PR #3372, opened 2018-11-02
refactor: Handle all feature toggles and limits through Features service

rjmackay:
- Refactor feature service to load feature toggles from config repo
- refactor: Update PrivateDeployment trait to use Features::isEnabled and grab config directly
- fix: data provider work without service features
- refactor: Remove roles.enabled from aura/di and use Features::isEnabled
- test: Fix RoleUpdateTest to mock features
- refactor: Remove csv-speedup.enabled from aura/di, use Features::isEnabled
- refactor: Remove webhook.enabled from aura/di, and use Features::isEnabled
- refactor: Remove post-locking feature toggle as its unused
- refactor: Remove redis.enabled from aura/di and use Features::isEnabled
- refactor: Use Features::isEnabled to check for data-import feature
- refactor: Move feature limits to Features::getLimit()

This is to make life easier when refactoring multisite

Test checklist:
- [x] composer test

Ping @ushahidi/platform


app/Tools/Features.php
```diff
@@ -20,16 +21,40 @@ public function __construct(array $config)
...
+            if ($config->limits[$feature] === true) {
+                // Return infinity ie. unlimited
+                return INF;
+            }
+
+            return (int) $config->limits[$feature];
+        }
+
+        // Return infinity ie. unlimited
+        return INF;
```
kinstelli: More of a question: I couldn't find another example where we return INF in the codebase, and though it makes total sense logically, I seem to remember that it can get treated differently with different interpreters. Of course, I can't find an example now, aside from HHVM, so maybe I'm writing this comment for no useful reason.

src/App/Validator/Role/Update.php
```diff
@@ -44,7 +44,7 @@ protected function getRules()
 
     public function checkRolesEnabled(\Kohana\Validation\Validation $validation)
     {
-        if (!$this->feature_enabled) {
+        if (!Features::isEnabled('roles')) {
```
kinstelli: Yep, thank you. I like this better.

app/Tools/Features.php
rjmackay: I didn't even know it existed until I went looking for a better option than returning `true` and then have to test for it specially

PR #3375, opened 2018-11-06
refactor: Make data sources work with multisite refactor

rjmackay:
- Defer creating data source instances till they're needed
- Move creating datasources into manager
- Change interfaces on manager to often just return source names, not classes
- Move registering routes to static methods
- Handle config in data source manager instead of provider
- Reset data sources when site changes

Test checklist:
- [ ] composer test

Ping @ushahidi/platform

### DIFF
--- a/src/App/DataSource/DataSourceManager.php
+++ b/src/App/DataSource/DataSourceManager.php
@@ -2,38 +2,42 @@
 
 namespace Ushahidi\App\DataSource;
 
+use Laravel\Lumen\Routing\Router;
+use Ushahidi\Core\Entity\ConfigRepository;
+use InvalidArgumentException;
+
 class DataSourceManager
 {
+    /**
+     * Config repo instance
+     *
+     * @var \Ushahidi\Core\Entity\ConfigRepository;
+     */
+    protected $configRepo;
 
     /**
-     * Router instance
+     * Sources to be configured
      *
-     * @var \Laravel\Lumen\Routing\Router
+     * [name => classname]
+     *
+     * @var [string, ...]
      */
-    protected $router;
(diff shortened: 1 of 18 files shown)


src/App/DataSource/DataSourceManager.php
```diff
@@ -47,55 +51,89 @@ class DataSourceManager
...
+     *
+     * @return [string, ...]
+     */
+    public function getEnabledSources() : array
     {
-        if ($name) {
-            return isset($this->sources[$name]) ? $this->sources[$name] : false;
-        }
+        // Load enabled sources
+        $enabledSources = array_filter(
```
rjmackay: This is a bit gross since it makes DataSourceManager reliant on the config repo *but* it seems better than hiding that detail in a service provider instead

src/App/DataSource/DataSourceManager.php
```diff
@@ -104,25 +149,130 @@ public function getSourceForType($type)
...
+    protected function createEmailSource(array $config)
+    {
+        return new Email\Email(
+            $config,
+            app('mailer'),
+            app(\Ushahidi\Core\Entity\MessageRepository::class)
+        );
+    }
+
+    protected function createOutgoingemailSource(array $config)
```
kinstelli: nitpick: if we're calling it OutgoingEmail elsewhere, can we adjust the camel case here?

rjmackay: Not easily because we generate the method name in `resolve` with `$driverMethod = 'create'.ucfirst($name).'Source';`. We'd have to rely on `$name` already being camelcase

PR #3371, opened 2018-11-01
refactor: Reorganize multisite handling

rjmackay:
This pull request makes the following changes:
- Replace the old Multisite class with a collection of classes
    - MultisiteManager which handles which site is currently enabled, and switching DB
    - SiteRepo for looking up sites
    - Site to represent the site itself and get urls/config/etc
- Replace DB injection with injecting a Resolver class which gives us the current DB as needed
- Add MultisiteAwareJob class to run jobs in a multisite context
- Updating *all* the repos to use new approach

Test checklist:
- [x] composer test
- [x] I certify that I ran my checklist

Depends on #3372

Fixes ushahidi/platform#2787

Ping @ushahidi/platform

PR #3407, opened 2018-11-20
refactor: Move CSV imports into queued jobs

rjmackay:
- Move csv/:id/import handler into CSVController
- Replace generic ImportUsecase with ImportCSVUsecase since it was already specific
- Move code from ImportListener (and a little from the controller) into ImportCSVPostsUsecase
- Trigger ImportPosts event and then ImportPostsJob from ImportCSVUsecase
- Add ImportPostsJob to run ImportCSVPostsUsecase

Cannot land till after #3371 

Test checklist:
- [ ] composer test

Ping @ushahidi/platform


app/Jobs/ImportPostsJob.php
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * Execute the job.
+     *
+     * @return void
+     */
+    public function handle(ImportCSVPostsUsecase $usecase)
+    {
+        $usecase->setIdentifiers(['id' => $this->csvId]);
+
+        $results = $usecase->interact();
```
rjmackay: Just fires off the usecase :)

rjmackay:
I've deployed this and #3408 to sad charlie. I have a few fixes incoming but it appears to work. *however* because we're not batching the imports we'll have to set the queue timeout *really long* before we deploy this

src/App/Listener/ImportPosts.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    protected $session;
+    public function __construct(Session $session)
+    {
+        $this->session = $session;
+    }
+
+    public function handle($id, Entity $entity)
+    {
+        $userId = $this->session->getUser()->getId();
```
rjmackay: Since we're sending this to another process we need to tell it which user to use

src/App/Listener/ImportPosts.php
```diff
@@ -0,0 +1,40 @@
...
+
+        // If csv-queue feature is enabled
+        if (Features::isEnabled('csv-queue')) {
```
rjmackay: @willdoran I completely forgot I added this. Basically this means without the feature flag this should be functionally the same as it is now: the CSV import runs synchronously in the HTTP request.

rowasc:
Keeping listed elsewhere until we decide if this can be merged yet

PR #3412, opened 2018-11-26
Hotfix: Avoid duplicate combine jobs

rjmackay:
- fix(ExportPosts): Queue combine job at start, rather than when batches finish

  Queue the CombineExportedPostBatchesJob job at the start, and let
it keep retrying. Rather than queuing when batches finish and ending
up with multiple combine jobs running at once

Test checklist:
- [x] composer test

Ping @ushahidi/platform

### DIFF
--- a/app/Jobs/ExportPostsJob.php
+++ b/app/Jobs/ExportPostsJob.php
@@ -7,6 +7,7 @@ use Ushahidi\Core\Entity\ExportJob;
 use Ushahidi\Core\Entity\ExportJobRepository;
 use Ushahidi\App\Multisite\MultisiteAwareJob;
 use Illuminate\Support\Facades\Log;
+use Illuminate\Support\Carbon;
 
 class ExportPostsJob extends Job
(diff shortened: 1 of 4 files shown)


app/Jobs/ExportPostsJob.php
```diff
@@ -88,5 +89,11 @@ public function handle(PostCount $usecase, ExportJobRepository $exportJobRepo)
                 ($batchNumber === 0)
             ));
         }
+
+        // Queue combine job in 1 min
+        dispatch(
+            (new CombineExportedPostBatchesJob($this->jobId))
```
rowasc: so it does the first try after 1 minute, but if not done, it'll just retry until it can do it, right?

rjmackay: It'll queue it to run in 1 minute. Then the CombineExportedPostBatchesJob checks if all the batches are done, and if they're not retries in 5 mins.

PR #3431, opened 2018-12-04
Migration/add import and mapping models

rjmackay:
- Add Import and ImportMapping models and repos
- These will be used to keep track of relations between imported data and the original IDs

Ping @ushahidi/platform

### DIFF
--- a/src/App/ImportUshahidiV2/ImportMapping.php
+++ b/src/App/ImportUshahidiV2/ImportMapping.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Ushahidi\App\ImportUshahidiV2;
+
+use Illuminate\Database\Eloquent\Model;
+
+class ImportMapping extends Model
+{
+
+    public $timestamps = false;
+
+    protected $attributes = [];
+
+    protected $casts = [
+        'import_id' => 'integer',
+        'source_type' => 'string',
+        'source_id' => 'integer',
+        'dest_type' => 'string',
+        'dest_id' => 'integer',
+    ];
+
+    protected $fillable = [
+        'import_id',
+        'source_type',
+        'source_id',
+        'dest_type',
+        'dest_id',
+    ];
+
+    public function import()
+    {
+        return $this->belongsTo(Import::class);
+    }
+}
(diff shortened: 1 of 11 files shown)


src/App/ImportUshahidiV2/Repositories/ImportRepository.php
```diff
@@ -0,0 +1,24 @@
...
+
+use Ushahidi\App\ImportUshahidiV2\Import;
+use Ushahidi\App\ImportUshahidiV2\Contracts\ImportRepository as ImportRepositoryContract;
+
+class ImportRepository /*extends EloquentRepository*/ implements ImportRepositoryContract
+{
+
+    public function create(Import $model) : int
+    {
+        return $model->save() ? $model->id : false;
```
rjmackay: This is slightly pointless. I could just use the model directly. For now it allows for swapping out Eloquent later though

src/App/ImportUshahidiV2/ImportMapping.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace Ushahidi\App\ImportUshahidiV2;
+
+use Illuminate\Database\Eloquent\Model;
+
+class ImportMapping extends Model
```
rjmackay: So I'm not sure about this... but I didn't want to add more repos/entities/etc if I didn't need to so I've just used Eloquent. On the one hand it's a chance to test using eloquent, but on the other it means we have multiple ways to do the same thing. What do you think?

rowasc: IDK , but we have to start using it at some point, so we might as well start with a small low-er impact thing like this  (compared to any other random place in platform, because it's a CLI )

src/App/ImportUshahidiV2/Repositories/ImportRepository.php
rowasc: _shrugs_
That's fine I think

src/App/ImportUshahidiV2/ImportMapping.php
rjmackay: Yea. If I have time I'll play with options a bit, ie. just using Eloquent vs using Eloquent and also implementing the `Entity` interface

Issue #3212, opened 2018-08-17
URL for OAuth end-points in API docs is wrong


batpad:
**Describe the bug**

The documentation says the URL for the OAuth end-point is `/api/v3/oauth/token`. However, that URL returns a 404. The correct URL seems to be `/oauth/token` (without the `/api/v3` prefix).

**To Reproduce**

1. Visit `/platform/api/v3/oauth/token` as per http://github.ushahidi.org/platform/docs/api/index.html#authorization-with-oauth-2

**Expected behavior**
For the URL to work as per the documentation. Instead, get a 404.

**Is there a workaround? What is it.**
Use the correct URL, which is `/oauth/token` without the `/api/v3` prefix

amalmurali47:
I can confirm this is reproducible for me as well.

rowasc:
Thanks, folks. Leaving for the day, but If @batpad or @amalmurali47 want to open a PR, I'll review when I come back online

rei2hu:
I can take a look at the endpoints in the api documentation if no one else is doing it.

rowasc:
@rei2hu that'd be great, thanks!. Also, sorry but what do you mean with `the demo page doesn't seem to be working?` which page is the one that is failing?

@tuxpiper you might be interested in following up here /collaborate with @rei2hu  since you're leading our doc improvement efforts

rei2hu:
I tried to access the demo page shown in the documentation (https://demo.api.ushahidi.io/) and it only responds with `Sorry, the page you are looking for could not be found.`.

I also tried to make a few requests like `curl -X POST https://demo.api.ushahidi.io/oauth/token` and I got the same thing so I'm not exactly sure how to go about checking the endpoints.

I am able to do `curl -X POST https://api.ushahidi.io/oauth/token` and get a proper error response though.

rowasc:
@rei2hu ah thanks for clarifying that.  I suspect that might have been an example URL all along, but pinging @tuxpiper in case he knows.  

To check out API responses without installing the platform locally, you can setup a free demo-tier ushahidi deployment (with our sass platform) https://www.ushahidi.com/  
Then you can use *mydeployment.api.ushahidi.io* (replace `mydeployment` with your deployment's name) to check out the api

--------

To get some context in this ticket : 
@tuxpiper is there a reason we removed the demo deployment? or was it always just an example URL? 
This is also a perfect example of how a playground idea like the one you suggested would be beneficial :D

tuxpiper:
[...]

PR #3433, opened 2018-12-05
Fix: URL for OAuth end-points in API docs

m-sameer:
Fix URL for OAuth end-points in API docs.

Fixes ushahidi/platform#3212 .

Ping @rowasc


tuxpiper:
Thank you @m-sameer !

PR #3451, opened 2018-12-10
feat: Cache form attributes

rjmackay:
- Add CachesData trait
- Use it in Form Attribute Repo to cache basic get requests

Test checklist:
- [x] composer test
- [x] I certify that I ran my checklist

Ping @ushahidi/platform

PR #3439, opened 2018-12-07
Import incidents to posts

rjmackay:
This pull request makes the following changes:
- Import incidents to post
- Update the default survey creation to save form attribute mappings
- Modify ImportMapping table to allow string source ids

Todo:
- [x] Bulk insert to post values for an efficient import #3453 
- [x] Write tests

Test checklist:
- [x] composer test
- [x] I certify that I ran my checklist

Fixes #3420 

Ping @ushahidi/platform

PR #3453, opened 2018-12-11
Batch insert post values and other migrations to speed up post import

rjmackay:
- Increase batch size to 200
- Merge in: https://github.com/ushahidi/platform/pull/3451
- Bulk insert post values
- Clean up weird behaviour when inserting post tags. We were looking up every single tag by tag, then by ID. Now I'm assuming a numeric value == an id. (And validation happens elsewhere)

Test checklist:
- [ ] composer test

Ping @ushahidi/platform

### DIFF
--- a/src/App/ImportUshahidiV2/Repositories/ImportMappingRepository.php
+++ b/src/App/ImportUshahidiV2/Repositories/ImportMappingRepository.php
@@ -26,7 +26,14 @@ class ImportMappingRepository /*extends EloquentRepository*/ implements ImportMa
 
     public function create(ImportMapping $model) : int
     {
-        return $model->save() ? $model->id : false;
+        $id = $model->save() ? $model->id : false;
+
+        if ($id) {
+            $key = "mapping.{$model->importId}.{$model->sourceType}.{$model->sourceId}";
+            Cache::tags(['import_mapping'])->put($key, $model->destId);
+        }
+
+        return $id;
     }
 
     public function createMany(Collection $collection) : array
@@ -38,6 +45,12 @@ class ImportMappingRepository /*extends EloquentRepository*/ implements ImportMa
         );
 
         $insertId = ImportMapping::resolveConnection()->getPdo()->lastInsertId();
+
+        // Save mappings to cache
+        $collection->each(function ($item) {
+            $key = "mapping.{$item->importId}.{$item->sourceType}.{$item->sourceId}";
+            Cache::tags(['import_mapping'])->put($key, $item->destId);
+        });
 
         return range($insertId, $insertId + $collection->count() - 1);
     }
--- a/src/App/Repository/Post/TagsRepository.php
+++ b/src/App/Repository/Post/TagsRepository.php
@@ -110,16 +110,18 @@ class TagsRepository extends ValueRepository
     protected function parseTag($tag)
     {
         if (is_array($tag)) {
-            $tag = $tag['id'];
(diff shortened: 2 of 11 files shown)


rjmackay:
It's worth noting that this and bits of #3439 could easily be applied to csv imports and make them much much faster.

src/App/ImportUshahidiV2/Repositories/ImportMappingRepository.php
```diff
@@ -26,7 +26,14 @@ class ImportMappingRepository /*extends EloquentRepository*/ implements ImportMa
 
     public function create(ImportMapping $model) : int
```
rjmackay: From my profiling, it looks a bit like adding the cache-on-write actually made this slower not faster, but I need to find a larger dataset to test on

src/App/Repository/Post/TagsRepository.php
```diff
@@ -88,16 +110,18 @@ public function updateValue($id, $value)
     protected function parseTag($tag)
     {
         if (is_array($tag)) {
-            $tag = $tag['id'];
+            return $tag['id'];
+        }
+
+        if (is_numeric($tag)) {
```
rjmackay: This makes a new assumptions that a numeric tag is always the ID. I did consider dropping this parsing entirely, but I think we might be relying on it for CSV imports to map tag names in the CSV to an ID.

src/App/Repository/PostRepository.php
```diff
@@ -1107,11 +1108,29 @@ public function createMany(Collection $collection) : array
         $newPostIds = range($insertId, $insertId + $created - 1);
 
         // Loop over entities, and aggregate values by attribute
-        $postValues = collect($newPostIds)->combine($collection)->map(function ($entity, $id) {
-            if ($entity->values) {
-                $this->updatePostValues($id, $entity->values);
+        $postsById = collect($newPostIds)->combine($collection);
+
```
rjmackay: If this doesn't make sense, I can try to make it more readable. It's an ugly transformation that *might* be clearer with a `foreach` instead of all the map/flatten/groupby stuff

src/App/Repository/Post/TagsRepository.php
willdoran: We are definitely relying on it so it needs to remain sadly, it's actually hugely useful in relation to tagging.

src/App/Repository/PostRepository.php
willdoran: I'm ok with this style but can you expand the comment to be more like:
"reformat set of Posts as...group by 'key' bulk insert attribitues then return X..."

src/App/ImportUshahidiV2/Repositories/ImportMappingRepository.php
willdoran: That's interesting we should definitely check to see why that might be

src/App/Repository/Post/TagsRepository.php
rjmackay: Do you think it needed to remain as it was? or is this partial refactor (assuming numeric = tag id) ok?

rjmackay: For the purposes of the importer I could pass tags as `['id' => 1]` style and that'd be completely safe to skip the lookup on

willdoran: No no I think this is fine

Issue #3435, opened 2018-12-06
Import reporters to contacts/users


rjmackay:
## Overview
Import reporters from v2 into contacts and users in v3. It's not immediately clear if we can do a 1:1 mapping. We might need to keep original data as a backup.

## Requirements

- All reporter data is preserved
- Reporters are associated with imported messages (or can be in messages task)
- Reporters are associated with imported users
- `Trusted` status is either added, or backed up in a mapped table.

## Test script

TBD

Aha! Link: https://ushahiditeam.aha.io/features/PROD-259

PR #3454, opened 2018-12-12
Import messages and reporters

### LINKED ISSUE #3427: Import v2 messages
Aha! Link: https://ushahiditeam.aha.io/features/PROD-261

rjmackay:
- Add import reporters step
   - Importers reporters as users + contacts
- Add import messages step
   - Imports messages to messages table and matches to imported contacts

Todo:
- [x] Add tests

Test checklist:
- [x] composer test

Fixes #3427 #3435 

Ping @ushahidi/platform

### DIFF
--- a/src/App/ImportUshahidiV2/Console/ImportUshahidiV2Command.php
+++ b/src/App/ImportUshahidiV2/Console/ImportUshahidiV2Command.php
@@ -98,8 +98,6 @@ class ImportUshahidiV2Command extends Command
         $this->info('Importing categories');
         $this->dispatcher->dispatchNow(new ImportUshahidiV2\Jobs\ImportCategories($importId, $dbConfig));
 
-        $this->info('Importing messages');
-
         // Import incidents to posts
         $this->info('Importing incidents to posts');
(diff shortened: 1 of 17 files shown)


rjmackay:
Because we're importing reporters into users, this potentially creates a lot of blank users. Some reporters have a name, but not many. There's a couple of ways we could resolve this:

- Modify our user model to be similar to TenFour. Users would become more like a 'Person' model, which might not actually be a user that can log in.
    - The changes needed are moving user email out of the user table into contacts and adding a flag to users to identify which entries are actual users who can log in, and which are just contacts/reporters.
    - This is a better data model in the long run. We'd have to filter out the reporters from the users from end and check other user related code.
- I could split the reporters' dataset by which ones have a name and only import those that have a name into users, and all the others would only go into contacts.
     - We'd still potentially have some junk users that aren't actually expected to log in and might not have an email

willdoran:
@rjmackay I think the path way of 
>Modify our user model to be similar to TenFour.

Is preferable. We would need to make sure we have tests that cover the basic user setup for regression testing.

PR #3458, opened 2018-12-13
refactor: Define Validator interface and rename old Validator to LegacyValidator

rjmackay:
- Define Validator interface as used by usecases
- Move old Validator tool into App and rename to LegacyValidator
- Remove old GetSet Trait that wasn't used
- Update existing Validators with new name

_This is to enable moving newer validators to use the laravel validator instead_

Test checklist:
- [ ] composer test

Fixes ushahidi/platform# .

Ping @ushahidi/platform


src/Core/Tool/Validator.php
```diff
@@ -11,67 +11,21 @@
 
 namespace Ushahidi\Core\Tool;
 
-use Ushahidi\Core\Entity;
-use Ushahidi\Core\Traits\GetSet;
-use Ushahidi\Core\Tool\Validation;
-use Ushahidi\Core\Tool\ValidationEngineTrait;
-
-abstract class Validator
+interface Validator
```
rjmackay: This is the key change in this PR. A validator really only needs to implement 2 functions, the rest was all Kohana related

PR #3460, opened 2018-12-17
Import media with incidents

rjmackay:
- Load incident media from v2 DB
- Save media into v3 DB
- Save into post media fields

This doesn't upload the media files yet. Need to look at options for that.

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


src/App/Repository/PostRepository.php
```diff
@@ -1145,6 +1150,27 @@ public function createMany(Collection $collection) : array
         return $newPostIds;
     }
 
+    public function createManyMedia($values)
```
rjmackay: This is unoptimized right now.

src/App/ImportUshahidiV2/Mappers/IncidentPostMapper.php
```diff
@@ -72,4 +87,34 @@ public function getCategories($importId, $categories)
             return $this->mappingRepo->getDestId($importId, 'category', $item);
         })->all();
     }
+
+    public function getMedia($media, $type, $user_id)
+    {
+        return collect($media)
+            ->where('media_type', $type)
+            ->map(function ($media) use ($type, $user_id) {
+                // Not sure what to do with non URL values yet, so just saving them as-is
```
rjmackay: This ^^. I'm not copying the media. Most of Crowdmap references an absolute URL on the CDN, so we can probably use that. Or do a manual copy for non-crowdmap deployments?

Issue #2890, opened 2018-05-02
Migrate scheduled tasks to use lumen scheduler and/or queues


rjmackay:
- Currently we run several schedule jobs via cron:
  - datasource:incoming
  - datasource:outgoing
  - webhooks send
  - notification:queue
  - savedsearch:sync

These should move to more appropriate places
- Use the lumen scheduler for datasource:incoming
- Move datasource:outgoing to jobs in a lumen job queue
- Move webhooks to jobs in a lumen job queue
- Move notifications to jobs in a lumen job queue
- Move saved search sync to event handlers, and maybe trigger jobs from there

The complex part of this is that jobs and scheduler need to be multisite aware.

#### It is ready for development when
- [ ] Has a description that includes purpose of this particular feature/improvement (how it will help the user and for which use cases)
- [ ] Can be delivered within a sprint (make sure issues are small pieces of work, not huge epics)
- [ ] Has all designs and UI assets available
- [ ] Has all (probable) dependencies identified
- [ ] Is estimated by the team
- [ ] Has acceptance criteria defined (do we want to do this before we send to dev? might help us identify dependencies/holes earlier?)

rowasc:
Robbie already opened an issue related to this , setting it in progress until we unblock his PRs

PR #3464, opened 2018-12-19
Run regular commands through artisan schedule:run

rjmackay:
- Schedule regular commands in Console\Kernel
- Update docs

Test checklist:
- [ ] composer test

#2890 

Ping @ushahidi/platform


tuxpiper:
mm yep it requires changes across the board. Unfortunately this is tricky because some deployments won't receive this change.

Is it the case that, before this change, the "artisan schedule:run" command would fail? This is something the deployment scripts could test for, in order to make a decision to use it or not.

rowasc:

PR #3485, opened 2019-03-23
vagrant installation error handling

NidhiKJha:
Solves the vagrant issue while running the platform.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,13 @@ If you get a warning like "In MemcachedConnector.php line 69:  Class 'Memcached'
 
    `vagrant up`
 
+ If you get an SSL error then run 
+
+   ```vagrant box add --insecure laravel/homestead```
+
+and retry with 
+
+   `vagrant up`
 Our vagrant box is built on top of Laravel's Homestead, a pre-packaged Vagrant box that provides a pre-built development environment. Homestead includes the Nginx web server, PHP 7.1, MySQL, Postgres, Redis, Memcached, Node, and all of the other goodies you might need.
 
 If you see an error like "Vagrant was unable to mount VirtualBox shared folders...", try upgrading VirtualBox or edit Homestead.yaml and change the folders to NFS as shown below, then re-run "vagrant" up.


tuxpiper:
Thank you @NidhiKJha for your contribution. Before merging this, I've got some changes to suggest. I have some qualms with what we would be suggesting people to do, if we added this to the documentation as is.

If someone has trouble with SSL/TLS, I would say it's *last resort* to entirely disable the protocol, as it means giving up not only the secrecy of communications, but also the assurance that they are authentic.

People could be exposing themselves to a man in the middle attack, where someone could send a malicious vagrant box, running code in the user's workstation without the user being aware (it would probably be a crypto miner, these days)

I think a better thing to try first, is to ensure their Vagrant setup is up to date, and in sync with the libraries present in the system. A way to do this is to uninstall vagrant and install it again.

Only if this fails, and a solution can't be found in Vagrant or Homestead issues and/or forums, I would suggest giving up encryption, with caution.

NidhiKJha:
@tuxpiper This is an issue me and my friend both faced during installation. I decided to make this PR after someone else also complained about it on the gitter chat.
Your concerns about security are perfectly valid but only in a production environment. Since most contributors will be just deploying locally, this serves as a quick and easy way to setup :)

tuxpiper:
@NidhiKJha  I understand this is a quick way to keep moving in face of such problem, and it is indeed valuable. We should have it in the documentation.

A lot of the attacks these days are not only on production servers, but on people's own devices as well. I agree a poisoned vagrant box that's going to run for limited time may not look like the end of the world, but it can be annoying. Also, in conjunction with other vulnerabilities, it *can* put someone at risk. As engineers working in a global environment that we don't fully control, we have to acknowledge that.

In summary, my point is, yes, we have a solution here, and there are other available and safe solutions to try before resorting to this one.

For merging this, we'll want to add a mention to consider reinstalling vagrant first, and also place the necessary warnings around disabling encryption.

NidhiKJha:
Thanks @tuxpiper :)

Could you please point me to some solutions.

rowasc:
@NidhiKJha Thanks for this PR to improve the docs. We appreciate it ! 


To being with, I'd add the following (before instructing them to run with --insecure)
- Before you continue: check that you are running the latest (stable) version of vagrant
    - To check this run `vagrant version` , It will tell you if you are in the latest version of vagrant or not.
    - If you are not in the latest vagrant, please update/replace vagrant with the latest version
Depending on the specific error regarding SSL (ie is it reporting a self signed certificate? or is it saying it's unable to find the local issuer certificate? ) 
     - After updating vagrant, clean up the local vagrant directories (ie .vagrant in your project's directory , ~/.vagrant/tmp ) and try again . 
- Other things to check and fix if necessary : Virtualbox is up to date

If that doesn't work, a user can start looking into why the issue is happening for them specifically.  This overflow answer might help as a general starting point https://stackoverflow.com/a/42209673
although it'll be environment/situation specific in the end to fix it. 
There's also the relatively new config `config.vm.box_download_ca_cert` which can be useful if we know the box url and can check with `curl -v <box url> ` to see what certificate is being used to fetch it and wether or not that's working 
pinging @tuxpiper about this: for some reason I cannot figure out the box url for homestead, it seems to have changed and I can only access the html view https://app.vagrantup.com/laravel/boxes/homestead-7/versions/0.2.1 , do you know how that would be done now?

tuxpiper:
@rowasc 
> pinging @tuxpiper about this: for some reason I cannot figure out the box url for homestead, it seems to have changed and I can only access the html view https://app.vagrantup.com/laravel/boxes/homestead-7/versions/0.2.1 , do you know how that would be done now?

I really have no idea :( would need to run some research

rowasc:
I'm closing this PR and moving the discussion to a ticket as it cannot be merged at the time

Issue #3529, opened 2019-05-04
the link to heroku 1 click deploy is not right


lexoyo:
**Describe the bug**

Note: I reported [the same bug on the client side](https://github.com/ushahidi/platform-client/issues/1295) and will propose a PR if it is ok for you. It is a small bug but very confusing to new users because you need to deploy the repo by hand and then you need to know all the env vars and create each one in the settings

When I click on the image "deploy to heroku" in the readme, I should have a UI which asks me the values for the required env vars, cf screenshot bellow

**Where was the bug observed**

The project [readme file](https://github.com/ushahidi/platform/blob/develop/README.md)

**To Reproduce**

1- Open [the readme](https://github.com/ushahidi/platform/blob/develop/README.md)
2- click on "deploy to heroku" (
![Screenshot from 2019-05-03 19-58-14](https://user-images.githubusercontent.com/715377/57171815-d1688a00-6ddd-11e9-9f90-f9e64007e053.png) )
3- You will be directed to `https://heroku.com/deploy` instead of the 1 click install UI of heroku

**Expected behavior**

I should be redirected to heroku 1 click UI which asks me the values for the required env vars

**Screenshots**

Heroku 1 click deploy UI which asks me the values for the required env vars, like this:
![Screenshot from 2019-05-03 19-55-06](https://user-images.githubusercontent.com/715377/57171780-6f0f8980-6ddd-11e9-8abc-dbc1553dac7b.png)

**URL / Environment where this happened**

N/A

PR #3530, opened 2019-05-04
Fixed link to heroku 1 click deploy

lexoyo:
The link to heroku now takes the file app.json into account so that the developer can see a nice UI to provide values to required env vars

- In the readme, the link leads to 1 click deploy

Testing checklist:
- [x] link to 1 click deploy

Fixes ushahidi/platform#3529 .

Ping @ushahidi/platform


README.md
```diff
@@ -19,7 +19,7 @@ Ushahidi 3
  [![Sponsors on Open Collective](https://opencollective.com/platform/sponsors/badge.svg)](#sponsors) 
 
 
-[![Deploy](https://www.herokucdn.com/deploy/button.png)](https://heroku.com/deploy)
+[![Deploy](https://www.herokucdn.com/deploy/button.png)](https://heroku.com/deploy?template=https://github.com/ushahidi/platform/tree/master)
```
rowasc: @tuxpiper should this point to master or develop? I think master but just checking in case you disagree 
do you see any issues with this approach (aside from the fact that we need to "hardcode" the branch in)

rowasc: Thanks
@lexoyo for submitting this 💯 
Question: do you think we should explicitly call out all the ENV params in the app.json ?

https://stackoverflow.com/questions/44026881/deploy-to-heroku-button-with-env-variables 
https://devcenter.heroku.com/articles/heroku-button#creating-the-app-json-file
was just checking this out and we could define the vars from https://github.com/ushahidi/platform/blob/master/.env.example in app.json too 

cc @tuxpiper

lexoyo: Usually it is the master branch because the use case is "As a developer I want to host an instance on heroku to show my boss that it is easy to self host this tool"

lexoyo: > we could define the vars from https://github.com/ushahidi/platform/blob/master/.env.example in app.json too

Yes sure

lexoyo: maybe for that we should merge this and start an issue to discuss what we want in a second step?

my PR is just a fix, the app.json was already there

tuxpiper:
@lexoyo how is it working for you? Are you in a heroku paid plan or in the free tier? Last time I tried, the free heroku tier would not complete the database migrations, because they were too big.

lexoyo:
Hi @tuxpiper 
Sorry for the late answer
I was only trying to fix the link in a first step

Since then I had the chance to test heroku and I believe the problem is with the conexion to the DB. Heroku provides an env var `CLEARDB_DATABASE_URL` with the DB URL but it looks like the migrate script does not use it. 

Do you have a pointer where to change that? Where is the `.env` used by the migrate script in the context of heroku?

tuxpiper:
@lexoyo I remember taking a stab at this some time ago, pls check this branch diff: https://github.com/ushahidi/platform/compare/3305-fix-heroku-deployment

The issue I ended up running into was that the free database tier in heroku sets up a rate limit on the number of transactions and that was quickly exceeded by phinx

PR #3615, opened 2019-08-08
[Release prep] develop to master

rowasc:
-

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


app/Console/Kernel.php
```diff
@@ -26,7 +26,6 @@ class Kernel extends ConsoleKernel
         \Ushahidi\Console\Command\ApikeySet::class,
         \Ushahidi\Console\Command\ConfigSet::class,
         \Ushahidi\Console\Command\ConfigGet::class,
-        \Ushahidi\Console\Command\Import::class,
         \Ushahidi\Console\Command\UserCreate::class,
```
rowasc: I don't remember why we needed to delete this :/ https://github.com/ushahidi/platform/pull/3430/files I wish I had notes on the decision leading to this change . Note to review after the rest of the code

rowasc: OK so it looks like this change was a pre requirement for this , which is moving the imports to queues and making them more useful in the process https://github.com/ushahidi/platform/pull/3407/files

But this means we probably want to roll it back until we have the new system with queuws in place and the necessary UI changes so that the user understands what's happening. @tuxpiper thoughts?

tuxpiper:
Thank you x1M for kicking this off @rowasc !

docs/using-the-ushahidi-platform-v3-v4/untitled.md
```diff
@@ -0,0 +1,11 @@
+---
```
tuxpiper: I think this file can be discarded . There is documentation for the v3 to v4 upgrade in master already: https://github.com/ushahidi/platform/tree/master/docs/upgrading-ushahidi  .

I'm not 100% sure if this document was intending to explain anything else...

app/Console/Kernel.php
tuxpiper: We do have platform api queue worker instances for ushahidi.io (in production and steve-buscemi) running `artisan queue:work` (for the exports). I think the import stuff should fit right in :+1:

docs/untitled-1.md
```diff
@@ -8,5 +8,3 @@ That's a tough question but thankfully, our team is on it. Please bear with us w
 
 Yes, after a few months we finally found the answer. Sadly, Mike is on vacations right now so I'm afraid we are not able to provide the answer at this point.
```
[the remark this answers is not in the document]
rowasc: yup, done

app/Console/Kernel.php
rowasc: ah! alright. Does that mean we could merge the import queues code then?

tuxpiper: I think we should definitely try! Is the plan to merge to master and then QA , or do some QA on develop first? I would suggest the latter if possible.

rowasc: Let's do develop first if we plan to merge the import queues work https://github.com/ushahidi/platform/pull/3407/files

rowasc: actually let's do it in develop either way :D at least a smoke test and then a regression in master?

tuxpiper: since develop is continuously deployed to steve-buscemi , we should be good to do that smoke test anytime

tuxpiper:
I just found out that PHP 7.2 and PHP 7.3 are running the tests properly .

Given that PHP 7.1 is out of active support already and will drop out of security support in December , maybe we should switch our testing and production systems to 7.2 or 7.3?

rowasc:
Yup that sounds reasonable lets do it (I think you mentioned doing it in sad Charlie first no? )

tuxpiper:
> Yup that sounds reasonable lets do it (I think you mentioned doing it in sad Charlie first no? )
Yess a refreshed sad charlie is coming together. I was feeling a bit conservative and went for 7.2

Issue #3264, opened 2018-08-31
User search in /settings/users is not working


rowasc:
**Describe the bug**
I don't see any users when I search by their real name, email or part of their email 

**Where was the bug observed**
Include Deployment and page URL (e.g. "https://test.ushahidi.io/views/map")

**To Reproduce**
1. Go to Settings -> Users as an admin
2. Use the search bar to search for users (search for an existing user)
3. You won't see any results

**Expected behavior**
3. You should see the user that matches the email

**Is there a workaround? What is it.**
Looking in the list to find a user without search

**URL / Environment where this happened**
http://rominacsvsms.ushahidi.io/

**Desktop Hardware Details (please complete the following information):**
 - OS: OSX
 - Browser : Chrome
 - Version: 68.0.3440.106

bobinson:
Fixing this can be of great help as we have volunteers registering and assigning roles to them is very difficult right now.

rowasc:
Thanks @bobinson for your feedback..  I think I have a fix that works with emails and usernames as expected, but it needs some QA still (mostly to make sure there's no unexpected issues) . I've asked @Obadha2 and @tuxpiper to take a look and I'll try to schedule the release for Monday in my afternoon if it all looks good for them too.
Thank you for your patience.

rowasc:
Hey @Obadha2 we deployed this today thanks to @tuxpiper who reviewed the code 🎉 . Could you test in Steve-buscemi when you come online? 
Sadly I need to run out to the hospital to deal with a family emergency right now  :( so I'm not releasing today as I originally planned, but any updates on how the testing goes would be great.
 
Thank you all for your patience and help as we fix this issue.

rowasc:
We just released this. Closing!  I was able to search/find my users in my deployment (phew).
Let us know how it goes if you try it? 
Thanks! 

Separate note: While doing testing for this specific fix, @Obadha2 discovered other pre-existing bugs in the same  area (filters & user filters) we are hoping to tackle soon. 

https://github.com/ushahidi/platform/issues/3649 UX issue that I'm aware of
and I think there were another few UX things he noticed weren't right
 @Obadha2 can you link them up when you start working tomorrow?

Obadha2:
#3656

PR #3616, opened 2019-08-09
fix(UserSearch): left join with contacts for user search

rowasc:
We should be searching in contacts and users, and not require that a contact exists to find a user through settings. 

- Changes inner join to left join so we can find users without contacts. 

Notes to reviewer/context on this change: 
I don't believe a contact entry should be a requirement to be able to find users? Lmk what you think @tuxpiper . I did some digging because I wasn't sure where this came from, for what it's worth before switching to using an inner join, we were doing an `OR WHERE IN()`  but I think we introduced this bug [here](https://github.com/ushahidi/platform/pull/2359 ) when we switched to an inner join a while back, doing the contact hydration changes.

Test checklist:
- [ ] https://github.com/ushahidi/platform/issues/3264

Fixes ushahidi/platform# .

Ping @ushahidi/platform#3264

### DIFF
--- a/src/App/Repository/UserRepository.php
+++ b/src/App/Repository/UserRepository.php
@@ -147,7 +147,7 @@ class UserRepository extends OhanzeeRepository implements
             $query->and_where_close();
 
             // Adding search contacts
-            $query->join('contacts')->on("$table.id", '=', 'contacts.user_id')
+            $query->join('contacts', 'left')->on("$table.id", '=', 'contacts.user_id')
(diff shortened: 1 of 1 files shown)


rowasc:
@tuxpiper also if you approve this on Monday, could you merge it and let @Obadha2 know so he can test it on Steve-buscemi  before I start my day? (I'm assuming you're like 5hs ahead of me, maybe I'm wrong :D ). Thanks <3

tuxpiper:
@Obadha2 this has been deployed to steve-buscemi , it should be good to test 👍

Issue #3506, opened 2019-04-03
API pre-install hook  – check PHP CLI version and extensions


tuxpiper:
Aha! Link: https://ushahiditeam.aha.io/features/PROD-236

tuxpiper:
Could this issue be actually speaking about duplicating functionality already available in composer? See `check-platform-reqs` command: https://getcomposer.org/doc/03-cli.md#check-platform-reqs

PR #3646, opened 2019-08-12
#3506 update composer.json PHP and extension requirements

tuxpiper:
- explicitly sets which PHP versions are actually supported
- adds a bunch of extended PHP extensions

We may still have to do something around "soft" requirements, such as redis, memcache etc. Those are not always required, but may be required if the .env configuration requires their use.

Test checklist:
- [ ]  run composer install 

Fixes ushahidi/platform#3506 .

Ping @ushahidi/platform

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,16 @@
         }
     ],
     "require": {
-        "php": ">=7.0 <=7.3",
+        "php": ">=7.0 <7.2",
+        "ext-curl": "*",
+        "ext-gd": "*",
+        "ext-json": "*",
+        "ext-mbstring": "*",
+        "ext-mcrypt": "*",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -15,7 +15,16 @@
         }
     ],
     "require": {
-        "php": ">=7.0 <=7.3",
+        "php": ">=7.0 <=7.1",
```
rowasc: I'm wondering if this should be < 7.2 to include all the minor versions of 7.1 ?

tuxpiper: You are very right :+1: and there's other stuff I missed . Preparing a new commit

tuxpiper: Fixed

PR #3774, opened 2019-09-26
Composerfix

rowasc:
- Fixing composer.lock because we should support php70 
 Note: I don't know that we want to do this actually, we probably should just drop php7.0 support for the release now 
Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


composer.lock
```diff
@@ -459,38 +459,33 @@
         },
         {
             "name": "doctrine/cache",
-            "version": "v1.8.0",
+            "version": "v1.6.2",
```
rowasc: 1.8.0 does not work in php70

PR #3795, opened 2019-10-19
transifex friendly format for lang strings

tuxpiper:
- transifex can't parse the short array format, this reverts to the long format

Ping @ushahidi/platform


rowasc:
is this about a php transifex module that isn't compatible? confused , maybe I'm lacking some context?

tuxpiper:
> is this about a php transifex module that isn't compatible? confused , maybe I'm lacking some context?

Sorry, my description was very barebones. The need for this adjustment is made manifest when uploading these files to transifex, so that it parses them and is made aware of which strings are contained in the translation project.

Here are the formats supported by transifex: https://docs.transifex.com/formats/php

tuxpiper:
oh yea, the linter is not happy... huh, I thought we had it running along with the tests. Codeship hasn't complained.

tuxpiper:
@rowasc added linter parameters to ignore these files .

I chose to ignore the whole folder thinking that in the future we may check out from transifex translations and place them there. Who knows how well formatted they are going to come...

PR #3811, opened 2019-11-15
fix(PostValidation): title, content and form_id are required fields

rowasc:
- Improves validation rules for Post::create 

Test checklist:
(requires sending requests directly without the client ie through curl/postman/whatever) 

- [ ] Submit a post without a title. You should receive a validation error
- [ ] Submit a post without a content. You should receive a validation error
- [ ] Submit a post without a form id. You should receive a validation error
- [ ] Submit a post with content, title and form id. It should be created 

Fixes ushahidi/platform#3913 .

Ping @ushahidi/platform

Aha! Link: https://ushahiditeam.aha.io/features/PROD-780


webong:
Hey, @rowasc just checking and noticed my PR https://github.com/ushahidi/platform/pull/3878 tries to solve the same issue here ... I did make a change that stopped the tests from failing, could you have a look at it?

rowasc:
@webong yea, thanks for flagging. I was just thinking about that last night. I'm not a fan of the approach in https://github.com/ushahidi/platform/pull/3878 changing to not be recursive because the entire app validation works through those merge_recursive statements (so it would be less consistent) so it may be good to add the testing suite to this one since I apparently never got around to it...unless we are willing to basically redo a bunch of our validators. Totally a fan of tests passing :) Would you be willing to change the approach to something more like this? If so we can just combine efforts . 
thanks again, you are killing it!

anje123:
@rowasc while reviewing this issue I made a PR #3914, but I noticed they were some other issues related to this,  as I stated here #3913, but will it be ok, if the ushahidi/kohana-validation is edited and installed back or do we need to create another class manually in the ushahidi platform to handle that, or maybe am I missing something.

tuxpiper:
Closing in favor of #3991

tuxpiper:
Reopening because #3991 didn't actually solve everything specified as a test checklist in this PR

PR #3822, opened 2019-11-28
fix(replica-host): we need to check for empty to make up for the fact…

rowasc:
… that ''  is considered blank for our templates

-

Test checklist:
- [ ] ugh

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -59,7 +59,7 @@ return [
         'mysql' => [
             'host'      => env('DB_HOST', 'localhost'),
             'read' => [
-                'host' => env('DB_HOST_REPLICA', env('DB_HOST', 'localhost')),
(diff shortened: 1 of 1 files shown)


config/database.php
```diff
@@ -59,7 +59,7 @@
         'mysql' => [
             'host'      => env('DB_HOST', 'localhost'),
             'read' => [
-                'host' => env('DB_HOST_REPLICA', env('DB_HOST', 'localhost')),
+                'host' => !empty(env('DB_HOST_REPLICA')) ? env('DB_HOST_REPLICA') : env('DB_HOST', 'localhost'),
```
rowasc: so I have to explicitly check for empty sadly. This kinda sucks, it should be changed once we solve the underlying config problem we discovered.

tuxpiper: yep that's quite an eyeful 😄

PR #3831, opened 2020-01-09
Fix hdx and send it to develop for everyone

rowasc:
- Fixes bug that prevents datasets with *only* tags from being correctly generated
- Fixes issue that generated csv urls incorrectly when sent to the hdx endpoint
- Enables user-settings and hdx (HDX depends on user-settings, which was launched separately)

Test checklist:
- [  ]  I can setup my API key and userrname for HDX
- [  ]  I can change my API key and username
- [  ]  I can upload all my data to HDX
- [  ]   I can select only some fields to upload to HDX
- [  ]   I can select *tags* only (and it generates the CSV+HXL file correctly with the correct tags)
- [  ]   I can select *tags and attributes* (and it generates the CSV+HXL file correctly with the correct tags)
- [  ]   I can see the HDX exports in the export history tab
- [  ]   I get notified when an HDX upload succeeds
- [  ]    I can download a CSV+HXL file instead of uploading to HDX
- [  ]    I can leave the page while the CSV and HDX processes run
- [  ]   I certify that I ran my checklist

Fixes ushahidi/platform# .

Ping @ushahidi/platform


rowasc:
@tuxpiper did what I could to improve some bugs that I found while testing, and moved a formatter to a trait when it needed to be duped for another class but looking for any useful feedback here. Trying to get this merged next week so we can start testing it with @Obadha2 , since apparently there are no completed test scripts available for it so it'll take a bit of work to develop them from scratch pretty much.

PR #3834, opened 2020-01-22
WIP feat(LangHeaders): add language headers and another es file 

rowasc:
- Accepts language header and uses it

TODO: there are around 30 sprintf statements that do not use translations at all and need to be addressed (example: https://github.com/ushahidi/platform/blob/add-lang-headers/src/Core/Tool/ValidatorTrait.php#L53). I did a few as an example here when I found the fix.

Example of the missing part in first commit:
Screen Shot 2020-01-21 at 23 08 50

Second commit: 
Screen Shot 2020-01-22 at 00 12 39

Open questions: should we translate entity names? Sounds gross I do not want to do it. Also, this whole thing where we state the error twice is not intuitive. Showing both messages isn't great.. I seem to remember there was a reason for this structure, maybe for when there are multiple errors per field? Not sure but I think that was it

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


app/Http/Middleware/SetLocale.php
```diff
@@ -0,0 +1,23 @@
...
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @param  mixed  ...$features
+     * @return mixed
+     */
+    public function handle($request, Closure $next, ...$features)
+    {
+        $locale = $request->header('Accept-language');
```
tuxpiper: @rowasc  how does this work with complex but still valid values of the accept-language header? i.e. `fr-CH, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5` ?

rowasc: semed to work fine, luckily. I actually had the same concern

rowasc: gross no, I think I was half asleep when I checked this. Only supports `en` / `es` etc right now (because all it does is check if the language folder exists, so if there's no `es-ES blablabla` folder it won't find it)

So we need a small parser to tries to find es-ES, if it doesn't it fallbacks to es, and if it doesn't it fall backs to en.
--

tuxpiper:
@rowasc  was this good to review? I can't recall where we left it

rowasc:
@tuxpiper yea this was for review sorry I left it in my vacation 😬

tuxpiper:
looking and working good to me :+1: merging

Issue #3836, opened 2020-01-28
Error-message when trying to create new role is not making sense


Angamanga:
**Describe the bug**
When trying to save a new role, and no Name or associated permissions is added, the error-message that appears is not informative:

![Screenshot 2020-01-28 at 13 11 40](https://user-images.githubusercontent.com/8624777/73264186-7334d000-41d2-11ea-8977-3ce32424f3fb.png)

**Where was the bug observed**
https://annatest1000.ushahidi.io/settings/roles/create

**To Reproduce**

1. Login as admin
2. Go to settings-> Roles-> create a new role
3. Do not fill in name or role
4. Click on save

rowasc:
https://github.com/ushahidi/platform/pull/3837 got bored waiting for tests to run and fixed this one.. I think ? 😬

Angamanga:
@Obadha2 Could you QA in Steve Buscemi?
Testing-checklist:
1. Login as admin
2. Go to settings-> Roles-> create a new role
3. Do not fill in name or role
4. Click on save
- [ ] The error-message should look something like this: 
![Screenshot 2020-03-16 at 10 56 45](https://user-images.githubusercontent.com/8624777/76744651-df649680-6774-11ea-891c-2263af866f33.png)

Obadha2:
QA'd. While this passes, could we have "name must not be empty" appear just once, instead of twice? @Angamanga

Angamanga:
I think there is a reason for that I don't remember (even if it is pretty confusing). Do you remember @rowasc ?

rowasc:
yea... I think it's just a bad UI , it happens for other fffields too

PR #3837, opened 2020-01-30
fix(3836): roles error message when creating should reflect the problem instead of a generic runtime error

rowasc:
- Uses the update role rules + the create rules merged recursively so that it doesn't just override the `name` rules in one of them
- stops throwing a runtime exception for roles creation errors 🤷‍♂ 🤷‍♀ 

Test checklist:
- [ ]

Fixes ushahidi/platform#3836 .

Ping @ushahidi/platform


rowasc:
@Angamanga @tuxpiper I apparently forgot to ask for a code review here. If anyone has some bandwidth, check it out :D this basically makes it so that we don't throw runtime errors to users and fixes the create rules for roles which were just overriding the name validation 😬

PR #3860, opened 2020-03-04
chore(translations): add transifex CLI tool config and contributed tr…

tuxpiper:
…anslations

- adds config for `tx` tool to pull translations from transifex into the proper file locations
- adds translations contributed to transifex up to this point 

Test checklist:
- [x] install and configure transifex CLI on local
- [ ] run `tx pull -a --minimum-perc 1 --mode onlytranslated`
- [ ] check diffs on new and updated files

Fixes ushahidi/platform# .

Ping @ushahidi/platform


rowasc:
@tuxpiper q: --minimum-perc=1 means if there's at least 1% of completion then the language is downloaded/available? 

@Obadha2 is this something you could test in the morning? 
THanks

tuxpiper:
> @tuxpiper q: --minimum-perc=1 means if there's at least 1% of completion then the language is downloaded/available?

yess , this is as opposed to setting nothing , which in combination with the `-a` flag would result in files being created even for languages that don't have a single translated string

tuxpiper:
Merging for deploy in QA env

PR #3863, opened 2020-03-05
feat: send user email when password is changed

webong:
- Adds a UpdateUser usecase class  
- Notifies a user via email when the password changes.

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


webong:
[clabot:check]

Angamanga:
@tuxpiper @rowasc Could one of you review this pr? 🙏

tuxpiper:
Thank you @webong, this is a valuable contribution :+1:

Before we go ahead though, I would like to consult with my colleague @rowasc  and yourself as well about one thing.

* I see that the new usecase essentially repeats the steps of the base class `interact()` method , with the addition of the step for sending the e-mail notification. In terms of following Don't-Repeat-Yourself principles, I think this may not be ideal.

I can think of two other approaches which may be cleaner:

1. Incorporate the notification step into the base class `interact()` method.. say by adding a call to a generic `sendNotifications()` method which could be just empty in the base class, and be overridden by the e-mail sending in the new `UpdateUser` class.

2. Invoke the super-class `interact()` method   (`parent::interact();`) from the `UpdateUser` class own `interact()` implementation , followed by the call to `notifyUserPasswordChanged`

What do you folks think? Myself, I think that dealing with notifications is a generic enough aspect of the general Update use case, and should probably be included in the general `interact()` sequence of actions.

webong:
@tuxpiper thanks for the review ... I think the first approach is much better ... currently adjusting the UpdateUser class to work with the approach.

rowasc:
Agree on approach *1*. I can already think of at least a few use cases beyond users where having a general notifications approach would work out great.  🎉 a super exciting change ! thank you @webong for your work here

tuxpiper:
Thanks! This is looking good to me now :+1:

@rowasc would you like to take a look?

tuxpiper:
Thanks all!

Tagging @Angamanga for consideration to add in next release

Angamanga:
@all-contributors Please add @webong for code

Issue #3672, opened 2019-08-23
API Gateway Checker – add automated tests


tuxpiper:

individual-it:
WIP PR: #3880

individual-it:
#3880 ready for review

PR #3880, opened 2020-03-16
test: gateway checker

individual-it:
- added tests for gateway checker

Test checklist:
- [x] :robot:

Fixes ushahidi/platform#3672 .

Ping @ushahidi/platform

CC @tuxpiper is that what you had in mind?

### DIFF
--- a/tests/integration/bootstrap/RestContext.php
+++ b/tests/integration/bootstrap/RestContext.php
@@ -26,6 +26,9 @@ class RestContext implements Context
     private $restObjectType    = null;
     private $restObjectMethod  = 'get';
     private $client            = null;
+    /**
+     * @var \GuzzleHttp\Psr7\Response
+     */
     private $response          = null;
     private $requestUrl        = null;
     private $apiUrl            = 'api/v3';
@@ -36,6 +39,8 @@ class RestContext implements Context
     ];
     private $postFields        = [];
     private $postFiles         = [];
+
+    const DEBUG_MODE_SWITCH_FILE_PATH = __DIR__ . "/../../../bootstrap/install_debug_mode.enabled";
 
     /**
      * Initializes context.
@@ -202,6 +207,24 @@ class RestContext implements Context
     }
 
     /**
+     * @Given /^that I have enabled debug mode$/
+     */
+    public function thatIHaveEnabledDebugMode()
+    {
+        fopen(self::DEBUG_MODE_SWITCH_FILE_PATH, "w");
+    }
+
+    /**
+     * @AfterScenario
+     * @Given /^that I have disabled debug mode$/
+     */
+    public function disableDebugModeAfterScenario()
+    {
+        if (file_exists(self::DEBUG_MODE_SWITCH_FILE_PATH)) {
+            unlink(self::DEBUG_MODE_SWITCH_FILE_PATH);
+        }
+    }
+    /**
      * @When /^I request "([^"]*)"$/
      */
     public function iRequest($pageUrl)
@@ -211,9 +234,9 @@ class RestContext implements Context
         switch (strtoupper($this->restObjectMethod)) {
             case 'GET':
                 $request = (array)$this->restObject;
-                $id = ( isset($request['id']) ) ? $request['id'] : '';
+                $idAttachment = ( isset($request['id']) ) ? "/" . $request['id'] : '';
                 $response = $this->client
-                    ->get($this->requestUrl.'/'.$id, [
+                    ->get($this->requestUrl.$idAttachment, [
                         'query' => isset($request['query string']) ? trim($request['query string']) : null,
                         'headers' => $this->headers
                     ]);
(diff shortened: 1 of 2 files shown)


tests/integration/bootstrap/RestContext.php
```diff
@@ -201,6 +206,24 @@ public function thatThePostFileIs($fieldName, $fieldValue)
         $this->postFiles[$fieldName] = $fieldValue;
     }
 
+    /**
+     * @Given /^that I have enabled the debug mode$/
```
phil-davis:
```suggestion
     * @Given /^that I have enabled debug mode$/
```

tests/integration/bootstrap/RestContext.php
```diff
@@ -201,6 +206,24 @@ public function thatThePostFileIs($fieldName, $fieldValue)
...
+     */
+    public function thatIHaveEnabledTheDebugMode()
+    {
+        fopen(self::DEBUG_MODE_SWITCH_FILE_PATH, "w");
+    }
+
+    /**
+     * @AfterScenario
+     * @Given /^that I have disabled the debug mode$/
```
phil-davis:
```suggestion
     * @Given /^that I have disabled debug mode$/
```

tests/integration/gwchecker.feature
```diff
@@ -0,0 +1,68 @@
...
+        And the "data._SERVER.HTTP_ACCEPT_ENCODING" property is empty
+        And the "data._SERVER.HTTP_ACCEPT_LANGUAGE" property is empty
+        And the "data._SERVER.HTTP_AUTHORIZATION" property is empty
+        And the "data._SERVER.HTTP_CONNECTION" property is empty
+        And the "data._SERVER.HTTP_ORIGIN" property is empty
+        And the "data._SERVER.HTTP_REFERER" property is empty
+        And the "data._SERVER.ORIG_PATH_INFO" property is empty
+        And the guzzle status code should be 200
+        Examples:
+            | endpoint       | true      |
```
phil-davis:
```suggestion
            | endpoint       | gwCheckValue      |
```
Will the scenarios read more easily if the column name is changed to something like this?

tests/integration/gwchecker.feature
```diff
@@ -0,0 +1,68 @@
...
+
+    Scenario Outline: debug information should not be visible with disabled GW checker
+        Given that I have disabled the debug mode
+        And that I want to get all "Configs"
+        When I request "/config?gwcheck=<true>"
+        Then the response is empty
+        And the guzzle status code should be 204
+        And the "X-Ushahidi-Platform-Install-Debug-Mode" header should be "off"
+        Examples:
+            | true      |
```
phil-davis:
```suggestion
            | gwCheckValue      |
```
Same thought as above. "Naming is hard"

individual-it:
@phil-davis I've addressed your requests

tests/integration/gwchecker.feature
```diff
@@ -0,0 +1,68 @@
...
+        And the "data._SERVER.REQUEST_METHOD" property equals "GET"
+        And the "data._SERVER.QUERY_STRING" property equals "gwcheck=<gwCheckValue>"
+        And the "data._SERVER.HTTP_ACCEPT" property equals "application/json"
+        And the "data._SERVER.SCRIPT_NAME" property contains "/index.php"
+        And the "data._SERVER.REQUEST_URI" property equals "/api/v3/<endpoint>?gwcheck=<gwCheckValue>"
+        #the following line might fail when running tests on Apache
+        And the "data._SERVER.PATH_INFO" property equals "/api/v3/<endpoint>"
+        And the "data._SERVER.HTTP_USER_AGENT" property contains "GuzzleHttp"
+        And the "data._SERVER.DOCUMENT_ROOT" property contains "/platform"
+        And the "data._SERVER.SCRIPT_FILENAME" property contains "/platform/httpdocs/index.php"
```
rowasc, line 32: so this fails in codeship, where we appear to  have /httpdocs as doc root, not platform. Should we make it configurable maybe?
@tuxpiper @individual-it 

(sorry I didn't notice earlier 😬 )

rowasc: PHP 7.2.28 Development Server started 
 test/php-7.2 Listening on http://localhost:8000
 test/php-7.2 Document root is /var/www/httpdocs

individual-it: @rowasc that is strange, because it is the same in travis,
the root starts in `/home/travis/build/ushahidi/platform/httpdocs` the higher levels of the tree should not make any difference.

image

I cannot access the codeship build to debug :-(

rowasc: https://github.com/ushahidi/platform/blob/develop/codeship-services.yml#L4
https://github.com/ushahidi/platform/blob/develop/docker/build.Dockerfile#L3 

I think this is what makes the difference for the codeship script from what I'm seeing . It's the same in the other dockerfiles we have setup ie the test one. 

Notice /platform is higher up than httpdocs in the example, but this test checks for /platform's existance, so if we don't have it, it wouldn't work 
ie: /var/www/httpdocs vs /var/www/*platform*/httpdocs

individual-it: ahh I see, overlooked that, there is no `platform` at all. Let me have a go at it

individual-it: an other try in #3911

PR #3881, opened 2020-03-18
test: exceptions of invalid receive

individual-it:
- adding more test-cases for invalid receive calls in `DataSourceStorage.php`

Test checklist:
- [x] :robot: 

Fixes ushahidi/platform# .

Ping @ushahidi/platform


tests/unit/App/DataSource/DataSourceStorageTest.php
```diff
@@ -79,10 +79,45 @@ public function testReceive()
...
+                "\InvalidArgumentException",
+                "Symfony\Component\HttpKernel\Exception\HttpException",
+                400,
+                "Bad request: ; sorry my fault, but you are also not perfect"
+            ]
+        ];
+    }
+
     /**
-     * @expectedException Symfony\Component\HttpKernel\Exception\NotFoundHttpException
```
individual-it: using try/catch block so that also `getStatusCode` can be checked

tests/unit/App/DataSource/DataSourceStorageTest.php
```diff
@@ -105,31 +140,34 @@ public function testFailedReceive()
...
-            "YOU!",
-            null,
-            null,
-            null,
-            null,
-            1,
-            ['Title' => 'somekey']
-        );
-
-        $e->shouldHaveReceived('getMessage')->once();
```
individual-it: that line was never ran and even if it would have been it would not worked, because `getMessage` cannot be mocked see https://github.com/mockery/mockery/issues/581

individual-it:
is this PR any use at all, or should we close it?
@tuxpiper maybe?

individual-it:
closing because of no reponse

Issue #3804, opened 2019-11-14
"Outgoingemail" should not be a filter-option


Angamanga:
**Describe the bug**
In the platform-map-view, when logged in, there is an option for "Outgoingemails" in the data-source filters:

<img width="364" alt="Screenshot 2019-11-14 at 14 50 06" src="https://user-images.githubusercontent.com/8624777/68862685-2de22580-06ee-11ea-9de6-1ce7ec3315da.png">

This option should not be there.

**Where was the bug observed**
https://annatest1000.ushahidi.io/views/map when logged in

**To Reproduce**
1. Go to a deployment with email as a datasource set up
2. Log in as an admin
3. Go to map-view
4. Scroll down to datasources in the mode-context-bar
5. See the option in the list

**Expected behavior**
This option should not be available

bkplaravel2018:
Hello @Angamanga , I fixed this issue and committed also but I can't push my code. I got the error:  Tests\Unit\Core\Usecase\ReceiveMessageTest::testReceiveMessage Ushahidi\Core\Exception\ValidatorException: Failed to validate messages entity: data_source must be one of the available options. Please guide me how can I resolve?

bkplaravel2018:
Please review/accept the PR @Angamanga

PR #3890, opened 2020-03-22
Fix - "Outgoingemail" should not be a filter-option.

noone0212:
- Now, "Outgoingemail" is removed from filter options.

Test checklist:
- [ ]

Fixes ushahidi/platform#3804 .

Ping @ushahidi/platform

### DIFF
--- a/src/App/Repository/Config/data-provider.php
+++ b/src/App/Repository/Config/data-provider.php
@@ -12,13 +12,13 @@
 return [
     'providers' => [
         // List of data providers key=provider value=enabled
-        'smssync' => true,
+        'smssync' => false,
         'email' => true,
         'outgoingemail' => false,
-        'twilio' => true,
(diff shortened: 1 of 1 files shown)


noone0212:
clabot:check

rowasc:
@noone0212 pretty sure this breaks the outgoing email for messages, since it uses datasources. 
Checking out the git history, I can see that at one point we added isUserConfigurable() to datasources, and email is the only one which has it as "false" . It's likely that this bug is either in the way we filter datasourceproviders to send them to the client, or in which function is being called when trying to obtain the datasource lists. 
https://github.com/ushahidi/platform/commit/015aea4a47073888fbd5ceb4d3916c6eb347e9e3#diff-ab35945a2d056956f8334f977b25fb2bR29 reference commit which I think may be useful in tracking this down

PR #3907, opened 2020-04-01
perf(posts): save some of the multiple db queries performed to fetch …

tuxpiper:
…geojson from posts

- pre-fetches point and geometry values along with the mean post search query
- avoids queries into tables other than Post if their results are not going to be needed (as is the case for the geojson controller)

Test checklist:
- [ ] Perform usual map, data and post browsing operations

Fixes ushahidi/platform# .

Ping @ushahidi/platform


tuxpiper:
This is the first batch of database query optimizations for very popular requests to Platform API , such as fetching GeoJSON.

There's still a pretty obvious optimization pending to implement, in order to avoid repeatedly querying for some form table attributes. Leaving that for another PR as I'm looking into clean (and safe) ways to store that information in memory for the duration of the request/job.

app/Http/Controllers/API/Posts/GeoJSONController.php
```diff
@@ -40,7 +40,8 @@ public function prepBoundingBox(Request $request)
     protected function getFilters(Request $request)
     {
         return parent::getFilters($request) + [
-            'include_types' => ['point', 'geometry']
+            'include_types' => ['point', 'geometry'],
+            'output_core_post' => true
```
rowasc, line 44: may be good to add a comment on this new filter's meaning

src/App/Repository/PostRepository.php
```diff
@@ -106,6 +128,16 @@ protected function getTable()
...
     }
 
+    protected function fetchFormAttributes($form_ids = [])
+    {
+        $form_ids = is_array($form_ids) ? $form_ids : [ $form_ids ];
+
+        $attrs = $this->form_repo->getAllFormStagesAttributes($form_ids);
+
+        $this->form_attributes_by_form = $attrs->groupBy('form_id');
+        $this->form_attributes_by_key = $attrs->keyBy('key');
```
rowasc, line 138: this looks like a typo maybe? (keyBy)

tuxpiper: ah no, that's a Collection method , it works :+1:

app/Http/Controllers/API/Posts/GeoJSONController.php
tuxpiper: I added it here: 

https://github.com/ushahidi/platform/pull/3907/files#diff-15be46c5155c0769e0fb9dd020f43de8R493

since that's where all the filters are anyway... Shall I duplicate it?

src/App/Repository/PostRepository.php
tuxpiper: That takes the form attribute queries result and creates a new collection indexed by attribute key

app/Http/Controllers/API/Posts/GeoJSONController.php
rowasc: oh, I think that's fine then. Probably something we should have in a static const but not a worry.

src/App/Repository/PostRepository.php
rowasc: oh, fancy. Ok !

tuxpiper:
I think this should be good for merging and QA now

rowasc:
@tuxpiper sorry I assumed we were already testing :( Yes let's push to steve-buscemi if we haven't and get Walter to do some QA so we can push this and the cache changes to prod this week.

rowasc:
@Obadha2 can you test this ?   (this should be the highest priority to test)

Obadha2:
@rowasc I found this one issue [logged here](https://github.com/ushahidi/platform/issues/3922) 
But after consulting with @tuxpiper agreed this wasn't a blocker.

PR #3911, opened 2020-04-03
test: gateway checker second try

individual-it:
- added tests for gateway checker

this is the second try, because the first broke CI on codeship and had to be reverted see https://github.com/ushahidi/platform/pull/3880#pullrequestreview-386710771
`DOCUMENT_ROOT` & `SCRIPT_FILENAME` are now looking out for `httpdocs` and not `platform`. That should not make any big difference for the quality of the tests, we just want to make sure there is something sensible in those fields, and that way it should pass on more different platforms

Test checklist:
- [x] :robot:

Fixes ushahidi/platform#3672 .

Ping @ushahidi/platform


individual-it:
CC @tuxpiper @rowasc

Issue #3543, opened 2019-06-02
Posts are successful even when there's no correspoding form.


Timtech4u:
**Describe the bug** I am able to make posts to the `api/v3/posts` endpoint even when there's no corresponding form.

**Where was the bug observed** Bug was found on the _api/v3/posts_ endpoint, I made a post request using the API without having any survey(forms) and it was successful.

**To Reproduce** Steps to reproduce the behavior: 1. Make a POST request to the following endpoint `<domain_url>/api/v3/posts` 2. Send an empty request body, or send an invalid character with the request (ie use “ quotation marks instead of " in your JSON) 3. Check the _Data_ on your Ushahid Platform, you'll see the submitted posts as mine below:

**Screenshots** [https://user-images.githubusercontent.com/11256481/58759588-51d8f280-8524-11e9-93d6-43bae49216dc.png](https://user-images.githubusercontent.com/11256481/58759588-51d8f280-8524-11e9-93d6-43bae49216dc.png)

Aha! Link: https://ushahiditeam.aha.io/features/PROD-223

AriannaLanz:
Looking at this one!

rowasc:
HA. I was just trying to repro this in case you needed help... *fun fact*:  This can only be reproduced by either sending an invalid character (notice the difference between " and “ ) OR sending an empty body in the request. It's not related to the form ID itself but to  fact that the validation is being skipped/ passing essentially if the body is empty (!!!!!)

Issue #3913, opened 2020-04-06
Not So Good API Request validation


anje123:
**Describe the bug**
Based on the API Documentation https://docs.ushahidi.com/platform-developer-documentation/tech-stack/api-documentation, the datatype like string, numeric, object where given but the  Ushahidi platform validation does not accurately follow it, in a case where when I send a post request to the
/api/v3/posts endpoint, I can insert a numeric value in a field that requires a string value, without no  error response sent back 

**Where was the bug observed**
All Endpoint that requires an input value e.g /api/v3/posts 
**To Reproduce**
1. Go to ushahidi platform API docs
1. get authenticated by using the bearer auth  token 
2. send an invalid post request to /api/v3/posts endpoint, using POSTMAN as suggested previously

**Screenshoot**
![Screenshot from 2020-04-06 11-17-06](https://user-images.githubusercontent.com/31329500/78572504-77f3b100-781f-11ea-92bb-aa4b665220cc.png)

**Expected behavior**
you get a 200 ok response when you are supposed to get a 400 or 422 error response

**Is there a workaround? What is it.**
After some checking, I noticed the Ushahidi platform uses the ushahidi/kohana-validation which does not have a function that handles does validations, so I think the API validation is still a little bit vulnerable

**Desktop Hardware Details (please complete the following information):**
 - OS: ubuntu 18.04
 - Browser postman

Aha! Link: https://ushahiditeam.aha.io/features/PROD-134

tuxpiper:
Thank you for the detailed report @anje123 !

We will be handling some of these with https://github.com/ushahidi/platform/pull/3811 , using this issue as anchor .

I think we will be seeing general improvements to validations as we move on to the next major version of API, which will drop kohana for good. .

@all-contributors please add @anje123 for bug

PR #3914, opened 2020-04-06
Update the Post-Create Validation Class

anje123:
This is to solve issue #3543 which is been able to make posts to the api/v3/posts endpoint even when there's no corresponding form , but I think there more to this issue please kindly check out this issue #3913

- throw an error if required fields are not entered.
- does not save null values to the ushahidi database

Test checklist:
- [ ]

Fixes ushahidi/platform#3543.

Ping @ushahidi/platform


tuxpiper:
Dropping in favor of previously existing #3811 since they have the same changes

PR #3916, opened 2020-04-07
Implement video upload in platform

RachealN:
This pull request implements the following below
- [x] Migrations, model and factories
- [ ] Controller
- create/upload a video
- edit and update
- Delete a video
- Preview the video before uploading
- Restore the video in case it is deleted accidentally
- [ ] Unit tests for the above functionality


migrations/2020_04_06_180910_create_video_uploads_table.php
```diff
@@ -0,0 +1,38 @@
...
+class CreateVideoUploadsTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        if(! Schema::hasTable('video_uploads')) {
```
rowasc, line 16: I'd probably leans towards using `media` rather than a new table.

rowasc:
Some general notes here and mostly questions
- I'd generally prefer that we enable integrations with providers that do video well, with good secure APIs, rather than save them in our own services and have to take care of it. 
- Vimeo had issues regarding API keys and security. Is that the same if we look at new versions of their API, and is that something that we can instead support through Youtube and other providers? 
- What's the usecase for uploading directly to our servers? Is it possible to integrate instead with youtube / vimeo/ others and upload through an API?

migrations/2020_04_06_180910_create_video_uploads_table.php
RachealN: well, if we are to use AWS S3 services for hosting our videos, That means we won't be needing this, tables

tuxpiper:
There was a question in gitter about focusing on AWS S3 as the storage backend. Here's my reply:

@RachealN 

It’s definitely a solution, it would work for the deployments we maintain at Ushahidi right now. And ultimately of course you’ve got the choice of what you’d implement .

However, I think it would be more useful for the community as a whole not to be tied in to a single vendor for a functionality (which is kind of the problem we have with Vimeo / Youtube right now) . 

Would you like to look at using an abstraction layer like flysystem ( https://flysystem.thephpleague.com/v1/docs/ ) ?

RachealN:
> Thanks @RachealN for this iniciative.
> Some general notes here and mostly questions for you and @tuxpiper
> 
> * I'd generally prefer that we enable integrations with providers that do video well, with good secure APIs, rather than save them in our own services and have to take care of it. But I'm curious about the usecase for us saving them.
> * Vimeo had issues regarding API keys and security _the way we used to implement it_ if I'm not mistaken. Is that the same if we look at new versions of their API, and is that something that we can instead support through Youtube and other providers?
> * What's the usecase for uploading directly to our servers? Is it possible to integrate instead with youtube / vimeo/ others and upload through an API?

RachealN:
> There was a question in gitter about focusing on AWS S3 as the storage backend. Here's my reply:
> 
> @RachealN
> 
> It’s definitely a solution, it would work for the deployments we maintain at Ushahidi right now. And ultimately of course you’ve got the choice of what you’d implement .
> 
> However, I think it would be more useful for the community as a whole not to be tied in to a single vendor for a functionality (which is kind of the problem we have with Vimeo / Youtube right now) .
> 
> Would you like to look at using an abstraction layer like flysystem ( https://flysystem.thephpleague.com/v1/docs/ ) ?

@tuxpiper and @rowasc , Thank you for your feedback
I will go ahead and implement video uploads using AWS S3 and in order not to be tied on a single vendor for functionality like as you have mentioned, I will follow the guidelines specified in flysystem documentation you have shared above. Thank you for your guidance and I keep you posted on any progress

RachealN:
@tuxpiper @rowasc @Angamanga 
am working on implementing video uploads using AWS S3 as we agreed, but I am having a challenge. I tried to sign up for the free tier package(12months free) on AWS but my credit card was rejected for some reason besides I was just going to use my account for testing purposes.
I am kindly requesting if you can please create the Ushahidi account on AWS and give me credentials. that's when I will be able to host videos on S3
Looking forward to your help and if it is not possible still let me know. Thanks

tuxpiper:
@RachealN we will discuss this internally to see if we could somehow do it.

In the mean time, have you checked minio? It's often used for developing code that uses the AWS S3 API , without having to rely / setup an AWS service.

RachealN:
> @RachealN we will discuss this internally to see if we could somehow do it.
> 
> In the mean time, have you checked minio? It's often used for developing code that uses the AWS S3 API , without having to rely / setup an AWS service.

Thanks for your feedback, let me checkout Minio Documentation and see if it can work for us. I will keep you posted in case of any progress

PR #3917, opened 2020-04-07
fix(perf): cache media storage url

tuxpiper:
- cache media files URL resolution , which can be an expensive operation when cloud storage is involved

Test checklist:
- [ ] verify several times that photos are available in posts

Fixes ushahidi/platform# .

Ping @ushahidi/platform


rowasc:
@Obadha2 can we have this tested today? 
Thanks!
 cc @tuxpiper

rowasc:
Tried this in steve, it looks great 🎉

PR #3925, opened 2020-04-09
fix(Twitter): urls are now formatted with the std https://twitter.com…

rowasc:
…/{userid}/status/{tweetId}

- This change was necessary because the old /statuses path stopped working at some point, which means all the twitter urls we grabbed from that point onwards were broken. Not super great
- There doesn't seem to be any way to get the correct URL from the API response we get from /search API, which is very dissapointing because if the format changes again this will stop working. The only thing that saves us here is that because this is a more user facing url, its likely that it will be redirected even if it changes, rather than just stop working

- Changes twitter urls to be in the format /user/{twitterUserId}/status/{twitterStatusId}. Migrates old twitter URLs as well as generates new ones in this format

Test checklist
# BEFORE you merge this PR, ensure that you
- [x] Have a deployment with tweets
- [x] Check that the tweet's link doesn't work (ie it's a 404)

# After you merge this PR and it lands in steve
- [ ] Check some old tweets and verify the links work now
- [ ] Get new tweets from the twitter datasource, the links should work 

@Obadha2 this is important for context see above the checklist 

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/migrations/20200409232153_rewrite_twitter_urls.php
+++ b/migrations/20200409232153_rewrite_twitter_urls.php
@@ -27,18 +27,21 @@ class RewriteTwitterUrls extends AbstractMigration
      */
     public function up()
     {
-        $sql = "UPDATE messages 
-                INNER JOIN contacts on contacts.id = messages.contact_id
-                SET messages.message = REPLACE(messages.message, concat('https://twitter.com/statuses/', messages.data_source_message_id), concat('https://twitter.com/', contacts.contact, '/status/', messages.data_source_message_id))
-                WHERE `messages`.`type` = 'twitter'";
+        $sql = "UPDATE messages INNER JOIN contacts on contacts.id = messages.contact_id " .
+                "SET messages.message = " .
+                "REPLACE(messages.message, concat('https://twitter.com/statuses/', messages.data_source_message_id), " .
+                "concat('https://twitter.com/', contacts.contact, '/status/', messages.data_source_message_id)) " .
+                "WHERE `messages`.`type` = 'twitter'";
         $this->execute($sql);
     }
 
     public function down() {
-        $sql = "UPDATE messages 
-                INNER JOIN contacts on contacts.id = messages.contact_id
-                SET messages.message = REPLACE(messages.message, concat('https://twitter.com/', contacts.contact, '/status/', messages.data_source_message_id), concat('https://twitter.com/statuses/', messages.data_source_message_id))
-                WHERE `messages`.`type` = 'twitter'";
+        // phpcs:ignore
+        $sql = "UPDATE messages INNER JOIN contacts on contacts.id = messages.contact_id " .
+                "SET messages.message = REPLACE(messages.message, " .
+                "concat('https://twitter.com/', contacts.contact, '/status/', messages.data_source_message_id), " .
+                "concat('https://twitter.com/statuses/', messages.data_source_message_id)) " .
+                "WHERE `messages`.`type` = 'twitter'";
         $this->execute($sql);
     }
 }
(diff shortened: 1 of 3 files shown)


migrations/20200409232153_rewrite_twitter_urls.php
```diff
@@ -0,0 +1,38 @@
...
+     *    renameColumn
+     *    addIndex
+     *    addForeignKey
+     *
+     * Remember to call "create()" or "update()" and NOT "save()" when working
+     * with the Table class.
+     */
+    public function change()
+    {
+        $sql = "UPDATE messages
```
rowasc, line 30: my big fear with this change is that it'll be pretty heavy in some deployments... like my ridiculous testing one with like 300k tweeets :/

src/App/DataSource/Twitter/Twitter.php
```diff
@@ -187,13 +187,20 @@ public function fetch($limit = false)
...
+                    'from' => $user_id,
                     'to' => null,
-                    'message' => 'https://twitter.com/statuses/' . $id,
+                    /** 
+                     * Best compromise I could find was just make the proper urls with user_id rather than only 
+                     * tweet id (for which there is an unofficial formula)... since there doesn't seem to be a way to grab the URL from the 
+                     * API itself in the v1.1 search endpoint. 
+                     * Fun fact: if the user id is wrong, twitter still takes you to the correct Tweet... they just use the tweet id
+                    **/
+                    'message' => "https://twitter.com/$user_id/status/$id",
```
rowasc, line 203: couldn't find a way to get twitter urls... so I'm using this more obvious user facing URL instead of /statuses... which I suppose gives me more peace of mind... but still, would prefer if twitter actually gave us the url to use, you know?

rowasc:
@tuxpiper this is medium-gross, but there wasn't a good way to get the URL so I suppose is the best we can do right now (unless we figure out a better way ? lol). Not super happy about the fact that this may become broken if twitter decides to deprecate this url format without redirects like they did for /statuses tho

migrations/20200409232153_rewrite_twitter_urls.php
```diff
@@ -0,0 +1,44 @@
...
+     *    addForeignKey
+     *
+     * Remember to call "create()" or "update()" and NOT "save()" when working
+     * with the Table class.
+     */
+    public function up()
+    {
+        $sql = "UPDATE messages 
+                INNER JOIN contacts on contacts.id = messages.contact_id
+                SET messages.message = REPLACE(messages.message, concat('https://twitter.com/statuses/', messages.data_source_message_id), concat('https://twitter.com/', contacts.contact, '/status/', messages.data_source_message_id))
```
tuxpiper, line 32: I thought this couldn't be done, because when I was checking a database with tweet messages, I couldn't see the tweet user handles anywhere. Turns out that we are storing the numeric ID for the user.

It also turns out that twitter *at this point in time now* will accept that numeric ID in the tweet URL  (i.e. https://twitter.com/:userNumId/status/:tweetId ) and redirect to the URL with the user handle.

So that's good now, it makes me queasy though 🤷  Should we start storing the screen name, should we start storing both? Big questions for a Friday

src/App/DataSource/Twitter/Twitter.php
tuxpiper: I'd so much love that too!

Obadha2:
@rowasc only new tweets coming in have working links.
Old ones before I merged remain broken.

rowasc:
@Obadha2 oh that's unfortunate :/ Checking

rowasc:
@Obadha2 which deployment did you use? just so I go into it

Obadha2:
https://test-deployment.steve-buscemi.ush.zone/views/data @rowasc

Issue #3686, opened 2019-09-03
Facebook: How users want to pull in reports from Facebook


Erioldoesdesign:
**Which user group(s) are the primary audience for this feature/enhancement.**

https://bristolstreetharassmentproject.ushahidi.io
https://www.bristolwomensvoice.org.uk/

These folks do work around street harassment and stopping hate crime in Bristol in the UK.

Specify key use cases: They are a team of program/data workers and advocacy/storytelling. They also work closely with local authority and council to affect policy change and physical changes around the city.

e.g. They received reporst around harassment in a particular green space and they used platform reports to help get more police presence and street lighting put in that specific area.

**Is your feature request related to a problem? Please describe.**
They currently collect posts via survey responses but are interested in both Twitter hashtag and email data sources. 
Through these conversations they talked about their activity on facebook:
https://www.facebook.com/BristolWomensVoice/
https://www.facebook.com/BristolZeroTolerance/
https://www.facebook.com/events/487004545403788/

They mentioned that facebook is very valuable as an engagement channel for them and they receive comments on posts as well as facebook messages of people reporting street harassment toward sthem

**Describe the solution you'd like**
This kind of deployment user would greatly benefit from further work on facebook messenger bots that report directly to the deployments and/or a way of identifying a specific facebook post's comments as potential reports.

**Describe alternatives you've considered**
A conversation was had with the deployment owner around 'manual entry of posts/data' I advise that they could add these into the deployment themselves but the key signifier was levels of effort and time spent manually transferring reports to the deployment.

Automation is key here.

**Additional context**
TBC I'll try to access a potential facebook report via their channels.

Aha! Link: https://ushahiditeam.aha.io/features/PROD-180

rowasc:
Thanks Eriol, this is great info to have.
@Erioldoesdesign @Angamanga wondering if the FB Bot prototype could work for them at least to get them started (if they have a server to set it up ) ? I know it had a lot of issues but it'd be better than manually copying, I think? 
I'd like to find a way to fund this using the learnings from Anna's FB Bot and any more user research we can get to improve upon it since it really sounds like a feature we should have :/

Angamanga:
@rowasc @Erioldoesdesign If they have a server to set it up + then it needs to be approved by Facebook. Usually not that complicated, the Uchaguzi-one got approved within a couple of days. But I would suggest we look into the version of the fb-api first, the one we use is pretty old and that might complicate the approval-process...

rowasc:
[...]

PR #3941, opened 2020-04-19
initial commit for adding FB messenger Bot

rohit645:
-Initial configuration for a Facebook Messenger Bot. 

Feature addition in reference to ushahidi/platform#3686 .

Ping @ushahidi/platform

### DIFF
--- a/app/Http/Controllers/API/Bot/BotController.php
+++ b/app/Http/Controllers/API/Bot/BotController.php
@@ -15,7 +15,6 @@ class BotController extends Controller
 
     public function index()
     {
-           
     }
     //
 }
--- a/app/Http/Controllers/API/Bot/BotController.php
+++ b/app/Http/Controllers/API/Bot/BotController.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace Ushahidi\App\Http\Controllers\API\Bot;
+
+use Illuminate\Http\Request;
+use Ushahidi\App\Http\Controllers\Controller;
+
+class BotController extends Controller
+{
+    /**
+     * Create a new controller instance.
+     *
+     * @return void
+     */
+
+    public function index()
+    {
+           
+    }
+    //
+}
--- a/app/Http/Middleware/VerifyBot.php
+++ b/app/Http/Middleware/VerifyBot.php
@@ -1,0 +1,23 @@
+<?php
+
+namespace Ushahidi\App\Http\Middleware;
+
+use Closure;
+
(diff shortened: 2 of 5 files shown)


rowasc:
@rohit645 thanks for your initial config PR. As early feedback: can you feature flag this so that we have the ability to do soft launches safely once the FB Bot is integrated?

rohit645:
> @rohit645 thanks for your initial config PR. As early feedback: can you feature flag this so that we have the ability to do soft launches safely once the FB Bot is integrated?

@rowasc Sorry, I didn't get what you wanted to say. Can you please elaborate? I don't have idea of how feature flag works?

rowasc:
@rohit645 sorry, my bad I should have expanded! 
So we use feature flags when we want to have the ability to deploy without releasing a feature. What I mean by this is that for instance, we could eventually merge the FB Bot code under a "disabled" state that only those who want it would change to enabled. 
To check if a feature is enabled, you can use something like:
`Features::isEnabled('fbbot')`  
before enabling the routes etc. 

Example we used to enable HXL (for HDX ) just for one deployment (HDX integration is the last "big" feature we released after a period of it existing but being only available for select deployments ) . 

`Features::isEnabled('hxl')` would give you the state of enabled/disabled of the HXL feature for instance. 
In the route you could add `feature:hxl` (or `feature:fbbot` )  as the middleware as well, in order to enable or disable the FBBot's routes on demand. 

and then in the client you'd check with something like ` !Features.isFeatureEnabled('hxl')`  if you wanted to show a part of the UI conditionally (just an example). 

Not sure if this helps at all. I should probably write a better guide. I can invest a bit more on this type of doc closer to end of week if that'd help since it's not documented

rowasc:
been thinking about this
@tuxpiper it's not clear to me that we will want to merge this *into* platform itself. I would prefer a chatbot as a separate repository, with the code here only being related to actually connecting the thing, much like Anna's bot was separate and like Tim's USSD setup (even better), because it alligns much better with our vision for the future as interconnected services. Thoughts?

tuxpiper:
> been thinking about this
> @tuxpiper it's not clear to me that we will want to merge this _into_ platform itself. I would prefer a chatbot as a separate repository, with the code here only being related to actually connecting the thing, much like Anna's bot was separate and like Tim's USSD setup (even better), because it alligns much better with our vision for the future as interconnected services. Thoughts?

agreed, I also think it would be much better in terms of architecture :+1:

PR #3953, opened 2020-04-22
fix(Twitter): urls are now formatted correctly for posts and messages

rowasc:
Using the format https://twitter.com/{userid}/status/{tweetId}

- This change was necessary because the old /statuses path stopped working at some point, which means all the twitter urls we grabbed from that point onwards were broken. Not super great
- There doesn't seem to be any way to get the correct URL from the API response we get from /search API, which is very dissapointing because if the format changes again this will stop working. The only thing that saves us here is that because this is a more user facing url, its likely that it will be redirected even if it changes, rather than just stop working
- this PR introduces an extra fix basically over my last PR which is that it updates the posts too when they match the message.... 

- Changes twitter urls to be in the format /user/{twitterUserId}/status/{twitterStatusId}. Migrates old twitter URLs as well as generates new ones in this format

Test checklist
# BEFORE you merge this PR, ensure that you
- [x] Have a deployment with tweets
- [x] Check that the tweet's link doesn't work (ie it's a 404)

# After you merge this PR and it lands in steve
- [ ] Check some old tweets and verify the links work now
- [x] Get new tweets from the twitter datasource, the links should work 

Fixes ushahidi/platform# .

Ping @ushahidi/platform


rowasc:
come on little pull request you can do it. 

Anyway, this passes for me locally. Tested up/down migration

rowasc:
@Obadha2 please check. thanks!

Obadha2:
@rowasc this still fails QA. It seems the merge undid previous changes. New tweets coming in have the link now looking like: 
/statuses/{twitterStatusId}
Screenshot 2020-05-04 at 11 39 18 AM

rowasc:
@Obadha2 that's odd, maybe something with steve :thinking: Will check what's up with the server since it passed everything locally .

rowasc:
agh I got it, it's becuase I forgot to revert your revert , so develop doesn't have the code for it

PR #3985, opened 2020-06-15
fix(csv): display full media URLs in the CSV . Move special filesyste…

tuxpiper:
…m URL logic to service provider

- uses Storage façade in CSV exporter to obtain the full media URL . Up until now we were just providing the path suffix ( i.e. "5/a/b/cheese-loncha-on-cat.gif" ) , which was not open-able
- moves specific logic that handles Rackspace files quirks off the formatter and into a class overriding FilesystemManager behavior

Test checklist:
- [ ] Ensure that attached photos still display properly
- [ ] Ensure that exported CSVs contain full URLs for the media ( i.e. "https://... " , instead of the path suffix only)

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -119,7 +119,7 @@ class AppServiceProvider extends ServiceProvider
         $this->app->singleton('filesystem', function ($app) {
             return $app->loadComponent(
                 'filesystems',
-                \Illuminate\Filesystem\FilesystemServiceProvider::class,
+                \Ushahidi\App\Providers\FilesystemServiceProvider::class,
                 'filesystem'
             );
         });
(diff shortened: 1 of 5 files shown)


tuxpiper:
@rowasc related to issue filed in intercom (cc'd you in there as well)

rowasc:
Thanks David!.
I had a feeling I fixed a similar issue like 6 months ago but apparently
only for the formatter :( .
Lmk how testing goes (you will need to switch the branch on steve or use
sad charlie, basically merge yourself and send to Walter). Was the
Cache::remember giving issues? I see it's removed

rowasc:
ah found the cache handler is just different nvm. Lgtm

PR #3991, opened 2020-07-01
Fix validation issue on post create and update

webong:
-

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,7 @@ modules/orm
 deployment.env
 docker.env
 vault.txt
+
+#Packages development
+/packages
+packages/*
--- a/.gitignore
+++ b/.gitignore
@@ -52,7 +52,3 @@ modules/orm
 deployment.env
 docker.env
 vault.txt
-
-#Packages development
-/packages
-packages/*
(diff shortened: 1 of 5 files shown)


rowasc:
@webong @tuxpiper can we get this into an environment and assigned to @Obadha2

tuxpiper:
> @webong @tuxpiper can we get this into an environment and assigned to @Obadha2

with platform-client branch = develop , correct @rowasc  ?

tuxpiper:
@Obadha2  deployed on https://pr3991.staging.ush.zone , with support user credentials as in Lastpass

tuxpiper:
@rowasc  @webong  could you please let us know what is the issue that this fixes?

We need a description of what to test for doing the QA

tuxpiper:
Tested and definitely an improvement :+1:

It's still possible to post something without setting the `form_id` , which I'm not sure if it's something that should be allowed. There seems to be some graceful handling on the client by later allowing the end user to assign the post to the form... but I could still do some very wacky stuff like creating posts with attributes belonging to different forms. That case would cause an unrecoverable error for the user on the client.

Anyway, one issue at a time

tuxpiper:
This fixes #3543 , supersedes PRs #3811 #3914

PR #4037, opened 2020-08-12
Update better translation

Solymanx7:
- Changes word "دورك" to better representing word of "role/s" which is "صلاحيات"

Aha! Link: https://ushahiditeam.aha.io/features/PROD-61


Solymanx7:
[clabot:check]

rowasc (changes requested):
Sorry that it wasn't clear to begin with, but the ultimate source for translations is actually at transifex.com . Would it be feasible for you at all to register and make the changes in there? Here's the URL for our project: https://www.transifex.com/ushahidi/ushahidi-v3-api
Thank you for your help!

PR #4042, opened 2020-08-13
Update tag.php

Solymanx7:
- Translated the remaining English sentences into Arabic

Aha! Link: https://ushahiditeam.aha.io/features/PROD-66

### DIFF
--- a/resources/lang/ar/tag.php
+++ b/resources/lang/ar/tag.php
@@ -1,11 +1,11 @@
 <?php
 
 return array(
-    'isSlugAvailable' => ':field :value is already in use',
+    'isSlugAvailable' => ':field :value مستخدم بالفعل',
     'role' => array(
         'isRoleValid' => 'يجب أن يطابق الدور الفئة الأساسية',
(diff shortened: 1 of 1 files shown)


tuxpiper:
Hi @Solymanx7 , thank you so much for these contributions!

Sorry that it wasn't clear to begin with, but the ultimate source for translations is actually at transifex.com  . Would it be feasible for you at all to register and make the changes in there? Here's the URL for our project: https://www.transifex.com/ushahidi/ushahidi-v3-api

rowasc (changes requested):
Sorry that it wasn't clear to begin with, but the ultimate source for translations is actually at transifex.com . Would it be feasible for you at all to register and make the changes in there? Here's the URL for our project: https://www.transifex.com/ushahidi/ushahidi-v3-api

tuxpiper:
Actually @rowasc , we have a bigger issue with API translations, please see my comment here: https://github.com/ushahidi/platform/issues/4069#issuecomment-707627216

PR #4056, opened 2020-08-25
fix: geolocation locale

leninpaulino:
- Fix the locale param, use the app locale if not sent 

Test checklist:

Fixes ushahidi/platform# .

Ping @ushahidi/platform


tuxpiper:
Makes sense to me :+1:

PR #4104, opened 2020-10-13
feat(cache-control): add middleware for setting cache-control header [WIP]

tuxpiper:
[WIP]: haven't even tested this yet. Early sharing for checking if this is in the right direction

- adds middleware for setting cache-control headers
- uses such middleware in the geojson controller

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/Http/Middleware/CacheControl.php
+++ b/app/Http/Middleware/CacheControl.php
@@ -1,0 +1,82 @@
+<?php
+
+namespace Ushahidi\App\Http\Middleware;
+
+use Closure;
+use Illuminate\Contracts\Auth\Factory as Auth;
+
+class CacheControl
+{
+    /**
+     * The authentication guard factory instance.
+     *
+     * @var \Illuminate\Contracts\Auth\Factory
+     */
+    protected $auth;
+
+    // TODO: make these configurable without modifying source code
+    private $policies = [
+        'performance' => [ 'max-age' => 'max-age=600' ],
+        'conservative' => [ 'max-age' => 'max-age=10' ],
+        'forbid-caching' => [ 'cache' => 'no-store' ]
+    ];
+
+    /**
+     * Create a new middleware instance.
+     *
+     * @param  \Illuminate\Contracts\Auth\Factory  $auth
+     * @return void
+     */
+    public function __construct(Auth $auth)
+    {
+        $this->auth = $auth;
+    }
+
+    protected function isAuthorized(string $guard)
+    {
+        return !($this->auth->guard($guard)->guest());
+    }
+
+    /**
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @param  string $policy  selected general policy
+     * @param  string $exceptAuthGuard
+     *                don't apply to requests satisfying given authentication guard
+     * @param  string $privateIfAuthGuard
+     *                set cache visibility to private for requests satisfying given authentication guard
+     * @return mixed
+     */
+    public function handle(
+        $request,
+        Closure $next,
+        string $policy,
+        ?string $exceptAuthGuard = null,
+        ?string $privateIfAuthGuard = null
+    ) {
+
+        // Skip for certain authorized requests
+        if ($exceptAuthGuard != null && $this->isAuthorized($exceptAuthGuard)) {
+            return $next($request);
+        }
+
+        // Obtain main and visibility policies
+        $policy = $this->policies[$policy];
+        $viz_policy = [ 'visibility' => 'public' ];
+        if ($privateIfAuthGuard != null && $this->isAuthorized($exceptAuthGuard)) {
+            $viz_policy['visibility'] = 'private';
+        }
+
+        // Merge policies and obtain cache-control header values
+        $final_policy = array_merge($policy, $viz_policy);
+        $cache_control = implode(',', array_values($final_policy));
+
+        // Set the header on the response
+        $response = $next($request);
+        $response->headers->set('Cache-Control', $cache_control);
+
+        return $response;
+    }
+}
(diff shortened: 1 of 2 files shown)


app/Http/Middleware/CacheControl.php
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @param  string $policy  selected general policy
+     * @param  string $exceptAuthGuard
+     *                don't apply to requests satisfying given authentication guard
+     * @param  string $privateIfAuthGuard
```
rowasc, line 48: are we expecting a bool here? (or am I just confused)

rowasc (commented):
Hey so I think this is definitely the right direction if we are doing it ourselves. Laravel apparently added a middleware for this in later versions (so we're not there yet) so maybe we would switch in v5 to that

https://github.com/laravel/framework/blob/43bea00fd27c76c01fd009e46725a54885f4d2a5/src/Illuminate/Http/Middleware/SetCacheHeaders.php I think this is the implementation when from they first added it

app/Http/Middleware/CacheControl.php
tuxpiper: no, a string, it's just like the Auth middleware: 
https://github.com/ushahidi/platform/blob/develop/app/Http/Middleware/Authenticate.php#L33
as seen used here:
https://github.com/ushahidi/platform/blob/develop/routes/config.php#L17

Kind of a boolean really since I think we would always use "api" or leave empty, but I just didn't want to introduce another pattern of usage.

tuxpiper:
> Hey so I think this is definitely the right direction if we are doing it ourselves. Laravel apparently added a middleware for this in later versions (so we're not there yet) so maybe we would switch in v5 to that
> 
> https://github.com/laravel/framework/blob/43bea00fd27c76c01fd009e46725a54885f4d2a5/src/Illuminate/Http/Middleware/SetCacheHeaders.php I think this is the implementation when from they first added it

Ahh that's useful 🙏 , I will try to design a bit closer to it in terms of option passing . It doesn't do (yet) the distinction between logged in and guest users though .

tuxpiper:
@rowasc new, modified approach up. Most of the code is "borrowed" from "future" laravel ;)

app/Http/Middleware/SetCacheHeadersIfAuth.php
```diff
@@ -0,0 +1,100 @@
...
+     * @return \Symfony\Component\HttpFoundation\Response
+     *
+     * @throws \InvalidArgumentException
+     */
+    public function handle($request, Closure $next, $ifAuth = null, $options = [])
+    {
+        $response = $next($request);
+
+        if ($ifAuth != null) {
+            if ($this->auth->guard($ifAuth)->guest()) {
```
rowasc, line 45: we can probably combine/move up this condiditon to the first one no?

rowasc (approved):
nice. This looks good! I left a very small comment but keep in mind my brain is mush so maybe it's silly :grimace:
Are we testing it already? how is it doing? I saw something in slack yesterday

app/Http/Middleware/SetCacheHeadersIfAuth.php
tuxpiper: yep :+1:

tuxpiper:
Pushed another update with the hooks and stuff to make it a proper configurable thing... it ended up being a bit of a rewrite, so take a look pls :>

What I could not figure out is testing it. I tried activating caching by using Config::set() in the test pre-conditions, but it just went ignored :/  Do you have any idea about that by chance?

Apart from this, it seems to be working well 👍

tuxpiper:
@Obadha2 just added test checklist

rowasc:
@tuxpiper ok so. Because this (geojson.feature) is an integration test that basically only executes an http client against the API running on the background (plus things like adding database configs), I can't really think of a practical way to do this app config change "live" in the feature test. 

So a few actions I can think of from there that may help:
- Adding unit tests like we did in the SignatureAuthTest and the CheckDemoExpirationTest files would be a way to check that the middleware itself is doing what we expect it to do. That would probably be a bit more flexible too, to check all the variations easily, and the examples should be helpful. 

- We can test the default scenario through the HTTP integration test 

Also,  I'm not completely sure but you may want to rename the new `routes.php` config file. For me at least it didn't really pick up the configuration at all in the middleware when I was logging things from the tests until I renamed it, which maybe makes sense due to it being a sort of default config file ? Not sure tbqh (before I realized the actual problem , I thought the problem was that file ... he)

tuxpiper:
Thank you for taking a look @rowasc 

> Because this (geojson.feature) is an integration test that basically only executes an http client against the API running on the background (plus things like adding database configs), I can't really think of a practical way to do this app config change "live" in the feature test.

sad ... (actually, double sad, because I didn't think of this)

> Adding unit tests like we did in the SignatureAuthTest and the CheckDemoExpirationTest files would be a way to check that the middleware itself is doing what we expect it to do. That would probably be a bit more flexible too, to check all the variations easily, and the examples should be helpful.

yep that's the way, then 👍 

> Also, I'm not completely sure but you may want to rename the new routes.php config file. For me at least it didn't really pick up the configuration at all in the middleware when I was logging things from the tests until I renamed it, which maybe makes sense due to it being a sort of default config file ? Not sure tbqh (before I realized the actual problem , I thought the problem was that file ... he)

What did you rename it to?

rowasc:
>What did you rename it to?
"control.php" or something, I was just messing around didn't really think about the name much. Maybe something like cache-control would be a good name :/

Obadha2:
Ok this passes QA. 👍 
@rowasc @tuxpiper

tuxpiper:
Thank you @Obadha2 !

Since it seems to work as it is, I suggest we roll it and start using it for alleviating pressure in our production systems.

Issue #3977, opened 2020-05-28
API link fields curr, next, and prev lose filtering


bb:
**Describe the bug**

In the API endpoint `/api/v3/posts`, the response fields `curr`, `next` and `prev` lose filtering information from the query. E.g. when filtered for `status[]=draft` (i.e. not published), I expect the curr URL to keep this information. Also, the `next` URL should keep it, so I could e.g. write a client which could browse through the API by just looking at the 'next' field and proceed to that URL; e.g. for paging.

E.g. the output of https://quakemap.api.ushahidi.io/api/v3/posts?created_after_by_id=537&has_location=all&order=desc&order_unlocked_on_top=true&orderby=created&source%5B%5D=sms&source%5B%5D=twitter&source%5B%5D=web&source%5B%5D=email&status%5B%5D=draft contains no results (as I filtered only for drafts and the following item 538 is published). and has `curr":"https:\/\/quakemap.api.ushahidi.io\/api\/v3\/posts?orderby=created&order=desc&offset=0"`... however, https://quakemap.api.ushahidi.io/api/v3/posts?orderby=created&order=desc&offset=0 is a completely different result (seems to be everything).

**Where was the bug observed**

In Quakemap: 
https://quakemap.api.ushahidi.io/api/v3/posts?created_after_by_id=537&has_location=all&order=desc&order_unlocked_on_top=true&orderby=created&source%5B%5D=sms&source%5B%5D=twitter&source%5B%5D=web&source%5B%5D=email&status%5B%5D=draft

**To Reproduce**
1. Go to 'https://quakemap.api.ushahidi.io/api/v3/posts?created_after_by_id=537&has_location=all&order=desc&order_unlocked_on_top=true&orderby=created&source%5B%5D=sms&source%5B%5D=twitter&source%5B%5D=web&source%5B%5D=email&status%5B%5D=draft' (no need to login, click or scroll).
2. Observe `"count":0` right at the beginning.
3. Copy the URL from the `curr` field; in my case `https://quakemap.api.ushahidi.io/api/v3/posts?orderby=created&order=desc&offset=0` and open it in a new browser tab. Notice that all the slashes `/` are escaped by `\` which needs to be removed. _This might be yet another bug as escaping slashes is clearly not necessary in JSON._
4. Observe `"count":534 (or something different if data was added or removed in the meantime, but most likely not 0 as before).

**Expected behavior**

[...]

PR #4125, opened 2020-10-17
Fixes paging URLs of collections of Posts

AngadSethi:
- Adds a function `getSearchFilters` to the `SearchData` class which returns an array of search filters minus the sorting filters
- Modifies `CollectionFormatter` class wherein search filters are obtained from above function and merged with paging parameters

Test checklist:
- [ ] The paging URLs now contain search filters which are appended after the sort filters.
- [ ] If there are no search filters, none are added to the URL.
- [ ] Sorting filters and parameters are left untouched.

Fixes ushahidi/platform #3977


AngadSethi:
Signed.
[clabot:check]

Issue #3939, opened 2020-04-17
Limit the amount of posts that may be fetched with GET /api/v3/posts request


tuxpiper:
**Which user group(s) are the primary audience for this feature/enhancement.**
This was initially reported by a deployer with significant engagement. They handle personal data that is made available publicly, and it's not an option to hide it completely as sharing data is how that deployment works.

**Is your feature request related to a problem? Please describe.**
They were seeing how others where API-scraping their published posts for unknown purposes.

While they could rate limit the requests on an IP-address basis, these efforts were hampered by the fact that the requesting party could just ask for more posts per request.

**Describe the solution you'd like**

* Unprivileged requests should have a default maximum (i.e. 20) in how many posts they can fetch in one go
* Privileged requests should either be unlimited or have a different higher limit

What constitutes a privileged request? I think it would have to be either associated to a user permission ("Manage Posts"?) or to an API key permission/scope

This limit should **not** affect other endpoints that deal with post data, such as /api/v3/geojson . 

By nature those endpoints discard most of the post data and may even do further processing of the data with the effect of aggregating it and further reducing the amount of individual post detail exposed.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

Aha! Link: https://ushahiditeam.aha.io/features/PROD-123

rowasc:
regarding privileged: I think the way we do it for exports is through permissions     with DATA_IMPORT_EXPORT so that would fit the same pattern... altough it may be possible that a scope would have been the correct decision to be honest 🤔  
From slack chat
>I think last week my conclusion is that scopes generally would be a better fit when we are doing resource actions matched to scopes rather than this in particular (unless we consider it a separate action) because we would need to do some weird stuff with ?querystrings in our scope checking middleware to even handle it if we follow the ?limit= offset thing.. since we apply the middlewares in the routes (and you can't assign a middleware through checking querystrings in routes.php/whatever without some hacky hacks)

wino:
@tuxpiper @rowasc I took a crack at this one and have a forked PR here: https://github.com/ushahidi/platform/pull/4142
I went with the "Manage Posts" permission mentioned above for now. I think it could be switched to use scopes if needed, but I didn't have time today to figure it out in my first crack at this project.
I also tried to look into writing a test but didn't quite figure out how to handle the auth. I could give that another shot next week if needed.

tuxpiper:
This will need to be ported over to the v5 side

PR #4142, opened 2020-10-30
fix: limit the amount of posts that may be fetched with an unprivileg…

wino:
…ed GET /api/v3/posts request

-

Test checklist:
- [ ]

Fixes ushahidi/platform#3939 .

Ping @ushahidi/platform

Aha! Link: https://ushahiditeam.aha.io/features/PROD-891


app/Http/Controllers/API/Posts/PostsController.php
```diff
@@ -85,6 +85,7 @@ public function store(Request $request)
     public function index(Request $request)
     {
         $filters = $this->demoCheck($this->getFilters($request));
+        $filters['limitUnprivileged'] = 1;
```
tuxpiper, line 88: I think we would actually always want to have limits. It's really not a good practice not to have them. In that case, this activation here is probably unnecessary.

src/App/Repository/PostRepository.php
```diff
@@ -56,6 +56,8 @@ class PostRepository extends OhanzeeRepository implements
     // Provides `postPermissions`
     use InteractsWithPostPermissions;
 
+    const UNPRIVILEGED_REQUEST_LIMIT = 20;
```
tuxpiper, line 59: Given we would always have limits  in place, we would just need different limits for privileged vs unprivileged.

Also, it would be good to have them configurable through a file in `config/` (and environment variables)

src/App/Repository/PostRepository.php
```diff
@@ -389,6 +391,13 @@ public function setSearchParams(SearchData $search)
             $this->search_query->offset(intval($sorting['offset']));
         }
 
+        if ($search->getFilter('limitUnprivileged')) {
+            $isUnprivileged = !$this->postPermissions->canUserManagePosts($this->getUser());
```
tuxpiper, line 395: Suggest to add comment here how the post management permission is associated to bigger limits , and also that we would like to do that with API scopes in the future.

src/Core/Usecase/Post/SearchPost.php
```diff
@@ -29,7 +29,8 @@ protected function getPagingFields()
             'orderby' => 'post_date',
             'order'   => 'desc',
             'limit'   => null,
-            'offset'  => 0
+            'offset'  => 0,
+            'limitUnprivileged'  => 0
```
tuxpiper, line 33: Just as my comment above, I think we can dispense with this.

tuxpiper (changes requested):
Thank you @wino , please see my comments regarding how I think this feature should go.

Also, please take a look at the failing tests.

Cheers!

app/Http/Controllers/API/Posts/PostsController.php
wino: Ok I can do that. The way the issue was written it sounded like only non-privileged needed a limit.

src/App/Repository/PostRepository.php
wino: Good idea, will look into that.

app/Http/Controllers/API/Posts/PostsController.php
tuxpiper: > Ok I can do that. The way the issue was written it sounded like only non-privileged needed a limit.

Yes, that's right, that's pretty much how it was written.  I think it's just that we have learned a thing or two as we have seen more traffic spikes 😅

wino: @tuxpiper I totally get having real world results changing your reqs. HOwever, there was also this comment in the issue:
"This limit should not affect other endpoints that deal with post data, such as /api/v3/geojson"
So it seems like we still need some kind of activation from this point only?

src/App/Repository/PostRepository.php
wino: see above

tests/integration/acl.feature
```diff
@@ -196,7 +196,7 @@ Feature: API Access Control Layer
         When I request "/posts"
         Then the guzzle status code should be 200
         And the response is JSON
-        And the "count" property equals "13"
+        And the "count" property equals "12"
```
wino, line 199: I stared at these tests for a while and ran them manually via postman and am convinced that these updated numbers are now correct. What is happening is that without the new limit these queries pull 21+ results from the db and THEN SearchUsecase has a block of code around line 95 that "remove[s] any entities that cannot be seen" i.e. those that auth fails for the particular user. It just so happens that 1 of the "good" results is the 21st result that is left out by the limited query, so we end up with 1 less in the result. I think this serves to prove that the limiter for unprivileged requests is actually working.

src/App/Repository/PostRepository.php
wino: done

wino:
@tuxpiper I converted to config values for both privileged and unprivileged. I left in the filter param (slightly renamed) as the original issue only wanted this to apply for the index endpoint. I can still change this if you want but I wanted to be sure about that. I fixed the tests and commented about the fix as well.

app/Http/Controllers/API/Posts/PostsController.php
tuxpiper: Ah yea, it's definitely more of an endpoint dependent restriction, not something that should be driven from the model or usecase layer. For the GeoJSON endpoint there should also be limits (it's a very expensive query! once you get to 500+ posts) , but those could be different as the concerns are different.

tests/integration/acl.feature
tuxpiper: Ah right... that's a yucky thing of the current implementation. Makes sense 👍

config/posts.php
```diff
@@ -0,0 +1,10 @@
+<?php
+/**
+ * Posts Config
+ */
+
+return [
+    // Max number of posts that can be fetched for privileged/unprivileged requests
+    'privileged_request_limit' => 100,
```
tuxpiper, line 8: Could we make these accessible with environment variables? i.e. `API_POST_LIST_MAX_LIMIT` and `API_POST_LIST_ADMIN_MAX_LIMIT` ?

tests/integration/acl.feature
wino: Yeah I debated whether to try to fix it to have php do the limiting after the inaccessible records are removed but didn't want to blow up the status quo at this point. That seems like a future design consideration that likely exists across all the assets.

config/posts.php
wino: sure

app/Http/Controllers/API/Posts/PostsController.php
wino: I'll rename the config vars to allow for adding separate values for other endpoints in the future.

config/posts.php
wino: @tuxpiper env vars added

src/App/Repository/PostRepository.php
```diff
@@ -389,6 +389,23 @@ public function setSearchParams(SearchData $search)
             $this->search_query->offset(intval($sorting['offset']));
         }
 
+        if ($search->getFilter('limitPosts')) {
+            // the post management permission is associated with bigger limits
+            // we would like to do this with API scopes in the future
+            $isUnprivileged = !$this->postPermissions->canUserManagePosts($this->getUser());
+            $requestedLimit = $sorting['limit'] ?? 0;
+            if ($isUnprivileged) {
+                $unprivilegedLimit = env('API_POST_LIST_MAX_LIMIT', config('posts.list_max_limit'));
```
tuxpiper, line 398: the convention in Laravel/Lumen afaik is to put the env lookup in the config file, i.e. , as you can see in config/database.php:

```
'mysql' => [
            'host'      => env('DB_HOST', 'localhost'),
            'read' => [
...
```

could you please use that convention too?

tuxpiper (changes requested):
Just a final little thing about following a convention. Once that is done, will pass to QA. Thank you!

src/App/Repository/PostRepository.php
wino: will do when I get a chance

wino: @tuxpiper I made this change

wino: the tests are failing during the env setup stage with: Could not open input file: composer-setup.php
I tried merging master using the button on here and it is still the same error. I don't think that's related to my changes 🤷

tuxpiper (approved):
Looking good to me 👍  Thank you @wino !

Since this is a fairly significant change, @rowasc  would you like to take a peek before merging?

tuxpiper:
@Obadha2 this is good for QA .

Since it's not really a user facing feature, I think you can save it for the general smoke test, to check that nothing breaks. If looking for something more specific, I'd say checking map and data views work properly on filters with 100+ results.

tuxpiper:
We need to replicate this change on the v5 branch, still good to have on the v3 API . I'll create another PR merging this onto the v5 branches, and flagged as Work In Progress to bring the feature over to v5.

@Obadha2 merging this one shouldn't have any effect on the v5 code that we are releasing lately.

tuxpiper:
Opened https://github.com/ushahidi/platform/pull/4203 to start bringing this over to the v5 tree

Issue #4156, opened 2020-11-06
Support for WhatsApp as SMS through Twilio datasource


tuxpiper:
**Which user group(s) are the primary audience for this feature/enhancement.**
Deployers which may be able to obtain an activated WhatsApp account from Twilio.

**Is your feature request related to a problem? Please describe.**
No support for WhatsApp in Platform currently

**Describe the solution you'd like**
Twilio provides a WhatsApp gateway service that operates almost identically to their SMS service. Would like Platform to support this service just as it does SMS with Twilio.

It would be nice, but it's not required that these messages to be presented as "WhatsApp" in Platform, they may still be found under the "SMS" source.

**Describe alternatives you've considered**
Other providers, but this specific approach benefits from a specially low technical barrier.

**Additional context**
https://www.twilio.com/whatsapp

Aha! Link: https://ushahiditeam.aha.io/features/PROD-900

PR #4157, opened 2020-11-06
feat(twilio-whatsapp): properly mark Twilio messages originating from WhatsApp, for 2-way replies

tuxpiper:
- Adds WhatsApp contact type
- Passes contact type on to outgoing dataproviders
- Modifies Twilio data provider to be aware of messages coming from WhatsApp and ensure replies are sent through the same channel.

Test checklist:
- [ ] Sign up for Twilio's WhatsApp sandbox and add device to use it
- [ ] Set up Twilio datasource in Platform
- [ ] With the device, send message via WhatsApp to the sandbox phone number
- [ ] In the device, ensure the configured automatic reply is received
- [ ] As an admin in Platform, check that the message is received in Platform
- [ ] As an admin in Platform, send back a reply to the received message
- [ ] Check that the response message is received in the device

Fixes ushahidi/platform#4156 .

Ping @ushahidi/platform

### DIFF
--- a/src/App/DataSource/Twilio/TwilioController.php
+++ b/src/App/DataSource/Twilio/TwilioController.php
@@ -32,9 +32,26 @@ class TwilioController extends DataSourceController
             abort(403, 'Incorrect or missing AccountSid');
         }
 
+        // Detect contact type
+        $to = $request->input('To');
+        $from = $request->input('From');
+        if (strpos($from, ':')) {
+            $channel = (explode(':', $from))[0];
+            switch ($channel) {
+                case 'whatsapp':
+                    $contact_type = Contact::WHATSAPP;
+                    break;
+                default:
+                    $contact_type = Contact::PHONE;
+                    break;
+            }
+        } else {
+            $contact_type = Contact::PHONE;
+        }
+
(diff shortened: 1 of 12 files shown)


src/App/DataSource/Twilio/TwilioController.php
```diff
@@ -32,9 +32,26 @@ public function handleRequest(Request $request)
             abort(403, 'Incorrect or missing AccountSid');
         }
 
+        // Detect contact type
+        $to = $request->input('To');
+        $from = $request->input('From');
+        if (strpos($from, ':')) {
```
rowasc, line 38: I think if you do a strpos for whatsapp directly you can avoid the switch no?

rowasc (approved):
I left a small comment as optional since I made an assumption. LMK. This is so great. Thank you

src/App/DataSource/Twilio/TwilioController.php
tuxpiper: sure! I'd rather leave it though. My intention here was to leave it right open for other channel implementations, since it seems Twilio may be having more of those. So the first `i`f detects the presence of a channel marker and the `switch` will match the channel.

PR #4201, opened 2021-02-09
fix(performance): avoid some SQL queries with light request-level cac…

tuxpiper:
…hing and query grouping/preloading

- Reduces the amount of SQL queries needed while building post query responses.
- Introduces a request-context cache to the Resource classes that are in charge of rendering JSON responses. This is useful for entities that are repeated in a fully hydrated JSON response, such as categories
- Adds some lazy eager loading during JSON rendering and to relationship accessors as well

Test checklist:
- [ ] Post detail and post search endpoints on /api/v5 working correctly (and hopefully faster)

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/v5/Http/Controllers/PostController.php
+++ b/v5/Http/Controllers/PostController.php
@@ -45,7 +45,8 @@ class PostController extends V5Controller
      */
     public function show(int $id)
     {
-        $post = Post::find($id);
+        $post = Post::withPostValues()->where('id', $id)->first();
+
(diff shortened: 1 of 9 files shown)


tuxpiper:
This is a bit of a band-aid . Ultimately, small but frequently accessed information such as categories and survey definitions should be readily available in memory while these requests are being processed.

PR #4304, opened 2021-09-01
hotfix(flsms): add field for configuring the frontline server

tuxpiper:
- adds a configuration field for the frontline data source, so an alternative frontline server is configured. It is a required field when configuring, but deployments with the data source already configured and enabled will default to cloud.frontlinesms.com . 

Test checklist:
- [ ] Configure the data source and test message sending and receiving using frontline's phone simulator. 

Fixes ushahidi/platform#4305 .

Ping @ushahidi/platform


webong (approved):

tuxpiper:
Screenshot 2021-09-02 at 16 50 17

The new configuration screen with the added field.

PR #4394, opened 2022-01-25
Fix: Save post content values 

webong:
- To check if a field value is empty, in the case it's an array value. If empty, the process is skipped.

Ping @ushahidi/platform


tuxpiper:
LGTM! Merging

tuxpiper (approved):

PR #4395, opened 2022-01-28
Refactor: fix logic to check if a field input value exists

webong:
- Checks if a field input value exists and not if empty.

Ping @ushahidi/platform

### DIFF
--- a/v5/Http/Controllers/PostController.php
+++ b/v5/Http/Controllers/PostController.php
@@ -406,15 +406,22 @@ class PostController extends V5Controller
                 continue;
             }
             foreach ($stage['fields'] as $field) {
-                $type = $field['type'];
-                $value = is_array($field['value']) && array_key_exists('value', $field['value'])
-                    ? $field['value']['value']
-                    : $field['value']; // Fall back to the field value if "value" key doesn't exist
-                if (!isset($value) || (is_array($value) && empty($value))) {
+                if (!isset($field['value'])) {
                     continue;
                 }
+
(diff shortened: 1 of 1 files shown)


v5/Http/Controllers/PostController.php
```diff
@@ -406,16 +406,23 @@ protected function savePostValues(Post $post, array $post_content, int $post_id)
...
-                $type = $field['type'];
-                $value = is_array($field['value']) && array_key_exists('value', $field['value'])
-                    ? $field['value']['value']
-                    : $field['value']; // Fall back to the field value if "value" key doesn't exist
-                if (!isset($value) || (is_array($value) && empty($value))) {
+                if (!isset($field['value'])) {
+                    continue;
+                }
+
+                // We only want to check if a field input value is set, not if it's empty
```
tuxpiper, line 413: After checking with you @webong , this makes sense 👍  

So, if we PUT a post and there’s an attribute in the body with `"value": { "value": null }` , that means the attribute will not be touched (essentially behaving  a bit like a PATCH request)

But if it’s `"value": { "value": {} }`, that should clear the attribute's value.

My only comment is then that I think I would have understood the comment more quickly if I was given examples as coming from the API .. instead of remembering what PHP considers set or empty.

tuxpiper (approved):
Just left a fussy comment regarding a comment. Otherwise, it looks good to me.

tuxpiper:
Goes with ushahidi/platform-client#1760

PR #4407, opened 2022-04-01
feat: transient USSD-specific endpoint for post creation

tuxpiper:
- adds a POST /api/v5/post/_ussd endpoint , which works just the same as /api/v5/post but requires an additional `source_info` object with information about the USSD sender

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


tuxpiper:
@webong please take a look today. I'll be fixing the linting etc a bit later!

PR #4434, opened 2022-06-21
Migrate/laravel-rollup

webong:
- Move commands from src/Console/Command/ -> app/Console/Commands
- Update Service Providers
- Renamed src/Core/Tool -> src/Core/Tools
- Migrate from Lumen 5.5 -> Laravel 5.5
- Update routes files
- Update artisan file
- Update composer.json
- Refactor interfaces and traits to contracts and concerns respectively
- Modify v3 Usecases, Repositories, Formatters

Test checklist:
- [ ] Unit and Features tests are passing

Ping @ushahidi/platform


webong:
## Review
- [Here](https://github.com/ushahidi/platform/pull/4434/commits/05bc1b44b49f3812672c8d44d46030a697a0ccf0) moves all interfaces to the Contracts folder
- [Here](https://github.com/ushahidi/platform/pull/4434/commits/cf8c333b5d1c8f778afd90d8a8d1af705d266df7) moves all necessary traits to the Concerns folder
- [Here](https://github.com/ushahidi/platform/pull/4434/commits/fb9e9356c6efb38ea9e4baea979f268e13762f51) modifies the service providers.
- [Here](https://github.com/ushahidi/platform/pull/4434/commits/28bdf1e0235d178ae66358a474016bc10e3d5335) refactors the route files.

app/Http/Controllers/RESTController.php
```diff
@@ -227,31 +226,28 @@ abstract protected function getResource();
      */
     protected function getRouteParams(Request $request)
     {
-        return $request->route()[2];
```
tuxpiper, line 230: @webong @Mh-Asmi funny little change. What do you think happened here? Why did we have a hardcoded array index of [2]? 🤔 how is the change equivalent?

config/view.php
```diff
@@ -0,0 +1,33 @@
+<?php
```
tuxpiper, line 1: @webong @Mh-Asmi is this relevant/necessary for this project?

app/Http/Kernel.php
```diff
@@ -0,0 +1,75 @@
...
+            // \Illuminate\Cookie\Middleware\AddQueuedCookiesToResponse::class,
+            // \Illuminate\Session\Middleware\StartSession::class,
+            // \Illuminate\Session\Middleware\AuthenticateSession::class,
+            // \Illuminate\View\Middleware\ShareErrorsFromSession::class,
+            // \Ushahidi\App\Http\Middleware\VerifyCsrfToken::class,
+            // \Illuminate\Routing\Middleware\SubstituteBindings::class,
+        ],
+
+        'api' => [
+            'throttle:60,1',
```
Mh-Asmi, line 47: I have concerns about the value of the throttle "rate limit", and I think we need to raise it because this number is applied to all APIs together and not to each API alone @tuxpiper @webong

Mh-Asmi: For more information about the rate limit, it works on the public IP of the network, so if I exceed the rate limit while using my device, all other devices on the same private "wifi" network will also be blocked!

app/Http/Controllers/RESTController.php
tuxpiper: ahh interesting :) Thank you 🙏

app/Http/Kernel.php
webong: Oh really, I used to think the underlying of the rate-limiting middleware applies the counts per route. 

So with this information, do we take out the throttle middleware from the route group or do we just increase the number of request counts.

app/Http/Controllers/RESTController.php
```diff
@@ -310,7 +307,9 @@ protected function prepResponse(array $responsePayload = null, Request $request)
         if (! in_array($request->method(), $this->cacheableMethods)) {
             $response->headers->set('Cache-Control', 'no-cache, no-store, max-age=0, must-revalidate');
         }
-        $response->headers->set('Content-language', app('translator')->getLocale());
+
+        $response->headers->set('Content-language', Lang::getLocale());
```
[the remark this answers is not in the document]
webong: Yes it is.

config/view.php
webong: Since we don't make use of Blade for mailing, it's sort of not needed. Actually the view config file doesn't really do anything.

app/Tools/LumenMailer.php
```diff
@@ -4,16 +4,15 @@
  * Ushahidi Mailer
```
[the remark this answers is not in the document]
webong: Haha ... very true, I'll get it renamed to something less awkward.

Mh-Asmi:
Thank you @wisdom for your work,  I'm also good with these changes "Laravel and architecture".
I don't think we have any risks, I would suggest just changing the throttle value to at least 600.

tuxpiper:
@AmTryingMyBest marked this as "QA by Dev". By the time you read this, it may already be merged.

Took this path because it's critical to not keep this branch open any longer (too many changes) and because the assessed risk is low.

These changes are all about shuffling some files around, changing some files, and updating some libraries. There shouldn't be any changes at all to functionality. I guess at worst things would break due to some dangling reference. But that would probably raise a failure in the automated tests for this repo.

Ultimately, if there has been some oversight, hopefully, we would catch it with the smoke tests for the release.

Let me know if you have any concerns 🙏

Mh-Asmi (approved):

PR #4444, opened 2022-07-13
fix lock feature 

Mh-Asmi:
- in `get api/v3/posts`
	1- fix the problem of returning expired locks
	2- return the name of the owner of the lock 
	3- return if the lock is breakable for the current user

- in `put api/v3/posts/{id}/lock`	
	1- switch the owner of the lock if the current user can break the lock "has admin role"
	
- in  `get api/v5/posts`
	1- add active locks to post
	2- return the name of the owner of the lock 
	3- return if the lock is breakable for the current user

- in 	`update api/v5/posts/{id}`
	1- check if the post had a lock of another user before saving 
	2- release the lock after completing  saving post  

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


webong (approved):

webong:
@Mh-Asmi this looks all good to me, we can get it deployed for QA now.

PR #4448, opened 2022-07-19
feat: added addons for mteja and africastalking datasource

webong:
- Added new datasource addon
- Modified datasource manager to accept custom callback datasources

Ping @ushahidi/platform

### DIFF
--- a/src/Addons/Mteja/ShortMessageController.php
+++ b/src/Addons/Mteja/ShortMessageController.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace Ushahidi\Addons\Mteja;
+
+/**
+ * Mteja callback controller
+ *
+ * @author     Ushahidi Team <team@ushahidi.com>
+ * @package    Addons\Mteja
+ * @copyright  2022 Ushahidi
+ * @license    https://www.gnu.org/licenses/agpl-3.0.html GNU Affero General Public License Version 3 (AGPL3)
+ */
+
+use Illuminate\Http\Request;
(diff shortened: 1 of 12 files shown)


Mh-Asmi (approved):

Mh-Asmi:
it's good for me

src/Addons/Mteja/ShortMessageController.php
```diff
@@ -0,0 +1,50 @@
...
+
+        $this->save([
+            'type' => MessageType::SMS,
+            'from' => $from,
+            'contact_type' => Contact::PHONE,
+            'message' => $request->input('text'),
+            'to' => $request->input('to'),
+            'title' => null,
+            'datetime' => null,
+            'data_source_message_id' => "mteja-" . UUID::uuid4()->toString(),
```
tuxpiper, line 37: @webong I see in the mteja interaction logs that there is a messageId field in there. Is it possible we may get it in the callback from them as well?

webong: Really can't say, as I am hoping when I get a real response data from them, would be easier to know what value is accessible. 

As at now, some fields from their csv logs are missing in their api response payload as specified in their docs for a callback response, 

`  
{
  "to": "string",
  "from": "string",
  "text": "string",
  "tags": [],
  "sentimentScore": 0,
  "sentimentMagnitude": 0,
  "smsFlowId": 0,
  "questionText": "string",
  "fieldName": "string",
  "numericalData": 0
}
`

PR #4452, opened 2022-07-22
feat: add "Delete Posts" and "Delete Own Posts" permissions

Mh-Asmi:

- add new permissions "Delete posts" and "Delete their own posts"
- remove the ability to delete from permission "Manage Posts" 
- emove ability to delete from permission " Edit their own Posts"
- give grants "Delete Posts" permission to existing roles with "Manage Posts" permission
- give grants "Delete their own posts" permission to existing roles with "Edit their own posts" permission

Test checklist:
- [ ]

Fixes ushahidi/platform#4453 .

Ping @ushahidi/platform


webong (approved):
Looks good to me 👌🏽

src/Core/Tools/Authorizer/PostAuthorizer.php
```diff
@@ -80,7 +80,11 @@ public function isAllowed(Entity $entity, $privilege)
         }
 
         // First check whether there is a role with the right permissions
-        if ($this->acl->hasPermission($user, Permission::MANAGE_POSTS)) {
+        if (($privilege != "delete") && ($this->acl->hasPermission($user, Permission::MANAGE_POSTS))) {
```
kubabialy, line 83: Let's use strict comparison here. 
```suggestion
        if (($privilege !== "delete") && ($this->acl->hasPermission($user, Permission::MANAGE_POSTS))) {
```

src/Core/Tools/Authorizer/PostAuthorizer.php
```diff
@@ -139,13 +143,21 @@ public function isAllowed(Entity $entity, $privilege)
...
         if ($this->isUserOwner($entity, $user)
-            && in_array($privilege, ['update', 'delete', 'lock'])
+            && in_array($privilege, ['update', 'lock'])
             && $this->acl->hasPermission($user, Permission::EDIT_OWN_POSTS)) {
             return true;
         }
+        
+        // If the user is the owner of this post & they have delete own posts permission
+        // they are allowed to edit or delete the post.
+        if ($this->isUserOwner($entity, $user)
```
kubabialy, line 156: It's a small nitpick, but for readability sake I believe it would be beneficial to just move this if condition to separate method.

I was thinking about something like:

```php
private function canUserDelete(Model $entity, User $user, string $privilege): bool
{
    return $this->isUserOwner($entity, $user)
        && $privilege === 'delete'
        && $this->acl->hasPermission($user, Permission::DELETE_OWN_POSTS);
}
```

And here just check for:

```php
if ($this->canUserDelete($entity, $user, $privilege)) {
    return true;
}

File: v5/Policies/PostPolicy.php [hunk 3]
```diff
@@ -199,14 +203,22 @@ public function isAllowed($entity, $privilege)
...
         if ($this->isUserOwner($entity, $user)
-            && in_array($privilege, ['update', 'delete', 'lock'])
+            && in_array($privilege, ['update', 'lock'])
             && $authorizer->acl->hasPermission($user, Permission::EDIT_OWN_POSTS)) {
             return true;
         }
 
+        // If the user is the owner of this post & they have delete own posts permission
+        // they are allowed to edit or delete the post.
+        if ($this->isUserOwner($entity, $user)
```
kubabialy, line 216: Same applies here.

kubabialy (commented):
Overall looks good, I left some comments regarding readability improvements.

src/Core/Tools/Authorizer/PostAuthorizer.php
Mh-Asmi: thanks @kubabialy

src/Core/Tools/Authorizer/PostAuthorizer.php
Mh-Asmi: I totally agree with you,  this file has many if conditions like this one 
later may we can refactor all of them together to make the code more readable

kubabialy (approved):

PR #4454, opened 2022-07-25
fix: permissions of messages and contacts 

Mh-Asmi:
- add Contacts read permission for every logged-in user
- link the permissions of Messages with Manage Posts

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/src/Core/Tools/Authorizer/ContactAuthorizer.php
+++ b/src/Core/Tools/Authorizer/ContactAuthorizer.php
@@ -59,8 +59,8 @@ class ContactAuthorizer implements Authorizer
             return true;
         }
 
-        // Logged in users can search contacts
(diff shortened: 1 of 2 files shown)


webong (approved):

webong:
Good to go  🚀

tuxpiper:
Heads up to @AmTryingMyBest that we are expediting the release of this , because Uchaguzi is needing it very very much 🔥 🔥 

The likelihood of breaking functionality is very small here. The only real risk I can think of would be allowing users to do some other operation that we didn't really intend to allow.

We will get back to the whole subject of permissions shortly in the future, it looks like it deserves a complete review

Issue #4455, opened 2022-07-26
The source is not correctly being reported in the post detail view


tuxpiper:
**Describe the bug**

In this example, a post coming from Twitter (as appropriately indicated in the left-hand side of the screen) is shown as coming from the Web
![Screenshot 2022-07-26 at 20 21 03](https://user-images.githubusercontent.com/467393/181084508-da34cf3f-35fa-4116-98f5-e09123fb8b3f.png)

**Where was the bug observed**
https://uchaguzi.or.ke/posts/20600

**To Reproduce**
1. Go to https://uchaguzi.or.ke/posts/20600
2. Observe

**Expected behavior**
The post details component should show the correct source of the post.

**Is there a workaround? What is it.**
Nope

**Other observations**
It seems like the resources returned by the /api/v5/post/:id endpoint don't have the `source` property that the resources returned by the v3 API have.

PR #4456, opened 2022-07-28
fix: return the source of the post

Mh-Asmi:
- add source value to post details in `api/v5/posts`

Test checklist:
- [ ]

Fixes ushahidi/platform#4455 .

Ping @ushahidi/platform


v5/Http/Resources/PostResource.php
```diff
@@ -123,6 +123,16 @@ private function hydrateResourceRelationships($request)
         return $result;
     }
 
+    private function getPostSource()
+    {
+        $source = "web";
```
kubabialy, line 128: I would suggest using setting the web as a const in the top of class declaration so the context of what 'web' is is clear.

```php
class PostResource
{
    private const DEFAULT_SOURCE_TYPE = 'web';
    ...
```

v5/Http/Resources/PostResource.php
```diff
@@ -123,6 +123,16 @@ private function hydrateResourceRelationships($request)
...
+        $message = $this->message;
+        if ($message) {
```
kubabialy, line 130: Minor change to prevent redundant control flow.

```suggestion
        return $message && isset($message->type)
          : $message->type
          : self::DEFAULT_SOURCE_TYPE;
```

kubabialy (commented):
A bit of a cleanup comments, other than that great work

v5/Http/Resources/PostResource.php
```diff
@@ -123,6 +124,14 @@ private function hydrateResourceRelationships($request)
         return $result;
     }
 
+    private function getPostSource()
+    {
+        $message = $this->message;
+        return $message && isset($message->type)
+        ? $message->type
```
kubabialy, line 131: @Mh-Asmi last thing would be to fix the indentation

So something like:
```php
return $message && isset($message->type)
    ? $message->type
    : self::DEFAULT_SOURCE_TYPE;
```

kubabialy (commented):
Just a quick code-style fix and we're good to go!

v5/Http/Resources/PostResource.php
tuxpiper: @Mh-Asmi we would like to push this one. 🙏 Can you apply Kuba's adjustment suggestion, please?

tuxpiper (approved):
Looking good :+1:

tuxpiper:
@AmTryingMyBest  this is good for QA now

AmTryingMyBest:
@tuxpiper is this verifiable from the front-end? If so, how?

AmTryingMyBest:
QA'd, passes. 👍

Issue #4470, opened 2022-08-08
Failing to send SMS as a reply to a phone number that we first saw with USSD


tuxpiper:
**Describe the bug**
We are not able to send SMS to phone numbers that were first seen with a USSD or WhatsApp interaction.

**Where was the bug observed**
https://uchaguzi.or.ke

**To Reproduce**
1. Go to https://{your uchaguzi copy}/posts/25370
2. Write an outgoing reply such as : `Thank you. If you have any feedback related to elections , feel free to share it to 23102 for FREE. You can also dial *603#` 
3. You will see an error like ...

**Expected behavior**
The message should be sent

**Is there a workaround? What is it.**
None

**If the bug is datasource related (e.g. SMS or Targeted SMS), please specify the configuration details (e.g. Frontline + Clickatell, etc):**
USSD / WhatsApp + SMS

tuxpiper:
A chat excerpt about this (source: https://ushahidi.slack.com/archives/CFC4WQG4V/p1659951527034329 )

The payload to send the message looks like this:
(davidlosada)
```
POST /api/v3/messages
...
{"message":"Thank you. If you have any feedback related to elections , feel free to share it to 23102 for FREE. You can also dial *603#","direction":"outgoing","contact_id":9720,"parent_id":25909}
Notice the parent_id to the message table. message id=25909 looks like this:
id	parent_id	contact_id	post_id	user_id	data_source	data_source_message_id	title	message	datetime	type	status	direction	created	additional_data	notification_post_id
25909	NULL	9720	25920	NULL	ussd	random-d56a162e-4256-43b9-9645-838988d2af1b	(Fulfilled USSD survey)	(Fulfilled USSD survey)	2022-07-31 14:25:03	ussd	received	incoming	1659277503	NULL	NULL
(edited)
```

Validation error returned by the API:
```
{"errors":[{"status":422,"message":"Failed to validate messages entity:\ntype must be one of the available options"},{"status":422,"title":"type must be one of the available options","message":"type must be one of the available options","source":{"pointer":"\/type"}}]}
```

(wisdom)
The error is saying /api/v3/messages can’t accept a message where the type is ussd … from the payload there’s no type.  So you’re right, the child message is on default assuming the parent data type (channel).

https://github.com/ushahidi/platform/blob/5a64484dc71ca76f8aae72b03c4167a569383f54/src/Core/Usecase/Message/CreateMessage.php#L31-L38 (edited) 

[...]

PR #4472, opened 2022-08-08
Allow ussd messages to send via sms in africastalking source

webong:
- Update AfricasTalkingSource to handle ussd message type 
- Update Message create validator rules to work with ussd message type

Fixes ushahidi/platform#4470 .

Ping @ushahidi/platform


tuxpiper:
Keeping it as a custom solution put together for Uchaguzi

PR #4479, opened 2022-09-01
feat: add CQRS boilerplate

kubabialy:
- Adding CQRS boilerplate for future development.

Test checklist:
- [x] Unit tests added and passing
- [x] Example added

Ping @ushahidi/platform

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -2,32 +2,25 @@
 
 namespace Ushahidi\App\Providers;
 
-use Ushahidi\App\Bus\Command\CommandBus;
-use Ushahidi\App\Bus\Command\Example\ExampleCommand;
-use Ushahidi\App\Bus\Command\Example\ExampleCommandHandler;
-use Ushahidi\App\Bus\Query\Example\ExampleQuery;
-use Ushahidi\App\Bus\Query\Example\ExampleQueryHandler;
-use Ushahidi\App\Bus\Query\QueryBus;
-use Ushahidi\App\Tools\Features;
-use Ushahidi\Core\Tools\Verifier;
-use Ushahidi\Factory\UsecaseFactory;
-use Ushahidi\Addons\Mteja\MtejaSource;
-use Ushahidi\Core\Usecase\Post\Export;
 use Illuminate\Support\ServiceProvider;
-use Ushahidi\Core\Usecase\Export\Job\PostCount;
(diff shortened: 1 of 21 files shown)


app/Providers/AppServiceProvider.php
```diff
@@ -195,4 +202,24 @@ public function registerFeatures()
             return new Features($app[ConfigRepository::class]);
         });
     }
+
+    private function registerBusses(): void
+    {
+        $this->app->singleton(CommandBus::class, function ($app) {
+            $commandBus = new CommandBus($app);
+
+            $commandBus->register(ExampleCommand::class, ExampleCommandHandler::class);
```
webong, line 211: I am considering we have an automated way to register actions to the bus, maybe specified in a config file.

```
// config/cqrs.php
<?php
return [
   'queries' => [
        \Ushahidi\App\Bus\Query\Example\ExampleQuery::class => \Ushahidi\App\Bus\Query\Example\ExampleQueryHandler::class,
 ],
 
  'commands' => [
        \Ushahidi\App\Bus\Command\Example\ExampleCommand::class => \Ushahidi\App\Bus\Command\Example\ExampleCommandHandler::class,
 ],
];
```

Then in the `registerBusses` function in service provider we have this to register queries (vice versa for commands)

```
foreach (config('cqrs.queries') as $action => $handler) {
        $queryBus->register($action, $handler);
 }
```

webong:
@kubabialy looks good to me 👍🏽

app/Providers/AppServiceProvider.php
kubabialy: I don't think it's any more automated than what's in there really. But adds on complexity as there more files to keep track of.

webong (approved):

app/Providers/AppServiceProvider.php
webong: In the scenario, we have about 10 - 20 commands and queries to register, one would have to register in the AppServiceProvider making it a large file to work with, this is what I look at avoiding. Also, we only have one file `cqrs.php` in the config folder, where any one can go and register an action to the bus.

About complexity, I really don't think so ... this is a pattern widely adopted in the laravel ecosystem i.e the `app.php` uses such approach in registering providers.

Mh-Asmi: I agree with @webong about this point,  this file might be a quite large after adding many commands
& queries

Mh-Asmi (approved):

Issue #4500, opened 2022-10-10
Add TOS endpoints v5 implementation.


tuxpiper:
Needs PR

Mh-Asmi:
@tuxpiper this is the PR
https://github.com/ushahidi/platform/pull/4501

PR #4501, opened 2022-10-10
V5/tos

Mh-Asmi:
- add Pagination and Sorting Handling
- Error Handling  
- create v5 authorization
- add V5/Tos End Points
- add integration test for v5/tos 

Test checklist:
- [ ]

Fixes ushahidi/platform#4500 .

Ping @ushahidi/platform


config/paging.php
```diff
@@ -0,0 +1,10 @@
+<?php
+/**
+ * Paging Config
+ */
+return [
```
kubabialy, line 5: I think we could just keep them as constants. Not sure if there's a reason to keep them as environmental variables. They will be the same regardless of the environment, right?

src/Ushahidi/Modules/V5/Commands/AbstractBaseCommandHandler.php
```diff
@@ -0,0 +1,33 @@
...
+     * @return void
+     */
+    abstract protected function run(Command $command);
+
+    /**
+     * @param Command $command
+     * @return void
+     * @throws InvalidArgumentException
+     */
+    abstract protected function isSupported(Command $command);
```
kubabialy, line 26: No need for it there.

src/Ushahidi/Modules/V5/Commands/AbstractBaseCommandHandler.php
```diff
@@ -0,0 +1,33 @@
...
+
+abstract class AbstractBaseCommandHandler extends AbstractCommandHandler
+{
+    use Errors;
+    
+    /**
+     * @param Command $command
+     * @return void
+     */
+    abstract protected function run(Command $command);
```
kubabialy, line 19: I am not sure if it's a good idea. We should work with our interfaces not abstract classes to ensure a specific behaviour. I understand the reasoning but we have to think of a different way to introduce that idea.

src/Ushahidi/Modules/V5/Commands/CreateCommand.php
```diff
@@ -0,0 +1,59 @@
+<?php
+
+namespace Ushahidi\Modules\V5\Commands;
+
+use Illuminate\Database\Eloquent\Model;
+use App\Bus\Command\Command;
+
+class CreateCommand implements Command
```
kubabialy, line 8: I'd remove that. Queries and Commands should be specific, generalisation in this case will lead to more errors in the future.

src/Ushahidi/Modules/V5/Commands/Tos/CreateTosCommand.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Ushahidi\Modules\V5\Commands\Tos;
+
+use Ushahidi\Modules\V5\Commands\CreateCommand;
+use Ushahidi\Modules\V5\Models\Tos;
+
+class CreateTosCommand extends CreateCommand
```
kubabialy, line 8: See - because of 'generic' command you don't know what type of data you're working with here. Consider Commands and Queries to be DTO's that hold input data and input data only.

src/Ushahidi/Modules/V5/Commands/Tos/CreateTosCommandHandler.php
```diff
@@ -0,0 +1,61 @@
...
+        $this->isSupported($command);
+
+      
+        $this->authorizeForUser('store', Tos::class);
+        $input = $this->setInputDefaults($command->getInput(), 'store', $this->getUser());
+        $tos = new Tos();
+        if (!$tos->validate($input)) {
+            $this->errorInvalidData("Tos", $tos->errors->messages());
+        }
+        DB::beginTransaction();
```
kubabialy, line 33: I'd move the whole piece with DB to a repository.

src/Ushahidi/Modules/V5/Commands/Tos/CreateTosCommandHandler.php
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * @param CreateTosCommand $command
+     * @return void
+     */
+    protected function run(Command $command)
+    {
+        $this->isSupported($command);
+
+      
+        $this->authorizeForUser('store', Tos::class);
```
kubabialy, line 27: Shouldn't Controller already authenticate/authorise you to access data?

src/Ushahidi/Modules/V5/Common/Authorize.php
```diff
@@ -0,0 +1,67 @@
+<?php
+
+namespace Ushahidi\Modules\V5\Common;
+
+use App\Auth\GenericUser;
+use Illuminate\Support\Facades\Auth;
+use Illuminate\Contracts\Auth\Access\Gate;
+use Ushahidi\Modules\V5\Common\UserContext;
+
+trait Authorize
```
kubabialy, line 10: i would convert it to a class and use as an actual dependency. This way it's easier to follow and actually modify behaviour later on. And to test it properly.

src/Ushahidi/Modules/V5/Common/Errors.php
```diff
@@ -0,0 +1,79 @@
...
+
+namespace Ushahidi\Modules\V5\Common;
+
+use Exception;
+use Ushahidi\Core\Exception\NotFoundException;
+use Ushahidi\Core\Exception\ValidatorException;
+use Ushahidi\Core\Exception\AuthorizerException;
+use Symfony\Component\HttpKernel\Exception\HttpException;
+
+trait Errors
```
kubabialy, line 11: Please don't use traits for that. It's a pointless abstraction a little bit and will make reviewing the code later much easier. Keep it stupid simple, even if that means Extending NotFoundException, ValidatorException itd. This trait does not bring much value imho

src/Ushahidi/Modules/V5/Common/OwnerAccess.php
```diff
@@ -0,0 +1,30 @@
+<?php
+namespace Ushahidi\Modules\V5\Common;
+
+use App\Auth\GenericUser as user;
+use Illuminate\Database\Eloquent\Model;
+
+trait OwnerAccess
```
kubabialy, line 7: This should be an ACL. @tuxpiper something to consider in the future.

src/Ushahidi/Modules/V5/Common/Pageable.php
```diff
@@ -0,0 +1,30 @@
...
+
+    /**
+     * @var int
+     */
+    private $per_page;
+
+    private function setPagingParameters(Request $request):void
+    {
+         $this->per_page =
+         $this->hasCustomPerPage($request)?(int)$request->get("per_page"):config('paging.items_per_page');
```
kubabialy, line 19:
```suggestion
         $this->per_page = $this->hasCustomPerPage($request)
             ? (int) $request->get("per_page")
             : config('paging.items_per_page');
```

src/Ushahidi/Modules/V5/Common/PrivAccess.php
```diff
@@ -0,0 +1,42 @@
...
+trait PrivAccess
+{
+    /**
+     * Get a list of all possible privilges.
+     * By default, returns standard HTTP REST methods.
+     * @return Array
+     */
+    protected function getAllPrivs()
+    {
+        return ['read', 'create', 'update', 'delete', 'search'];
```
kubabialy, line 22: Just convert that to a const.

src/Ushahidi/Modules/V5/Common/PrivateDeployment.php
```diff
@@ -0,0 +1,45 @@
...
+     * @return boolean
+     */
+    public function isPrivate()
+    {
+        // if feature enabled and site set private in config
+        if (Features::isEnabled('private') && $this->getSite()->getSiteConfig('private', false)) {
+            return true;
+        }
+
+        return false;
```
kubabialy, line 29:
```suggestion
        return Features::isEnabled('private') && $this->getSite()->getSiteConfig('private', false)
```

src/Ushahidi/Modules/V5/Common/PrivateDeployment.php
```diff
@@ -0,0 +1,45 @@
...
+     * @return boolean
+     */
+    public function canAccessDeployment($user)
+    {
+        // Only logged in users have access if the deployment is private
+        if ($this->isPrivate() and !$user->id) {
+            return false;
+        }
+
+        return true;
```
kubabialy, line 43:
```suggestion
        return $this->isPrivate() and !$user->id
```

kubabialy (changes requested):
I am missing unit tests for Handlers as well.

src/Ushahidi/Modules/V5/Actions/Tos/Commands/CreateTosCommand.php
```diff
@@ -0,0 +1,57 @@
...
+namespace Ushahidi\Modules\V5\Actions\Tos\Commands;
+
+use App\Bus\Command\Command;
+
+class CreateTosCommand implements Command
+{
+    /**
+     * @var array
+     */
+    private $input;
```
kubabialy, line 12: Lets add the fields to command rather than keeping them in array. It would make it easier to work with.

kubabialy (approved):
Looks good. The only thing is that when we decide what to do with array $input in commands we should create a ticket and apply that suggestions to every command we have. 

While arrays won't break the implementation they lack the proper typing which makes the code more difficult to validate and understand.

PR #4505, opened 2022-10-13
Feat: Setup basics of the Authzn component

webong:
- Introduces an Authzn component in the platform src
- Introduces the Repository pattern with Eloquent compatibility
- Provides the ability to utilize Core Authorizers in Laravel Gate Policies
- Implemented the usage of Laravel Request class for validation (Using the Categories for Example)
- Other necessary modifications

Ping @ushahidi/platform


kubabialy:
I am somehow not terribly convinced about the idea of using this generic abstract class 'EloquentRepository'. We are trying to get ahead of expectations by introducing functionality that we probably won't need, but which introduces a certain complexity that we will have to manage.

While the idea of introducing generic operations is not bad in itself, in connection with the library used (Eloquent) the use of repositories in the way promoted by Hibernate or Doctrine is extremely cumbersome because database operations in Eloquent are mostly performed by the Model. 

It seems to me that it would be easier to do a per-model repository and in a specific repository contain specific operations for a particular model, which would give us a bit more room for manoeuvre in case we want to add another data source such as a text-search engine.

src/Ushahidi/Modules/V5/Models/Concerns/HasSlug.php
```diff
@@ -0,0 +1,89 @@
...
+            if (strlen($last_n) > 0) {
+                $last_count = intval($last_n);
+            }
+        }
+        // just the $slug if no similar slug found is null, start/keep counting otherwise
+        $value = $last ? "{$slug}-" . ($last_count+1) : $slug;
+
+        return $value;
+    }
+}
```
kubabialy, line 89: What do we need it for? I see it was moved from BaseModel, but what's the purpose of converting it into the Trait?

src/Ushahidi/Modules/V5/Models/Concerns/HasValidator.php
```diff
@@ -0,0 +1,46 @@
...
+     *
+     * @param array $inputArray associative array of values for this object to validate against and fill this object
+     */
+    public function validateAndFill($inputArray)
+    {
+        // must validate input before injecting into model
+        $this->validate($inputArray);
+        $this->fill($inputArray);
+    }
+}
```
kubabialy, line 46: Same here.

src/Ushahidi/Modules/V5/Repository/RoleRepository.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    protected static $root = Role::class;
+
+    public function doRolesExist(?array $roles = null)
+    {
+        if (!$roles) {
+            return true;
+        }
+
+        return $this->whereIn('role', $roles)->exists();
```
kubabialy, line 22:
```suggestion
        return $roles ? $this->whereIn('role', $roles)->exists() : true;
```

src/Ushahidi/Modules/V5/Repository/TagRepository.php
```diff
@@ -0,0 +1,22 @@
...
+use Ushahidi\Core\EloquentRepository;
+use Ushahidi\Contracts\Repository\Entity\TagRepository as TagRepositoryInterface;
+
+class TagRepository extends EloquentRepository implements TagRepositoryInterface
+{
+    protected static $root = Tag::class;
+
+    public function doesTagExist($value)
+    {
+    }
```
kubabialy, line 16: What do we need it for if its empty? If it's required by some interface/abstract class then it means we have to review that interface/abstract class.

src/Ushahidi/Modules/V5/RepositoryService.php
```diff
@@ -0,0 +1,38 @@
...
+     *
+     * @return void
+     */
+    public static function resolveRepositoryBinder()
+    {
+        if (isset(static::$repositoryBinderResolver)) {
+            call_user_func(static::$repositoryBinderResolver);
+        }
+    }
+}
```
kubabialy, line 38: Do we need this wrapper to register repositories? Laravel autowire should be able to handle that easily.

PR #4506, opened 2022-10-14
Chore: Modified V5/categories

webong:
- Implemented the usage of Laravel Request class for validation

Ping @ushahidi/platform


kubabialy (changes requested):
This pull request does not follow the rules we assumed. It does not have the division (CQRS) we agreed on. 

Do you want me to convert it into the CQRS after we merge other PRs?

PR #4509, opened 2022-11-15
V5/permissions

Mh-Asmi:
- add the APIs of V5\permissions using CQRS 

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


src/Ushahidi/Modules/V5/Policies/PermissionsPolicy.php
```diff
@@ -0,0 +1,99 @@
...
+        
+        $authorizer = service('authorizer.permission');
+        $user = $authorizer->getUser();
+
+      // These checks are run within the user context.
+      //$user = $this->getUser();
+
+      // Only allow admin access
+        if ($this->isUserAdmin($user)
+          and in_array($privilege, ['search', 'read'])) {
```
kubabialy, line 93: Don't use `and` - unless you specifically want to perform it. The way it works is different from && and might cause some funky stuff. 

If you look at following example:
```php
$foo = true;
$bar = false;

$isTrue = $foo and $bar;
```

`$isTrue` actually equals false. Because the operation it performs looks actually like that `($isTrue = $foo) and $bar`. This can cause some potential issues in the future. I would opt for `&&` instead.

src/Ushahidi/Modules/V5/Http/Controllers/PermissionsController.php
```diff
@@ -0,0 +1,63 @@
...
+     *
+     * @return TosCollection
+     * @throws \Illuminate\Auth\Access\AuthorizationException
+     */
+    public function index(Request $request, QueryBus $queryBus)
+    {
+       // $this->authorizeForCurrentUser('index', Permissions::class);
+        $resourceCollection = new PermissionsCollection(
+            $queryBus->handle(
+                new FetchPermissionsQuery(
```
kubabialy, line 43: I think we could use the FetchPermissionsQuery constants as a default values in the constructor itself, here: https://github.com/ushahidi/platform/pull/4509/files#diff-12be26afa959905d2b900fbfc37c45b829e65e705d4b8ecffa5c0256618e4907R20

`$request->query('foo')` returns null if it won't find the value inside the query string so constructor would default to a specific value set in the method declaration, and that way we won't have to think about it next time we use `FetchPermissionsQuery`

src/Ushahidi/Modules/V5/Repository/Permissions/PermissionsRepository.php
```diff
@@ -0,0 +1,36 @@
...
+     * @param string $order
+     * @return Permissions[]
+     */
+    public function fetch(
+        int $limit,
+        int $skip,
+        string $sortBy,
+        string $order,
+        array $search_data
+    ): LengthAwarePaginator;
```
kubabialy, line 25: We should probably return Collection, and Controller should convert that to the LengthAwarePaginator.

src/Ushahidi/Modules/V5/Actions/Permissions/Queries/FetchPermissionsQuery.php
```diff
@@ -0,0 +1,53 @@
...
+namespace Ushahidi\Modules\V5\Actions\Permissions\Queries;
+
+use App\Bus\Query\Query;
+
+class FetchPermissionsQuery implements Query
+{
+    const DEFAULT_LIMIT = 0;
+    const DEFAULT_ORDER = "ASC";
+    const DEFAULT_SORT_BY = "id";
+    const AVAILABLE_SEARCH_FIELDS = ['q' /* LIKE name */];
```
kubabialy, line 12: If there's more than just one field maybe it's worth creating a DTO class to represent fields we accept?  Something like

```php
class PermissionsAvailableSearchFieldsDTO
{
    private $query;

    public function __construct(string $query)
    {
         $this->query = $query;
    }

    public function getQuery(): string
    {
        return $this->query;
    }
}
```

If it's just query/q then maybe we should just convert it to the string? There would be less operation around it to get that `q` value.

src/Ushahidi/Modules/V5/Policies/PermissionsPolicy.php
```diff
@@ -0,0 +1,99 @@
...
+     * @param Permissions $permissions
+     * @param string $privilege
+     * @param user $user
+     * @return bool
+     */
+    public function isAllowed($permissions, $privilege, $user = null):bool
+    {
+        
+        $authorizer = service('authorizer.permission');
+        $user = $authorizer->getUser();
```
kubabialy, line 86: You don't have to overwrite $user that you passed into the method.

src/Ushahidi/Modules/V5/Policies/PermissionsPolicy.php
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * @param Permissions $permissions
+     * @param string $privilege
+     * @param user $user
+     * @return bool
+     */
+    public function isAllowed($permissions, $privilege, $user = null):bool
+    {
+        
+        $authorizer = service('authorizer.permission');
```
kubabialy, line 85: Can we inject it through the constructor?

kubabialy (commented):
Overall looks really good, just couple of notes.

src/Ushahidi/Modules/V5/Policies/PermissionsPolicy.php
Mh-Asmi: this will be changed after merge , I will use the tool\authorizer

src/Ushahidi/Modules/V5/Policies/PermissionsPolicy.php
Mh-Asmi: this will be changed after merge , we will use the entity

src/Ushahidi/Modules/V5/Http/Controllers/PermissionsController.php
Mh-Asmi: I tried to do as you suggest but when it constructor doesn't take the default value when passing null.

kubabialy (approved):

Issue #4262, opened 2021-06-04
Posts with certain script types (e.g Gujarati language) display an error when being submitted


CeciliaHinga:
**Describe the bug**
When one creates a post using the Gujarati or Punjabi language, the post is not successfully created when submitted, displaying an error at the bottom of the screen.

**Where was the bug observed**
Ushahidi.io

**To Reproduce**

1. Go to map/data view
2. Click on add new post (+)
3. Select the survey you want to fill
4. Enter the information required in Gujarati or Punjabi language on the title and description fields
5 Click on submit, see error  "**title fields.title does not match the required format..... slug Slug is required.**" 

**Expected behavior**
The Ushahidi Platform can store any language/script added to the Unicode standard. Meaning, posts with the above languages should be successfully submitted without displaying any error.

**Is there a workaround? What is it?**
I haven't discovered any yet

**Screenshots**
<img width="863" alt="Create_Post_-_test254" src="https://user-images.githubusercontent.com/20906968/120812783-fb10b400-c555-11eb-903c-6bde17934ab1.png">

tuxpiper:
The problem with the slug is addressed now, but the "title fields.title does not match the required format" validation is still failing

tuxpiper:
Maybe this is not a problem with newer versions of Lumen. Should give it a try.

PR #4514, opened 2022-11-29
Create custom rule to validate standard text values

webong:
- Adds a custom rule to validate text value

Test checklist:
- [ ] Try creating a post using Bengali characters
- [ ] Try creating a post using Gujarati characters
- [ ] Try creating a post using Latin based characters

Fixes ushahidi/platform#4262 .

Ping @ushahidi/platform

### DIFF
--- a/v5/Rules/StandardText.php
+++ b/v5/Rules/StandardText.php
@@ -1,0 +1,53 @@
+<?php
+
+namespace v5\Rules;
+
+use Illuminate\Contracts\Validation\Rule;
+
+class StandardText implements Rule
+{
+    protected $patterns = [
+        // Regex that only allows letters, numbers, punctuation, and space.
+        '/^[\pL\pN\pP ]++$/uD',
+        // Regex for bengali characters
+        '/^[\p{Bengali} ]{0,100}$/u',
+        // Regex for bengali characters
+        '/^[\p{Gujarati} ]{0,100}$/u'
+    ];
+
+    /**
+     * Determine if the validation rule passes.
+     *
+     * @param  string  $attribute
+     * @param  mixed  $value
+     * @return bool
+     */
+    public function passes($attribute, $value)
+    {
+        $passes = false; // Default pass value
+
+        foreach ($this->patterns as $pattern) {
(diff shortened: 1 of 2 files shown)


v5/Rules/StandardText.php
```diff
@@ -0,0 +1,53 @@
...
+use Illuminate\Contracts\Validation\Rule;
+
+class StandardText implements Rule
+{
+    protected $patterns = [
+        // Regex that only allows letters, numbers, punctuation, and space.
+        '/^[\pL\pN\pP ]++$/uD',
+        // Regex for bengali characters
+        '/^[\p{Bengali} ]{0,100}$/u',
+        // Regex for bengali characters
```
tuxpiper, line 14: copy-pasted comment :)

v5/Rules/StandardText.php
```diff
@@ -0,0 +1,53 @@
...
+
+use Illuminate\Contracts\Validation\Rule;
+
+class StandardText implements Rule
+{
+    protected $patterns = [
+        // Regex that only allows letters, numbers, punctuation, and space.
+        '/^[\pL\pN\pP ]++$/uD',
+        // Regex for bengali characters
+        '/^[\p{Bengali} ]{0,100}$/u',
```
tuxpiper, line 13: What would happen if someone mixes character sets in their input? I.e. they write something in Bengali, but then in the middle of the sentence, they write i.e. "Twitter", because they are referring to it as a source.

Would that fail validation?

Mh-Asmi (approved):
it's good for me

v5/Rules/StandardText.php
webong: Oops ... I tested this edge case, and it fails validation.

webong: @tuxpiper so I updated the regex to handle a wide case of characters

https://github.com/ushahidi/platform/pull/4514/files#diff-4d687cdb0cd21587e307890aa061024d77106c84dc4f4174933f90871434314dR12-R13

This works when someone mixes character sets in their titile.

Issue #4522, opened 2023-02-16
Error saving/getting files to storage


lupa18:
**Describe the bug**
When I upload a file, then it get saved at: **storage/app/6/3/** but when accessing the frontend it tries to get at: **storage/6/3/**

So images can't be displayed. 

**Where was the bug observed**
https://elavizor.tedic.org/

**To Reproduce**
1. Install platform using docker, also client
2. Upload an image
3. You will see is saved on your backend
4. But, doesn't show con frontend

**Expected behavior**

You should see your images here: 
![image](https://user-images.githubusercontent.com/463429/219500846-ba917222-a717-44a4-8d43-82cc66621640.png)

They are in the backend: 
![image](https://user-images.githubusercontent.com/463429/219500976-c793575d-344e-4c57-a068-8be909b76f00.png)

**Additional context**
Installed and running using docker

PR #4523, opened 2023-02-17
Update filesystems.php

lupa18:
Without this chane, then you'll have an error getting/accessing your images (static content): #4522

- Change storage public path

Fixes ushahidi/platform#4522

Ping @ushahidi/platform


lupa18:
thanks @tuxpiper ! I will try. Is this documented anywhere?

Issue #4524, opened 2023-02-20
V3: Problem with CSV Exporter


webong:
**Describe the bug**
A problem with the data they download on CSV. If you look at the attached screenshot, the data under ‘Categories’ is displaying correctly (with each word in English and Arabic) but the data under ‘Place of incident’ and ‘Consequence of the violence’ has a string of random characters where the Arabic words should be. (Reported by @JosephUshahidi on Slack).

**Where was the bug observed**
https://ceasefire-mena.ushahidi.io/views/map

**Screenshots**
![image](https://user-images.githubusercontent.com/30173059/220095855-bc387403-8d5b-4ddd-8320-b7dcc4f9d4d0.png)

PR #4525, opened 2023-02-20
fix: CSV formater not decoding values for checkbox attribute

webong:
- Includes a method that decodes the value for checkbox-based attributes on the CSV Formatter.

Fixes ushahidi/platform#4524.

Ping @ushahidi/platform

### DIFF
--- a/src/App/Formatter/Post/CSV.php
+++ b/src/App/Formatter/Post/CSV.php
@@ -194,7 +194,8 @@ class CSV extends API
 
         foreach ($records as $record) {
             $values = $this->formatRecordForCSV($record, $attributes);
-            fputcsv($stream, $values);
+            // fputcsv($stream, $values);
+            fputcsv($stream, mb_convert_encoding($values, 'UTF-8'));
         }
 
         return $this->writeStreamToFS($stream);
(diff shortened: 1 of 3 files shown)


src/App/Formatter/Post/CSV.php
```diff
@@ -379,6 +354,39 @@ private function getValueFromRecord($record, $keyParam, $attributes)
         return $return;
     }
 
+    private function writeStreamToFS($stream)
```
tuxpiper, line 357: @webong, just double checking, did anything really change with this method? Or is it just that the differ lost context as the placement changed inside the file? It looks pretty much the same to me...

src/App/Formatter/Post/CSV.php
```diff
@@ -395,7 +403,7 @@ private function multiColumnArray($recordValue, $headingKey, $key)
         return isset($recordValue[$headingKey][$key]) ? ($recordValue[$headingKey][$key]) : '';
     }
 
-    private function singleColumnArray($recordValue, $headingKey, $separator = ',')
+    private function singleColumnArray($recordValue, $headingKey, $separator = ', ')
```
tuxpiper, line 406: @webong could you please explain this change?

src/App/Formatter/Post/CSV.php
```diff
@@ -487,6 +495,27 @@ protected function isLocation($value)
...
     }
 
+    /**
+     * Decode checkbox value
+     *
+     * @param mixed $value
+     *
+     * @return array
+     */
+    protected function decodeCheckboxValue($value)
```
tuxpiper, line 505: @webong  Would the result of this be that checkboxes are now exported into CSV encoded to JSON? In that case, this could get confusing to folks using the CSV

src/App/Formatter/Post/CSV.php
webong: Just the placement of the method changed.

src/App/Formatter/Post/CSV.php
webong: Not at all, after it's converted to json there's an extra formatter that uses the explode function in php to convert the object to a series of string.

tuxpiper (approved):

PR #4529, opened 2023-03-07
v5 / users - role - permissions update

Mh-Asmi:
- fix v5/users validation 
- fix v5/roles validation 
- add permissions to user
- add allowed privileges to users-role-permissions

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/src/Ushahidi/Modules/V5/Http/Resources/Permissions/PermissionsResource.php
+++ b/src/Ushahidi/Modules/V5/Http/Resources/Permissions/PermissionsResource.php
@@ -3,6 +3,7 @@ namespace Ushahidi\Modules\V5\Http\Resources\Permissions;
 
 use Illuminate\Http\Resources\Json\Resource;
 use Ushahidi\Modules\V5\Http\Resources\RequestCachedResource;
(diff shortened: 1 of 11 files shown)


src/Ushahidi/Modules/V5/Http/Resources/Permissions/PermissionsResource.php
```diff
@@ -11,6 +12,23 @@ class PermissionsResource extends Resource
 
     public static $wrap = 'data';
 
+
+    private function getResourcePrivileges()
+    {
+        $authorizer = service('authorizer.permission');
```
webong, line 18: Is there a way to call the Authorizer without using the service helper function, as this uses Aura DI, which we want to remove in the future?

Mh-Asmi: absolutely we will find another way, now I didn't find but I need to return this value, so the client side can build their SDK

webong (approved):

PR #4535, opened 2023-03-08
Throw exception when checking commands and queries

Mh-Asmi:
- Throw an exception when checking commands and queries

Test checklist:
- []

Fixes ushahidi/platform# .

Ping @ushahidi/platform


kubabialy (approved):
Looks good to me 👍

PR #4586, opened 2023-06-14
Resolve Broken OAuth Service for ushahidi.io based deployments

webong:
- Let's passport know about new database changes when a site is detected 
- Adjusts classed based on V5 PostController to be PHP intellisense friendly

Ping @ushahidi/platform


tuxpiper (approved):
LGTM :+1:

PR #4633, opened 2023-08-22
set all tasks not completed for updated post

Mh-Asmi:
-

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


tuxpiper (approved):
It looks good to me from purely a code review stance.

I have to admit I have no understanding as to what's the reason behind the change :)

Mh-Asmi:
@tuxpiper 
The problem is in the old condition when the client pass in an empty array, the old condition is seeing it as false

PR #4654, opened 2023-10-18
Mailbox DataSource

webong:

- Enables Mailbox datasource connect to an SSL IMAP or POP3 server with a self-signed certificate

Ping @ushahidi/platformTest checklist:


AmTryingMyBest (approved):
LGTM  👍

PR #4669, opened 2023-11-13
hide contact info for unauthorized users

Mh-Asmi:
- hide author and contact info for users don't have a manage-post permission
- confirm that this info contact info will not return inside message too
- apply these rules on `api/v5/posts` and `api/v5/posts/{id}` 

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/src/Ushahidi/Modules/V5/Actions/Post/Handlers/AbstractPostQueryHandler.php
+++ b/src/Ushahidi/Modules/V5/Actions/Post/Handlers/AbstractPostQueryHandler.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace Ushahidi\Modules\V5\Actions\Post\Handlers;
+
+use App\Bus\Action;
+use App\Bus\Query\Query;
+use Ushahidi\Modules\V5\Actions\V5QueryHandler;
+use Illuminate\Support\Facades\Auth;
+use Ushahidi\Modules\V5\Models\RolePermission;
+
+abstract class AbstractPostQueryHandler extends V5QueryHandler
+{
+    protected function userHasManagePostPermissions()
+    {
+        $user = Auth::user();
+        if (!$user || !$user->id) {
+            return false;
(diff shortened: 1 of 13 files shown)


src/Ushahidi/Modules/V5/Actions/Post/Handlers/AbstractPostQueryHandler.php
```diff
@@ -0,0 +1,38 @@
...
+        }
+        if ($user->role === "admin") {
+            return true;
+        }
+        $permissions =
+            RolePermission::select("permission")->where('role', '=', $user->role)->get()->pluck('permission');
+        if (in_array("Manage Posts", $permissions->toArray())) {
+            return true;
+        }
+        return false;
```
webong, line 27: This piece of code already exists in Ushahidi\Core\Tool\Acl class so no need for code duplication

webong (approved):
I'll give this a good pass since this is a hotfix. We can always refactor in future iteration

src/Ushahidi/Modules/V5/Actions/Post/Handlers/AbstractPostQueryHandler.php
webong: See here https://github.com/ushahidi/platform/blob/88c5381da67bf95bd28dc9f637282308a2357bcf/src/Ushahidi/Core/Tool/Acl.php#L36-L56

So the best way would be to inject the Acl class via the constructor into the handler, Laravel Container handles the resolution, and we can the do permission checks

`$this->accessControl->hasPermission($user, 'Manage Post');`

tuxpiper:
> I'll give this a good pass since this is a hotfix. We can always refactor in future iteration

Thank you @webong 🙏  Could you please create a quick technical task in Linear, so that we don't forget?

PR #4675, opened 2023-11-30
Hotfix/ush 1018

webong:
- In the V5 Config Handler, use laravel config helper to read default settings value

Fixes USH-1018

Ping @ushahidi/platform


Mh-Asmi (approved):
The code is good for me and forth x the issue, but I didn't check it on my local as I'm outside.
@webong please confirm that it is fixed by checking your local laravel log doesn't have this error log

PR #4678, opened 2023-12-04
fix: fallback to null when user attribute value is undefined

webong:
- Add null coalesce to the magic method __get on the generic user support class

Fixes USH-1033

Ping @ushahidi/platform


Mh-Asmi (approved):
it's good for me

PR #4679, opened 2023-12-12
Feature/ush 987 add mobile data source

webong:
- Require jenssegers/agent package for parsing user agents 
- Create migrations for entering source and metadata for a post entity
- Update the Create Post handler to check if request is from a mobile source and record it
- Update the Post Resources / Data Transformers to allow retrieveing the mobile source
- Update eloquent query to collect stats from mobile source

Fixes USH-987.

Ping @ushahidi/platform


Mh-Asmi (approved):
I checked it on my local device.
it passed all tests successfully 👍 , and I test some cases manually 
I like the adding of meta data to save the source info , thanks @webong 
Approved!

PR #4686, opened 2024-01-03
Edit post: Unable to specifically remove "Add caption" field for image

ushahidlee:
Fixes USH-777. Adds a patch endpoint to update a caption on an existing media item.


Mh-Asmi (approved):
good for me

PR #4714, opened 2024-02-07
Ush 1134

ushahidlee:
- It prevents the client from updating a role name, and returns a 422 if they try.
- It fixes an issue with updating an admin user, wiping all their permissions.
- Removes broken integration tests

https://linear.app/ushahidi/issue/USH-1134/backend-allows-admin-role-to-be-modified


Mh-Asmi (approved):
good for me

PR #4716, opened 2024-02-12
Allow user to delete their own contact

ushahidlee:
This allows the user to delete their own contact.

Fixes: https://linear.app/ushahidi/issue/USH-1132/delete-contact-button-is-still-unavailable-for-non-admin-users


Mh-Asmi (approved):
good 👍

PR #4717, opened 2024-02-13
return all permissions for admin role

Mh-Asmi:
- I update the sho of role to return all available permissions when the role is admin 
- I create migration to adding the missing permissions for admin

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


ushahidlee (approved):
Everything looks good and works as described

PR #4816, opened 2024-03-12
USH-1125: disable arbitrary limitation on CSV import file data size

tuxpiper:
- Disables arbitrary limitation of 20KB in imported CSV file
- Add a comment to the point that the limitation should be enabled only once it's configurable


src/Ushahidi/Modules/V5/Requests/CSVRequest.php
```diff
@@ -26,10 +26,14 @@ public function rules(Request $request)
...
     }
+    
     public function getMaxBytes()
     {
         // To Do: make it config
-        return 20000;
+        // David Losada update 2024-03-15: originally 20KB, that size is arbitrary and not useful to anyone.
+        //    Setting to 1TiB, instead
+        //    Let's make it something reasonable when the work to make this a config value happens.
+        return 1024 * 1024 * 1024 * 1024;
```
ushahidlee, line 36: I think this is a bit high. Can I suggest somewhere in the range of 20mb - 50mb?

Mh-Asmi (approved):

PR #4827, opened 2024-03-14
Email Issues

ushahidlee:
This PR contains a bugfix and new functionality.

Bug: The data source was selecting the last imported message from the platforms database and using its id as the starting point to fetch new messages from the mailbox. This means that if the data source was removed and then re-added, it would use the previously fetched message id as a starting point and not fetch messages until the new messages had ids that exceeded that value. This fix allows the data source to save the last fetch email id in its config, which is deleted with the data source.

New Functionality: I have added a new option to the setup which allows the user to specify whether they would like all email fetched (ie. everything in the inbox irrespective of the number of emails) or only unread email (the new option).

### DIFF
--- a/config/data-provider.php
+++ b/config/data-provider.php
@@ -31,7 +31,7 @@ return [
         'incoming_username' => '',
         'incoming_password' => '',
         'incoming_unread_only' => 'Unread',
-        'incoming_last_uid' => 0
+        'incoming_last_uid' => '0'
     ],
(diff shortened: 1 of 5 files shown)


Mh-Asmi (approved):
I didn't tet the branch but the code is good for me 👍

tuxpiper:
Nice 👌  thank you, @ushahidlee  !

PR #4922, opened 2024-04-22
exclude posts of survey with private locations

Mh-Asmi:
-

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/src/Ushahidi/Modules/V5/Http/Controllers/PostController.php
+++ b/src/Ushahidi/Modules/V5/Http/Controllers/PostController.php
@@ -45,11 +45,11 @@ use Ushahidi\Core\Concerns\AdminAccess;
 class PostController extends V5Controller
 {
 
-     // It uses methods from several traits to check access:
+    // It uses methods from several traits to check access:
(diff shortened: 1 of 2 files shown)


src/Ushahidi/Modules/V5/Http/Controllers/PostController.php
```diff
@@ -354,15 +354,26 @@ public function delete(int $id, Request $request)
 
     public function stats(Request $request)
     {
-        $stats = $this->queryBus->handle(PostsStatsQuery::FromRequest($request));
+        $surveys_with_private_location  = $this->queryBus->handle(new GetSurveyIdsWithPrivateLocationQuery());
```
ushahidlee, line 357: This query should be moved into the if clause and only be executed if the user cannot see posts with private location.

ushahidlee (changes requested):
Small change for better performance.

ushahidlee (approved):
👍

PR #4957, opened 2024-05-30
update post create to accept whatsapp posts

Mh-Asmi:
- add new column (contact_id) to posts table
- in posts creation for post of Whatsapp datasource :
       - I will check if there is a whatsapp contact for the passed phone number, if not I will create new one
       - use the id of the Whatsapp contact as value for contact_id column
       - set the source in post with value "whatsapp"
       - save contact info inside the metadata 
       - set the author name with contact number, and author email null

- in getting posts
     - return the correct source of the post
     - return the contact number in the author realname 
     - return the full contact object if the user is authorized users
     - return  only contact_id of contact object for unauthorized users

- in filters
    - add Whatsapp to source filter
    - return the correct count of Whatsapp posts

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


tuxpiper (approved):
Looks good 👍  I think I follow most of the changes.

PR #4986, opened 2024-08-06
fix the export in multisite  

Mh-Asmi:
- use SerializesAndRestoresModelIdentifiers instead of SerializeModels

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


tuxpiper (approved):
Cool! I like this approach 👍

PR #4996, opened 2024-10-01
chore(docker): allow composer to run as root

tuxpiper:
- Our containers are set up to run as root and composer expects this environment variable to be set up in that case

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


ushahidlee (approved):
An elegant solution :+1:

PR #4998, opened 2024-10-03
survey only parameters for relations

Mh-Asmi:
- add only for survey relations
- update paging for survey

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


tuxpiper (approved):
I see that I need to spend some more time with v5 code , there's only so much I understand here. 
In any case, it looks good to me 👍

PR #5013, opened 2024-11-11
chore(db): set collation specific to mariadb and fix phinx configuration

tuxpiper:
- Sets the server collation in local environment to utf8_unicode_520_ci , which is based on UCA 5.2.0 weight keys (http://www.unicode.org/Public/UCA/5.2.0/allkeys.txt) (as opposed to UCA 4.0 in utf8_generic_ci). It should sort alphabets more reliably.
- Addresses a configuration flaw in phinx , which would ignore the character set configuration at the time of creating tables. So it seems that setting the collation is required. 

Test checklist:
- [ ] run `make start`
- [ ] connect to the database and run `SHOW TABLE STATUS where name like '%';` , the collation should be utf8mb4_unicode_520_ci .
( if you do this before applying the changes, you would see that the collation is utf8_* , or utf8mb3_* , which means we'd using a 3-byte charset that will not accept all the emojis 🤔😓🤮💣😵 and other higher-page characters )

Fixes USH-1670

Ping @ushahidi/platform

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,9 +7,10 @@ services:
       MARIADB_DATABASE: ushahidi
       MARIADB_USER: ushahidi
       MARIADB_PASSWORD: ushahidi
+      LANG: C.UTF_8
     command:
       - --character-set-server=utf8mb4
-      - --collation-server=utf8mb4_unicode_ci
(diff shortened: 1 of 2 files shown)


Mh-Asmi (approved):
approved

PR #5014, opened 2024-11-19
only parameter for category

Mh-Asmi:
-

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


tuxpiper (approved):
It's quite a bit of change here and I can't seem to be able to get very deep into it,  but I think it looks good. Maybe we could write some tests in a pairing session.

PR #5026, opened 2025-01-21
exclude post list from the default of saved search query

Mh-Asmi:
- in this update the call of 
/api/v5/savedsearch  or  /api/v5/savedsearch/{id} will not include posts
but in only has posts then it will be returned 

Test checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


tuxpiper (approved):
Looks good to me 👍

